PR #6, opened 2022-02-26
Configuration

oravi:


macros/utils/datatypes.sql
```diff
@@ -0,0 +1,12 @@
+{%- macro type_bool() -%}
+    {{ return(adapter.dispatch('type_bool', 'elementary_data_reliability')()) }}
```
Maayan-s, line 2: Change the name of the package so our namespace will be shorter

macros/system/configuration/upload_edr_configuration.sql
```diff
@@ -0,0 +1,133 @@
+{% macro upload_tables_configuration() %}
+    {% if execute %}
+        {% set nodes = elementary_data_reliability.get_nodes_from_graph() %}
+        {% set table_monitors = [] %}
+        {% for node in nodes | selectattr('resource_type', 'in', 'seed,source,model') -%}
```
Maayan-s, line 5: What happens if these keys are not in the node?

macros/system/configuration/upload_edr_configuration.sql
```diff
@@ -0,0 +1,133 @@
...
+            {% set table_config_dict = elementary_data_reliability.get_table_config(node) %}
+            {% if table_config_dict is not none %}
+                {% do table_monitors.append(table_config_dict) %}
+            {% endif %}
+        {% endfor %}
+        {% if table_monitors|length > 0 %}
```
Maayan-s, line 11: You can just do 
{% if table_monitors|length %}
Do you prefer to be explicit?

macros/system/configuration/upload_edr_configuration.sql
```diff
@@ -0,0 +1,133 @@
...
+            {% do elementary_data_reliability.insert_dicts_to_table(this, table_monitors) %}
+        {% endif %}
+    {% endif %}
+    {{ return('') }}
```
Maayan-s, line 15: What does the empty return do?

macros/system/configuration/upload_edr_configuration.sql
```diff
@@ -0,0 +1,133 @@
...
+        {% if edr_config is mapping %}
+            {% do res.update(edr_config) %}
+        {% endif %}
+    {% endif %}
+    {% set edr_config = node.meta.get('edr') %}
+    {% if edr_config is mapping %}
+        {% do res.update(edr_config) %}
+    {% endif %}
+    {% set edr_config_in_columns = elementary_data_reliability.get_edr_config_in_columns(node) %}
+    {% if edr_config_in_columns | length > 0 %}
```
Maayan-s, line 93: same as just |length

macros/system/configuration/upload_edr_configuration.sql
```diff
@@ -0,0 +1,133 @@
...
+        {% do res.update({'columns': edr_config_in_columns}) %}
+    {% endif %}
+    {{ return(res) }}
+{% endmacro %}
+
+{% macro get_table_config(node) %}
+    {% set table_name = elementary_data_reliability.get_table_name(node) %}
+    {% set edr_config = elementary_data_reliability.get_edr_config(node) %}
+    {% if edr_config is mapping %}
+        {% set table_monitored = edr_config.get('monitored') %}
```
Maayan-s, line 103: What happens if monitored or columns_monitored are missing?

macros/utils/dicts.sql
```diff
@@ -0,0 +1,35 @@
+{% macro insert_dicts_to_table(table_name, dict_list) -%}
+    {% set insert_dicts_query %}
```
Maayan-s, line 2: Maybe check if table exists?

models/configuration_and_artifacts/column_monitors_config.sql
```diff
@@ -0,0 +1,10 @@
+{{
+  config(
+    materialized = 'table',
+    transient=false,
+    unique_key = 'full_name',
+    post_hook = "{% if execute %}{{ elementary_data_reliability.upload_columns_configuration() }}{% endif %}"
+  )
+}}
+
+{{ empty_table([('full_name', 'string'), ('database_name', 'string'), ('schema_name', 'string'), ('table_name', 'string'), ('column_name', 'string'), ('column_monitors', 'string')]) }}
```
Maayan-s, line 10: In the rest of the project I use:
'full_table_name'
'full_column_name'
'full_schema_name'

The config is not 100% part of the project, but maybe we should be consistent anyway.

macros/system/configuration/upload_edr_configuration.sql
oravi: It's not possible as it means there are no models / sources / seeds and even if it's possible I believe that it should be ok as the list should be empty and then I'm checking that tables_monitors is not empty so it just won't do anything

macros/system/configuration/upload_edr_configuration.sql
oravi: Yeah, I prefer to be explicit

PR #7, opened 2022-02-28
Dbt artifacts support

oravi:


macros/utils/table_operations.sql
```diff
@@ -24,6 +24,14 @@
...
 
+{% macro remove_empty_rows(table_name) %}
+    {% set columns = adapter.get_columns_in_relation(table_name) -%}
+    {% set delete_empty_rows_query %}
+        delete from {{ table_name }} where {% for column in columns -%} {{ column.name }} is NULL {{- " and " if not loop.last else "" -}} {%- endfor -%}
+    {% endset %}
+    {% do run_query(delete_empty_rows_query) %}
+{% endmacro %}
+
 {% macro get_dict_value_with_default(dict, key, default) -%}
```
Maayan-s, line 35: Change name to insensitive case get key

PR #22, opened 2022-05-17
Timestamp column test param

elongl:
An attempt to implement [Issue #72](https://github.com/elementary-data/elementary/issues/72) that adds support for passing the `timestamp_column` for individual tests rather than the global scope.  
Note that the documentation needs to be updated to clarify that this feature is supported.

I'd love to get some feedback and edit accordingly.  
Please let me know what you think.  
Thanks.


macros/edr/tests/test_all_columns_anomalies.sql
```diff
@@ -16,20 +16,25 @@
...
             {{- elementary.test_log('monitored_table_not_found', full_table_name) }}
             {{- return(elementary.no_results_query()) }}
         {%- endif %}
 
-        {%- set timestamp_column = elementary.insensitive_get_dict_value(table_config, 'timestamp_column') %}
-        {{- elementary.debug_log('timestamp_column - ' ~ timestamp_column) }}
-        {%- set timestamp_column_data_type = elementary.insensitive_get_dict_value(table_config, 'timestamp_column_data_type') %}
-        {{- elementary.debug_log('timestamp_column_data_type - ' ~ timestamp_column_data_type) }}
+        {% if timestamp_column %}
+            {%- set timestamp_column_data_type = elementary.find_normalized_data_type_for_column(model, 'timestamp_column') %}
```
oravi, line 28: The macro 'find_normalized_data_type_for_column' should receive the actual value of the variable timestamp_column and not the string 'timestamp_column' :)

macros/edr/tests/test_all_columns_anomalies.sql
```diff
@@ -16,20 +16,25 @@
...
         {%- if not model_relation %}
             {{- elementary.test_log('monitored_table_not_found', full_table_name) }}
             {{- return(elementary.no_results_query()) }}
         {%- endif %}
 
-        {%- set timestamp_column = elementary.insensitive_get_dict_value(table_config, 'timestamp_column') %}
-        {{- elementary.debug_log('timestamp_column - ' ~ timestamp_column) }}
-        {%- set timestamp_column_data_type = elementary.insensitive_get_dict_value(table_config, 'timestamp_column_data_type') %}
-        {{- elementary.debug_log('timestamp_column_data_type - ' ~ timestamp_column_data_type) }}
+        {% if timestamp_column %}
```
oravi, line 27: Looking at the code now I think we can write this code in a separate macro (also to avoid code duplication, I know it has been this way before but we can at least avoid adding more to it). Also, the whole purpose of get_table_config_from_graph is to extract the timestamp column. So I think we can rename it to get_timestamp_column_from_graph (the graph is dbt's object that holds all the metadata about the models) and if we don't have the timestamp_column (meaning, its value is none) call this get_timestamp_column_from_graph and then extrant the data type by calling your new function find_normalized_data_type_for_column.

elongl:
I've applied your changes, care to take a second look please?    
Also, could you perhaps clarify when does one use `{%-`? I couldn't quite pick up a pattern.  
I think some general rules should be applied or drop it altogether given that it only affects compiled code.
Thanks.

macros/edr/tests/test_utils/get_timestamp_column_from_graph.sql
```diff
@@ -0,0 +1,20 @@
...
+            {% for node in depends_on_nodes %}
+                {% if node.name | lower == model_name %}
+                    {% set elementary_config = elementary.get_elementary_config_from_node(node) %}
+                    {% if elementary_config and elementary_config is mapping %}
+                        {{ return(elementary_config.get('timestamp_column')) }}
+                    {% endif %}
+                {% endif %}
+            {% endfor %}
+        {% endif %}
+    {% endif %}
```
oravi, line 19: Make sure this macro returns something in case the timestamp column was not found, you can simply return none

macros/edr/tests/test_utils/find_normalized_data_type_for_column.sql
```diff
@@ -0,0 +1,12 @@
...
+
+{% set columns_from_relation = adapter.get_columns_in_relation(model_relation) %}
+{% if columns_from_relation and columns_from_relation is iterable %}
+    {% for column_obj in columns_from_relation %}
+        {% if column_obj.column | lower == column_name | lower %}
+            {{ return(elementary.normalize_data_type(column_obj.dtype)) }}
+        {% endif %}
+    {% endfor %}
+{% endif %}
+
```
oravi, line 11: Please make sure this macro returns none if the column was not found

oravi:
Thanks @elongl ! see last comments in the PR. Regarding your question around using '{%' vs '{%-' , when you use '{%-' in Jinja it removes whitespaces and break lines in the rendered text ([see this to learn more](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control)). We try to use it as much as possible as the result test query will contain less useless spaces and '\n' (the test query can be seen for example in our Slack alerts so it's important)

oravi:
@elongl good job here @elongl ! thanks for working on this 🥇

PR #27, opened 2022-05-30
Test names

oravi:


integration_tests/macros/unit_tests/test_get_relation_max_name_length.sql
```diff
@@ -0,0 +1,9 @@
+{% macro test_get_relation_max_name_length() %}
```
Maayan-s, line 1: don't you prefer to do it as different macros and adapter.dispatch?

integration_tests/macros/unit_tests/test_get_relation_max_name_length.sql
```diff
@@ -0,0 +1,9 @@
...
+    {% if target.type == 'redshift'%}
+        {{ assert_value(elementary.get_relation_max_name_length(), 127) }}
```
Maayan-s, line 3: Is it hardcoded in dbt as well? isn't there an adapter.max_name_length or something?

macros/utils/table_operations/get_relation_max_length.sql
```diff
@@ -0,0 +1,16 @@
...
+    {{ return(adapter.dispatch('get_relation_max_name_length', 'elementary')()) }}
+{% endmacro %}
+
+{# Snowflake and BigQuery #}
+{% macro default__get_relation_max_name_length(temporary, relation, sql_query) %}
+    {{ return(none) }}
+{% endmacro %}
+
+{% macro redshift__get_relation_max_name_length(temporary, relation, sql_query) %}
+    {{ return(127) }}
```
Maayan-s, line 11: Is it hardcoded in dbt as well? 
Just asking so we could depend on the adapter and not maintain this

Maayan-s (commented):
LGTM, just one small comment but not critical as this is a temporary solution

PR #30, opened 2022-06-07
Backend

oravi:


models/edr/alerts/alerts_data_monitoring.sql
```diff
@@ -1,11 +1,37 @@
 {{
   config(
-    materialized = 'incremental',
-    unique_key = 'alert_id',
-    on_schema_change = 'append_new_columns'
+    materialized = 'view'
```
Maayan-s, line 3: @oravi 
My only concern here is that the users already have these models as incremental tables...
In our tests we clear the schema, but for them, this will fail... I think.

macros/utils/graph/get_model_database_and_schema.sql
```diff
@@ -0,0 +1,15 @@
+{%- macro get_model_database_and_schema(package_name, model_name) -%}
```
Maayan-s, line 1: Does this work for models that are not from a package as well?
It is unclear. 
If so - can you add a comment and change the order of params?
If not - the name is not indicative.

Maayan-s (commented):
Two small comments

PR #31, opened 2022-06-09
Fix none timebase anomalies bug

IDoneShaveIt:
1. Fixed the integration_tests to actually test table anomalies without any timestamp column 
2. Added debug / no_timestamp tests modes (improved the e2e run to actually create data for no_timestamp tests)
3. Fixed the anomaly query to use bucket_end instead of bucket_start


dbt_project.yml
```diff
@@ -1,10 +1,10 @@
-name: 'elementary'
```
IDoneShaveIt, line 1: formatting

integration_tests/dbt_project.yml
```diff
@@ -1,7 +1,7 @@
-name: 'elementary_integration_tests'
```
IDoneShaveIt, line 1: formatting

integration_tests/run_e2e_tests.py
```diff
@@ -189,8 +204,21 @@ def e2e_tests(target, test_types):
...
     dbt_runner.run()
 
+    if 'debug' in test_types:
+        dbt_runner.test(select='tag:debug')
+        return [table_test_results, string_column_anomalies_test_results, numeric_column_anomalies_test_results,
+                    any_type_column_anomalies_test_results, schema_changes_test_results, regular_test_results,
+                    artifacts_results]
+    
+    if 'no_timestamp' in test_types:
+        dbt_runner.test(select='tag:no_timestamp')
```
oravi, line 217: There is no validation here, see how we validate other types of tests

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -1,6 +1,6 @@
 {% macro table_monitoring_query(monitored_table_relation, timestamp_column, is_timestamp, min_bucket_start, table_monitors, freshness_column=none) %}
 
-    {%- set max_bucket_end = "'"~ run_started_at.strftime("%Y-%m-%d 00:00:00")~"'" %}
+    {%- set max_bucket_end = "'"~ elementary.get_run_started_at().strftime("%Y-%m-%d 00:00:00")~"'" %}
```
Maayan-s, line 3: Should we do this change for the column_monitoring_query as well?
If not it means we can't tests it on the e2e

integration_tests/models/any_type_column_anomalies_no_timestamp.sql
```diff
@@ -0,0 +1,37 @@
+-- depends_on: {{ ref('any_type_column_anomalies_validation') }}
```
oravi, line 1: cosmetics - I'd call this model - no_timestamp_table_anomalies
Also - currently you only validate table_anomalies and there is no coverage for column anomalies so maybe worth adding a new fake data instead of relying on  any_type_column_anomalies_training.
I'd create a new fake training file that contains X rows and the validation file will contain X * 2. In terms of the columns you can have one column per column type (number, string, any type) and simulate one anomaly per type so we will have sanity also when there is no timestamp. WDYT?

integration_tests/macros/system/log_config_var.sql
```diff
@@ -0,0 +1,6 @@
+{# Loging the wanted config var as an elementary log (using elementary.edr_log) #}
+{# The dbtRunner catch this log when executed with run_operation #}
+{# This is used for accessing the integration tests vars #}
+{% macro log_config_var(var_name) %}
```
oravi, line 4: cosmetics - I'd call it return_config_var (the log mechanism is our way to return it back to the python)

integration_tests/models/any_type_column_anomalies_no_timestamp.sql
Maayan-s, line 1: This model has data for testing anomalies in columns as well. 
I think it's a small effort to add the column anomalies where there is no TS to e2e as well.

integration_tests/run_e2e_tests.py
```diff
@@ -181,6 +186,16 @@ def e2e_tests(target, test_types):
...
                     any_type_column_anomalies_test_results, schema_changes_test_results, regular_test_results,
                     artifacts_results]
+    
+    # Creates row_count metrics for anomalies detection.
+    if 'no_timestamp' in test_types:
+        current_time = datetime.now()
+        # run operation returns the operation value as a list of strings.
+        # so we we convert the days_back value into int.
+        days_back_project_var = int(dbt_runner.run_operation(macro_name="log_config_var", macro_args={"var_name": "days_back"})[0])
+        for run_index in range(days_back_project_var):
```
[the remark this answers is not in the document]
IDoneShaveIt: Well the validation is running afterward and override todays data.
I can do it with offset just to save one run (which is a good thing), but the logic is valid even like this :)

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
oravi: What do you think about doing it for all the run_started_at usages?

macros/edr/data_monitoring/data_monitors_configuration/get_buckets_configuration.sql
```diff
@@ -8,6 +8,18 @@
     {{ return(global_min_bucket_start) }}
 {% endmacro %}
 
+{# bucket_end represents the end of the bucket, so we need to add extra day to the timedelta #}
+{% macro get_global_min_bucket_end() %}
```
Maayan-s, line 12: I don't think this macro is in use, right?
If not remove it. 
If it is - I would make 'get_global_min_bucket_end(String=False)', returnd datetime by default and String if String=True

integration_tests/macros/e2e_tests/tests_validation.sql
```diff
@@ -167,6 +167,48 @@
                                                  'null_percent_bool']) }}
 {% endmacro %}
 
+{% macro validate_no_timestamp_anomalies() %}
+    {%- set max_bucket_end = "'"~ run_started_at.strftime("%Y-%m-%d 00:00:00")~"'" %}
+    {% set alerts_relation = get_alerts_table_relation('alerts_data_monitoring') %}
+
+    {# Validating row count for no timestamp table anomaly #}
+    {% set no_timestamp_row_count_validation_query %}
+        select distinct table_name
```
oravi, line 176: I think that if you run all the tests together this will fail

Maayan-s (approved):

PR #36, opened 2022-06-18
Query comment

Maayan-s:
- Model named 'information_schema_query_history' is a view of query history from all schemas with models. 
- Query comment is added using macro and extracted from query text using 'substr_between_two_strings' macro.
(This works only for Snowflake, but anything platform-specific is called using macros, so we can add later)

@jtalmi

### DIFF
--- a/macros/edr/system/configuration/get_schemas_for_tests_from_graph.sql
+++ b/macros/edr/system/configuration/get_schemas_for_tests_from_graph.sql
@@ -1,4 +1,4 @@
-{% macro get_configured_schemas_from_graph() %}
+{% macro get_schemas_for_tests_from_graph() %}
     {% set configured_schemas = [] %}
     {% if execute %}
         {% for test_node in graph.nodes.values() | selectattr('resource_type', '==', 'test') %}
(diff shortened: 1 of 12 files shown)


macros/edr/system/configuration/get_schemas_for_tests_from_graph.sql
```diff
@@ -1,4 +1,4 @@
-{% macro get_configured_schemas_from_graph() %}
+{% macro get_schemas_for_tests_from_graph() %}
```
Maayan-s, line 1: Just name change, irrelevant to the query history

macros/edr/system/system_utils/full_names.sql
```diff
@@ -64,21 +64,4 @@
 {% macro relation_to_full_name(relation) %}
```
Maayan-s, line 64: Moved the macro to more relevant place, irrelevant to the query history

models/edr/data_monitoring/schema_changes/table_changes.sql
```diff
@@ -37,7 +37,7 @@ table_removed as (
     from pre left join cur
```
Maayan-s, line 37: Just name change, irrelevant to the query history

models/edr/metadata_store/filtered_information_schema_tables.sql
```diff
@@ -4,7 +4,7 @@
   )
 }}
 
-{% set configured_schemas = elementary.get_configured_schemas_from_graph() %}
+{% set configured_schemas = elementary.get_schemas_for_tests_from_graph() %}
```
Maayan-s, line 8: Just name change, irrelevant to the query history

PR #37, opened 2022-06-19
Added better handling for singular tests

oravi:


macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -81,22 +81,26 @@
...
+                {% if split_parent_model_unique_id and split_parent_model_unique_id | length > 0 %}
+                    {% set parent_model_name = split_parent_model_unique_id[-1] %}
+                    {% if parent_model_name and parent_model_name in test_model_jinja %}
+                        {% do primary_parent_model_candidates.append(parent_model_unique_id) %}
+                    {% endif %}
                 {% endif %}
+            {% endfor %}
+            {% if primary_parent_model_candidates | length == 1 %}
+                {% set primary_parent_model_id = primary_parent_model_candidates[0] %}
             {% endif %}
```
Maayan-s, line 101: Just to make sure I got this right:
1. If there is a parent model = parent model
2. If there isn't you check the list of refs, and if it has only one = parent model
3. else (none or more than 1) - no parent model?

Maayan-s (commented):
Looks good, let's talk before you publish, I want to suggest something

macros/edr/dbt_artifacts/upload_dbt_tests.sql
oravi: no the logic is - 
1. if there is one parent model (depends on nodes for the test within the current package) just return it
2. if there is more than one parent model, try to find the model this test is defined on using a hint in the model jinja inside the test node
3. if couldn't found - no parent

Maayan-s: ok, got it. 
maybe add these as comments?
the names are confusing as they are all alike, so not very indicative of the flow

oravi: done

PR #39, opened 2022-06-29
Enable to specify the destination database and schema for artifacts

yu-iskw:
Solves https://github.com/elementary-data/elementary/issues/98

## Overview of the implementation
We introduce new two dbt variables to determine the destination database and schema for artifacts, instead of using the database and schema of the first node whose `package_name` is `elementary`.


oravi:
Hi @yu-iskw ! great to see your contributions, thanks so much for doing this. See my comment in [the original issue](https://github.com/elementary-data/elementary/issues/96). Waiting to hear from you before diving in to this PR, thanks!

yu-iskw:
@oravi Thank you for the comment. I totally misunderstood the current implementation. We might still have a point to discuss, but the PR is no longer required. So, let's close. Thank you very much.

PR #40, opened 2022-06-29
Added a flag to ignore test warnings in alert views

oravi:


IDoneShaveIt (approved):
LGTM but it bothers me that we have 3 models that looks the same except of a `test_type` we can create a macro for it, but we can do it later.

PR #52, opened 2022-07-11
Update anomaly_threshold_sensitivity model to have stddev value

IrinaSel:
anomaly_threshold_sensitivity model has average value, but not standard deviation which is used for z-score calculation. Adding it for better visibility.


Maayan-s (approved):
Congrats @IrinaSel for your first contribution!! 🎉 
We are honored 🙇‍♀️ 🙇‍♀️

PR #54, opened 2022-07-17
Made the model alerts configurable using vars.

elongl:


macros/edr/dbt_artifacts/upload_run_results.sql
```diff
@@ -1,12 +1,11 @@
 {% macro upload_run_results(results) %}
+    {% if elementary.get_config_var('disable_run_results') %}
```
elongl, line 2: I personally don't think the user should be able to configure this as it's a technical implementation detail that should be abstracted away from the user, rather than a setting such as `disable_model_alert` or `elementary_alerts_ignore_warn` that is relevant to him.

What's your take on it? @oravi

models/edr/alerts/failed_tests.sql
```diff
@@ -0,0 +1,10 @@
+with all_test_alerts as (
+     select * from {{ ref('elementary', 'alerts_schema_changes') }}
+     union all
+     select * from {{ ref('elementary', 'alerts_anomaly_detection') }}
+     union all
+     select * from {{ ref('elementary', 'alerts_dbt_tests') }}
+)
+
+select * from all_test_alerts
+where {{ not elementary.get_config_var('disable_test_alerts') }}
```
elongl, line 10: I've created this unified model because it allows a _root_ model to disable all the test alerts from.  
However, this poses a problem, if the user decides to re-enable a test type, it could potentially send lots of alerts. As I see it there are a couple of options:

1. Let it happen. It's not necessarily a bad thing.
2. Store the disablement time on the Elementary's schema.
3. Exclude the data at an earlier stage than the `alerts` models so that they just return empty "naturally".

What do you think? @oravi @IDoneShaveIt @Maayan-s

PR #55, opened 2022-07-17
Non test alerts

elongl:


macros/edr/dbt_artifacts/upload_run_results.sql
```diff
@@ -1,12 +1,11 @@
 {% macro upload_run_results(results) %}
+    {% if elementary.get_config_var('disable_run_results') %}
```
elongl, line 2: I personally don't think the user should be able to configure this as it's a technical implementation detail that should be abstracted away from the user, rather than a setting such as disable_model_alert or elementary_alerts_ignore_warn that is relevant to him.

What's your take on it? @oravi

PR #58, opened 2022-07-20
Support alias

civitaspo:
Signed-off-by: civitaspo <civitaspo@gmail.com>

Hi maintainers, thank you for the great product.

I found a bug that `elementary.get_model_graph_node` returns `None` when using [`alias`](https://docs.getdbt.com/docs/building-a-dbt-project/building-models/using-custom-aliases). `model_relation.name` is `alias` when using [`alias`](https://docs.getdbt.com/docs/building-a-dbt-project/building-models/using-custom-aliases), not model name. So, we should use model unique_id when search the node in `elementary.get_model_graph_node`.


macros/edr/tests/test_utils/get_model_graph_node.sql
```diff
@@ -4,11 +4,11 @@
         {% set test_graph_node = model %}
         {% set test_depends_on_unique_ids = test_graph_node.depends_on.nodes %}
         {# model relation is the relation object of the model where the test is defined #}
-        {% set model_name = model_relation.name | lower %}
+        {% set model_node = elementary.get_node_by_relation(model_relation) %}
         {% set depends_on_nodes = elementary.get_nodes_by_unique_ids(test_depends_on_unique_ids) %}
         {% if depends_on_nodes %}
             {% for node in depends_on_nodes %}
-                {% if node.name | lower == model_name %}
+                {% if node.unique_id == model_node.unique_id %}
```
oravi, line 11: Here we are already iterating over the depends_on_nodes list and compare each depends on node name to the model relation name. I would suggest adding here additional check that will validate if either the name is matched or the alias of the node is matched. For example, it could be something like - 
image

(Please note that alias does not always exist in a dbt graph node so this is why I used 'get' when accessing it)

oravi:
Hi @civitaspo! I successfully reproduced the issue and I can confirm it's a bug. Thanks so much for opening a PR! I also reviewed it and added some small comments. Happy to help if you have any questions about it! 👍

civitaspo:
Thank you for the review. I applied your suggestion in 945e4cb, so plz review again.

PR #59, opened 2022-07-20
Non test alerts

elongl:


macros/edr/dbt_artifacts/upload_run_results.sql
```diff
@@ -1,12 +1,11 @@
 {% macro upload_run_results(results) %}
+    {% if elementary.get_config_var('disable_run_results') %}
```
elongl, line 2: Do you think a user should be able to disable it? @oravi

models/edr/alerts/error_models.sql
```diff
@@ -0,0 +1,23 @@
...
+       path,
+       original_path,
+       schema_name,
+       message,
+       owner as owners,
+       tags,
+       alias,
+       status,
+       full_refresh
+from {{ ref('model_run_results') }}
```
oravi, line 22: Let's be explicit here and add 'elementary' to the ref? it's just that model_run_results is a name I think other artifact packages can use as well

models/edr/alerts/error_models.sql
```diff
@@ -0,0 +1,23 @@
...
+where {{ not elementary.get_config_var('disable_model_alerts') }} and lower(status) != 'success'
```
oravi, line 23: Please verify that this will be treated as boolean on all platforms, you can maybe use the jinja filter as_bool and then cast it to bool using elementary.type_bool()

models/edr/alerts/error_models.sql
```diff
@@ -0,0 +1,23 @@
+{{
```
oravi, line 1: cosmetics - al the alert views start with alerts so maybe for the sake of consistency let's rename to alerts_dbt_models or alerts_error_models?

models/edr/alerts/error_models.sql
```diff
@@ -0,0 +1,23 @@
...
+  config(
+    materialized = 'view',
+    bind=False
+  )
+}}
+
+select model_execution_id as alert_id,
```
oravi, line 8: The dbt way is to define a cte with 'with' statement to select the entire model_run_results, then the next CTE reads it and filters the result and the last one simply select everything (see alerts_dbt_tests for an example). This will make it easier for our users to read it

models/edr/alerts/failed_tests.sql
```diff
@@ -0,0 +1,28 @@
+{{
```
oravi, line 1: Why do we need this view here? this can be in the CLI in the alerts model no? I think this also means new CLI version won't work with old package version

models/edr/alerts/failed_tests.sql
```diff
@@ -0,0 +1,28 @@
...
+        select * from {{ anomaly_detection_relation }}
+    {% endif %}
+
+    {% if data_monitoring_relation %}
+        union all
+        select * from {{ data_monitoring_relation }}
+    {% endif %}
+)
+
+select * from failed_tests where {{ not elementary.get_config_var('disable_test_alerts') }}
```
oravi, line 28: Same comment on a boolean value cast

models/edr/alerts/failed_tests.sql
```diff
@@ -0,0 +1,28 @@
...
+{% set anomaly_detection_relation = adapter.get_relation(this.database, this.schema, 'alerts_anomaly_detection') %}
+// Backwards compatibility support for a renamed model.
+{% set data_monitoring_relation = adapter.get_relation(this.database, this.schema, 'alerts_data_monitoring') %}
+
+with failed_tests as (
+     select * from {{ ref('elementary', 'alerts_schema_changes') }}
+     union all
+     select * from {{ ref('elementary', 'alerts_dbt_tests') }}
+
+    {% if anomaly_detection_relation %}
```
oravi, line 17: Let's check if the relation_exists to double check we can query it and it's not cached or anything like that?

models/edr/dbt_artifacts/alerts/alerts_dbt_models.sql
```diff
@@ -1,30 +0,0 @@
-{{
```
oravi, line 1: This is a comment for the folders change above, why do we need this extra layer of hierarchy?

models/edr/run_results/model_run_results.sql
```diff
@@ -1,3 +1,10 @@
+{{
```
oravi, line 1: This is a comment for the folders change above - why elementary_test_results is now in run_results?

models/edr/alerts/error_models.sql
elongl: Yeah, I renamed to `alerts_dbt_models` but it kind of bothers me that the difference between `alerts_dbt_models` and `alerts_models` really isn't clear from their naming.

models/edr/alerts/failed_tests.sql
elongl: Basically the same as [this](https://github.com/elementary-data/elementary/pull/137/files#r926585866).

models/edr/alerts/failed_tests.sql
elongl: It [seems](https://github.com/dbt-labs/dbt-core/blob/8442fb66a591c4f5353f8baaaaf62f846c5f5171/core/dbt/include/global_project/macros/adapters/relation.sql#L70) that `load_relation()` is just a user-friendly way of invoking `get_relation()`.  
Or in their words _"A convenience wrapper for get_relation."_
Could you please elaborate a bit on why do it?

macros/edr/dbt_artifacts/upload_run_results.sql
```diff
@@ -1,12 +1,11 @@
...
+        {% do elementary.edr_log("Run results are disabled, skipping upload.") %}
```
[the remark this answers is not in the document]
Maayan-s: I think we want to do it because once the on-run-end hook is not added by the user, we want to communicate to him that it exists and runs and works. Also if it fails, we want them to understand that the failure was part of that process. 
@oravi @elongl

macros/edr/hooks/on_run_end.sql
```diff
@@ -1,4 +1,4 @@
 {% macro on_run_end() %}
```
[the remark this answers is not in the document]
Maayan-s: I suggested a single on-run-end, based on past feedbacks. @oravi

models/edr/alerts/error_models.sql
elongl: I've tested this an it works as expected.
Basically if you'd provide falsely values (`false`, `0`, etc.) it treats them as such, and everything else would be truly.
We also this in [elementary_alerts_ignore_warn](https://github.com/elementary-data/dbt-data-reliability/blob/master/models/edr/alerts/alerts_dbt_tests.sql#L35).

macros/edr/dbt_artifacts/upload_run_results.sql
```diff
@@ -1,12 +1,11 @@
...
     {% set edr_cli_run = elementary.get_config_var('edr_cli_run') %}
     {% if execute and not edr_cli_run and results %}
-        {% if 'model.elementary.dbt_run_results' not in graph.nodes %}
-            {% do elementary.edr_log("dbt_run_results is disabled, not uploading the run results.") %}
+        {% if elementary.get_config_var('disable_run_results') %}
+            {% do elementary.edr_log("Run results are disabled, skipping upload.") %}
             {{ return('') }}
         {% endif %}
-
+        {{ elementary.edr_log("Uploading run results.") }}
```
oravi, line 8: Make it debug?

macros/edr/hooks/on_run_end.sql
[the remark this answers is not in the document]
elongl: And as we now see, it does not affect functionality.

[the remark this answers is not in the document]
oravi: @Maayan-s which feedback? @elongl if you put the first line and "AAAAA" in a macro the entire on-run-end will fail instead of at least trying to run them separately, The handle_test_results is more important than the upload_run_results so we can place the handle_test_results first. Then also when they provide us the log we can know that at least this worked and the upload_run_result didn't

models/edr/alerts/error_models.sql
oravi: Did you test it on all platforms? and it's not true, in [here](https://github.com/elementary-data/dbt-data-reliability/blob/master/models/edr/alerts/alerts_dbt_tests.sql#L35) we don't add a bool condition with Jinja. We use the jinja to add a regular SQL condition

models/edr/dbt_artifacts/alerts/alerts_dbt_models.sql
elongl: This file was deleted.

models/edr/alerts/failed_tests.sql
oravi: Sounds good, resolving

macros/edr/hooks/on_run_end.sql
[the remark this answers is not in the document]
elongl: _if you put the first line and "AAAAA" in a macro the entire on-run-end will fail instead of at least trying to run them separately._ Hmm, I think I just showed that this is the case for both. The `on-run-end` runs sequentially and it'll stop once there's an error regardless whether it's within the hook's definition or within one of the macros that it invokes.

And I'll move `handle_test_results` to be first.

oravi (approved):

PR #66, opened 2022-07-27
Not specifying the package ref within the package itself.

elongl:


IDoneShaveIt (approved):
LGTM.
Waiting for @oravi comments as well :)

PR #69, opened 2022-08-05
Databricks support for dbt artifacts and results

Maayan-s:


macros/utils/cross_db_utils/to_char.sql
```diff
@@ -2,14 +2,15 @@
     {{ return(adapter.dispatch('to_char', 'elementary')(column, format)) }}
 {%- endmacro %}
 
+// Snowflake and Redshift
```
elongl, line 5: Please use `{# ... #}`, this syntax didn't seem to work in the past.

elongl (commented):
Looks great! Left one tiny comments.
Also, we now need to update the documentation in `docs` as well.

PR #70, opened 2022-08-08
Added a macro that prints the Elementary profile structure.

elongl:
Upon installation, the user runs `dbt run-operation elementary.print_elementary_profile` to get the `elementary` profile.

### DIFF
--- a/macros/edr/system/configuration/print_elementary_cli_profile.sql
+++ b/macros/edr/system/configuration/print_elementary_cli_profile.sql
@@ -50,6 +50,9 @@ elementary:
       type: {{ target.type }}
       host: {{ target.host }}
       http_path: {{ target.http_path }}
+      {%- if elementary_database %}
+      catalog: {{ elementary_database }}
+      {% endif %}
       schema: {{ elementary_schema }}
       token: <TOKEN>
       threads: {{ target.threads }}
--- a/macros/edr/system/configuration/print_elementary_cli_profile.sql
+++ b/macros/edr/system/configuration/print_elementary_cli_profile.sql
@@ -1,5 +1,5 @@
 {% macro print_elementary_cli_profile() %}
-  {% set elementary_database, elementary_schema = get_package_database_and_schema() %}
+  {% set elementary_database, elementary_schema = elementary.get_package_database_and_schema() %}
   {{ log('\n' ~ adapter.dispatch('print_elementary_profile')(elementary_database, elementary_schema), info=True) }}
 {% endmacro %}
 
--- a/macros/edr/system/configuration/print_elementary_cli_profile.sql
+++ b/macros/edr/system/configuration/print_elementary_cli_profile.sql
@@ -1,9 +1,9 @@
 {% macro print_elementary_cli_profile() %}
   {% set elementary_database, elementary_schema = elementary.get_package_database_and_schema() %}
-  {{ log('\n' ~ adapter.dispatch('print_elementary_profile')(elementary_database, elementary_schema), info=True) }}
+  {{ log('\n' ~ adapter.dispatch('print_elementary_cli_profile')(elementary_database, elementary_schema), info=True) }}
 {% endmacro %}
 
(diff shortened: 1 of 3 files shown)


macros/edr/system/configuration/print_elementary_cli_profile.sql
```diff
@@ -0,0 +1,60 @@
+{% macro print_elementary_cli_profile() %}
+    {% set elementary_database, elementary_schema = get_package_database_and_schema() %}
```
oravi, line 2: Use elementary prefix

macros/edr/system/configuration/print_elementary_cli_profile.sql
```diff
@@ -0,0 +1,60 @@
+{% macro print_elementary_cli_profile() %}
+  {% set elementary_database, elementary_schema = elementary.get_package_database_and_schema() %}
+  {{ log('\n' ~ adapter.dispatch('print_elementary_profile')(elementary_database, elementary_schema), info=True) }}
+{% endmacro %}
+
+{% macro snowflake__print_elementary_profile(elementary_database, elementary_schema) %}
```
oravi, line 6: It's also better to have a consistent name with the macro above

Maayan-s (commented):
As it's an operation for users (not an internal function), I would change the name. 
Think about the step in the onboarding:
"After you finish installing the CLI, run the operation 'print_elementary_cli_profile' to generate a connection profile that points to the monitored dbt project." 
This would be better:
"After you finish installing the CLI, run the operation 'generate_elementary_cli_profile', that will be used to point CLI to the monitored dbt project."

Maayan-s:
Another question I had - does it work in dbt cloud?

elongl:
> As it's an operation for users (not an internal function), I would change the name. Think about the step in the onboarding: "After you finish installing the CLI, run the operation 'print_elementary_cli_profile' to generate a connection profile that points to the monitored dbt project." This would be better: "After you finish installing the CLI, run the operation 'generate_elementary_cli_profile', that will be used to point CLI to the monitored dbt project."

What worries me is that calling it `generate` might convey that we create a new one on each invocation.
When I call it `print`, it's clearer that it's solely about printing the Elementary profile back to the user.
What do you think? @Maayan-s

elongl:
> Another question I had - does it work in dbt cloud?

It should, what makes you think that it won't?

elongl:
> Another question I had - does it work in dbt cloud?

No. `profiles.yml` is a Dbt CLI concept.
image

However, the macro itself would still work which would help users to define their profile in Dbt Cloud.

Maayan-s:
> > Another question I had - does it work in dbt cloud?
> 
> No. `profiles.yml` is a Dbt CLI concept.
> 
> image
> 
> However, the macro itself would still work which would help users to define their profile in Dbt Cloud.

Yes I was asking if all these jinja refs are available in cloud although it does not support profiles

Maayan-s:
> > As it's an operation for users (not an internal function), I would change the name. Think about the step in the onboarding: "After you finish installing the CLI, run the operation 'print_elementary_cli_profile' to generate a connection profile that points to the monitored dbt project." This would be better: "After you finish installing the CLI, run the operation 'generate_elementary_cli_profile', that will be used to point CLI to the monitored dbt project."
> 
> What worries me is that calling it `generate` might convey that we generate a new one on each invocation. When I call it `print`, it's clearer that it's solely about printing the Elementary profile back to the user. What do you think? @Maayan-s

We don't print an existing profile, we do generate one. No?
What do you mean by "generate a new one on each invocation"?

elongl:
> > > Another question I had - does it work in dbt cloud?
> > 
> > 
> > No. `profiles.yml` is a Dbt CLI concept.
> > image
> > However, the macro itself would still work which would help users to define their profile in Dbt Cloud.
> 
> Yes I was asking if all these jinja refs are available in cloud although it does not support profiles

Tested, it does work on Dbt cloud.

PR #74, opened 2022-08-11
Fix dimension on all platforms

IDoneShaveIt:


macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -1,7 +1,13 @@
 {% macro dimension_monitoring_query(monitored_table_relation, dimensions, where_expression, timestamp_column, is_timestamp, min_bucket_start) %}
     {% set metric_name = 'dimension' %}
     {%- set max_bucket_end = "'"~ elementary.get_run_started_at().strftime("%Y-%m-%d 00:00:00")~"'" %}
-    {%- set min_bucket_end = "'"~ (modules.datetime.datetime.strptime(min_bucket_start, "'%Y-%m-%d %H:%M:%S'") +  modules.datetime.timedelta(1)).strftime("%Y-%m-%d 00:00:00")~"'" %}
+
+    {# Different platforms returns different time formats, all starts with %Y-%m-%d %H:%M:%S #}
+    {# We use regex to extract the common patern #}
```
Maayan-s, line 6: Why is this required here and wasn't required on the other tests?

IDoneShaveIt: Because on the other tests we didn't needed to create the min_bucket_end variable.
In the dimension anomaly we query all of the dimension values we found in the previous runs, and we create 0 metrics/filter out outdated dimensions - this is why we need the min_bucket_end

Maayan-s: So why not use 'strftime("%Y-%m-%d 00:00:00")' like we do in all other places of the code?

Maayan-s: I don't like the regex solution, prefer to solve with formating like this, ot in the SQL that brings it

Maayan-s:
Approved!

PR #83, opened 2022-08-24
Support dbt utils breaking changes

IDoneShaveIt:
dbt are about to make a breaking change in dbt-utils (deprecating macros and moving them into dbt-core).
In order to use those macros without making our users upgrading their dbt-core/utils, we wrapped those `cross-db-utils` to support all versions.


macros/utils/cross_db_utils/type_bigint.sql
```diff
@@ -0,0 +1,7 @@
+{% macro type_bigint() %}
```
Maayan-s, line 1: We have a data_types folder on our utils, maybe it's a better fit (all the type macros)

macros/utils/cross_db_utils/timediff.sql
```diff
@@ -1,3 +1,4 @@
+{# Same as datediff, but used on timestamps instead #}
```
Maayan-s, line 1: This works for both timestamp and date (fix the comment, it's unclear)

macros/utils/cross_db_utils/timeadd.sql
```diff
@@ -1,3 +1,4 @@
+{# Same as dateadd, but used on timestamps instead #}
```
Maayan-s, line 1: This works for both timestamp and date (fix the comment, it's unclear)

macros/utils/cross_db_utils/current_timestamp.sql
```diff
@@ -0,0 +1,38 @@
+{% macro current_timestamp() -%}
```
Maayan-s, line 1: Is this just copying from dbt utils? 
If so we need to credit them here and add the licensing

macros/utils/cross_db_utils/type_bigint.sql
IDoneShaveIt: I just wanted to make it clear that this is created to support the deprecation of dbt-utils.
But sure I can move it over there..

macros/utils/cross_db_utils/current_timestamp.sql
```diff
@@ -0,0 +1,63 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License. 
+#}
+
+{# 
+    dbt_utils.current_timestamp is using dbt_utils.type_timestamp which is going to be deprecated from dbt-utils version 0.9.0,
+    therefore dbt_utils.current_timestap is going to be deprecated as well (without having a similar macro at dbt-core).
+    We copied this macro, with a small change that supports the deprecation, so we could continue and use it in our project.
+#}
```
Maayan-s, line 21: @oravi do you think this is ok? we give them the copyright and explain why we had to duplicate.

Maayan-s:
LGTM, just want @oravi to say that he is ok with the license and comment

PR #84, opened 2022-08-24
Clear tests use custom schema

elongl:


integration_tests/dbt_project.yml
```diff
@@ -21,5 +21,10 @@ vars:
...
 
 models:
+  elementary_integration_tests:
+    +schema: "{{ env_var('CUSTOM_SCHEMA', 'default') | replace('-', '_') }}"
   elementary:
-    +schema: elementary
\ No newline at end of file
+    +schema: elementary
+
+seeds:
```
elongl, line 29: Not entirely sure that the `seeds` should also use the custom schema, what do you think?  @oravi 
This will cause a lot of duplicated seeds but if seeds change (rarely), it won't override.

PR #86, opened 2022-08-25
Elementary's tests schema name is configurable.

elongl:


oravi (approved):

integration_tests/macros/e2e_tests/clear_tests.sql
```diff
@@ -1,10 +1,16 @@
 {% macro clear_tests() %}
     {% if execute %}
+        {% do drop_schema(elementary.target_database(), target.schema) %}
+
         {% set database_name, schema_name = elementary.get_package_database_and_schema('elementary') %}
         {% do drop_schema(database_name, schema_name) %}
-        {% set schema_name = schema_name ~ '__tests' %}
+
+        {% set tests_schema_name = elementary.get_config_var('tests_schema_name') %}
```
oravi, line 8: maybe we should rename to 'elementary_tests_schema_name'?

elongl: dbt provided the package namespaces so that you won't need to do `<my_package>_<config_var>`, and instead just:

```yml
my_package:
  config_var1: ...
  config_var2: ...
```

All of our other configuration keys don't include the `elementary` keyword and in my opinion, nor should this.

ma1ster:
Hi, is there a chance to make it possible that tests schema is the same as the one where elementary models resides ? I have only one schema in my data warehouse for this

Maayan-s:
Hi @ma1ster, this PR does not allow it as we ran into a dbt caching issue. 
The good news are @elongl already found a solution, and implemented on [this PR](https://github.com/elementary-data/dbt-data-reliability/pull/90).

We will release it on the next version, you can track [in this issue](https://github.com/elementary-data/elementary/issues/285), or on our Slack / social media / releases updates :)

PR #88, opened 2022-08-28
remove copy of dbt-utils macros (current_timestamp)

IDoneShaveIt:
remove copy of dbt-utils macros (current_timestamp)

### DIFF
--- a/macros/utils/cross_db_utils/current_timestamp.sql
+++ b/macros/utils/cross_db_utils/current_timestamp.sql
@@ -1,63 +1,20 @@
-{# 
-   Copyright 2021 dbt Labs, Inc.
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-
-       http://www.apache.org/licenses/LICENSE-2.0
-
-   Unless required by applicable law or agreed to in writing, software
-   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


macros/utils/cross_db_utils/current_timestamp.sql
```diff
@@ -1,63 +1,20 @@
...
-    current_timestamp
-{% endmacro %}
 
-
-{# 
-    dbt_utils.current_timestamp_in_utc is using dbt_utils.type_timestamp and dbt_utils.current_timestamp which are going to be deprecated from dbt-utils version 0.9.0,
-    therefore dbt_utils.current_timestamp_in_utc is going to be deprecated as well (without having a similar macro at dbt-core).
-    We copied this macro, with a small change that supports the deprecation, so we could continue and use it in our project.
-#}
 {% macro current_timestamp_in_utc() -%}
```
Maayan-s, line 12: Why do we need this then?

Maayan-s: Just as prep for the future?

IDoneShaveIt: Yes.
We will have to create this "in between" macro and use it instead of dbt-utils directly.
We already replaced the use of them in the previous PR, so this is nice to have an easy preparation for the future 🙂

Maayan-s: ok, approved

Maayan-s (approved):

PR #89, opened 2022-08-28
Fix no result for anomalies with one day data

IDoneShaveIt:
Fix no result for anomalies with one day data.
This is how it will look at the UI:
image


macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -113,6 +113,7 @@
                 column_name,
                 metric_name,
                 case
+                    when training_stddev is null then null
```
Maayan-s, line 116: Make sure this works for all platforms, we might need to pass null casted as int

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -149,6 +150,13 @@
              {%- endif %}
         order by bucket_end
     {%- endset -%}
+    {% set test_results_description %}
+        {% if elementary.insensitive_get_dict_value(anomaly_dict, 'anomaly_score') is none %}
+            Could not calculate anomaly score for this test due to lack of data.
```
Maayan-s, line 155: "Not enough data to calculate anomaly score."

models/edr/data_monitoring/anomaly_detection/metrics_anomaly_score.sql
```diff
@@ -55,8 +55,9 @@ metrics_anomaly_score as (
         dimension_value,
         metric_name,
         case
-           when training_stddev = 0 then 0
-           else (metric_value - training_avg) / (training_stddev)
+            when training_stddev is null then null
```
Maayan-s, line 58: Same comment about null

Maayan-s (changes requested):
Verify Redshift + change desc

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
IDoneShaveIt: Working on all platforms 🙂

models/edr/data_monitoring/anomaly_detection/metrics_anomaly_score.sql
IDoneShaveIt: Working on all platforms 🙂

PR #91, opened 2022-08-29
Renamed 'elementary_debug_logs' to 'debug_logs'.

elongl:
Keeping consistent with not specifying the `elementary` prefix.


IDoneShaveIt (approved):
LGTM!

PR #94, opened 2022-09-02
Databricks support for anomaly detection tests 

Maayan-s:
This is for anomaly tests, without schema changes test.

### DIFF
--- a/macros/edr/system/system_utils/full_names.sql
+++ b/macros/edr/system/system_utils/full_names.sql
@@ -14,6 +14,24 @@
 {%- endmacro %}
 
 
+{% macro split_full_table_name_to_vars(full_table_name) %}
+    {% set split_full_table_name = full_table_name.split('.') %}
+    {# Databricks full name sometimes is schema.table, no db #}
+    {%- if split_full_table_name | length == 2 %}
+        {% set database_name = None %}
+        {% set schema_name = split_full_table_name[0] %}
+        {% set table_name = split_full_table_name[1] %}
+    {%- else  %}
+        {% set database_name = split_full_table_name[0] %}
+        {% set schema_name = split_full_table_name[1] %}
+        {% set table_name = split_full_table_name[2] %}
+    {%- endif %}
+    {{ return(database_name, schema_name, table_name) }}
+{% endmacro %}
+
+
+
+
 {% macro full_name_split(part_name) %}
     {{ adapter.dispatch('full_name_split','elementary')(part_name) }}
 {% endmacro %}
--- a/macros/edr/system/system_utils/full_names.sql
+++ b/macros/edr/system/system_utils/full_names.sql
@@ -62,7 +62,12 @@
 
 
 {% macro relation_to_full_name(relation) %}
-    {%- set full_table_name = relation.database | upper ~'.'~ relation.schema | upper ~'.'~ relation.identifier | upper %}
+    {%- if relation.database %}
+        {%- set full_table_name = relation.database | upper ~'.'~ relation.schema | upper ~'.'~ relation.identifier | upper %}
+    {%- else %}
+    {# Databricks doesn't always have a database #}
+        {%- set full_table_name = relation.schema | upper ~'.'~ relation.identifier | upper %}
+    {%- endif %}
     {{ return(full_table_name) }}
 {% endmacro %}
 
--- a/macros/edr/system/system_utils/full_names.sql
+++ b/macros/edr/system/system_utils/full_names.sql
@@ -26,7 +26,7 @@
         {% set schema_name = split_full_table_name[1] %}
         {% set table_name = split_full_table_name[2] %}
     {%- endif %}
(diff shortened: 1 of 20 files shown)


macros/edr/system/system_utils/full_names.sql
```diff
@@ -62,7 +62,12 @@
 
 
 {% macro relation_to_full_name(relation) %}
-    {%- set full_table_name = relation.database | upper ~'.'~ relation.schema | upper ~'.'~ relation.identifier | upper %}
+    {%- if relation.database %}
+        {%- set full_table_name = relation.database | upper ~'.'~ relation.schema | upper ~'.'~ relation.identifier | upper %}
+    {%- else %}
+    {# Databricks doesn't always have a database #}
```
IDoneShaveIt, line 68: There are cases when it does have database?
If no, then wouldn't it be better to use adapter here instead of if statement?

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -100,9 +100,16 @@
 {% macro get_metric_test_result(run_result_dict, anomaly_dict, test_node, test_anomaly_scores_table) %}
     {% set full_table_name = elementary.insensitive_get_dict_value(anomaly_dict, 'full_table_name', '') %}
     {% set split_full_table_name = full_table_name.split('.') %}
-    {% set database_name = split_full_table_name[0] %}
-    {% set schema_name = split_full_table_name[1] %}
-    {% set table_name = split_full_table_name[2] %}
+    {# Databricks full name is schema.table, no db #}
```
IDoneShaveIt, line 103: This feels like it should be an adapter macro 🤔 
What if another WH won't support schemas? Then we will need to add another if statement which implies that an adapter macro will fit good here IMO

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -205,9 +212,15 @@
     {% endif %}
     {% set full_table_name = elementary.insensitive_get_dict_value(recent_anomaly_sample, 'full_table_name', '') %}
     {% set split_full_table_name = full_table_name.split('.') %}
-    {% set database_name = split_full_table_name[0] %}
-    {% set schema_name = split_full_table_name[1] %}
-    {% set table_name = split_full_table_name[2] %}
+    {%- if split_full_table_name | length == 2 %}
```
IDoneShaveIt, line 215: See previous comment.
Anyway making a macro that splits the full_table_name (even if not an adapter macro) is needed (duplicating code) 🙂

macros/utils/cross_db_utils/safe_cast.sql
```diff
@@ -1,5 +1,7 @@
 {% macro safe_cast(field, type) %}
-    {% if dbt_version >= '1.2.0' %}
+    {% if target.type == 'databricks' %}
```
IDoneShaveIt, line 2: Mmmm WDYT about creating an adapter for safe cast that by default uses this safe_cast (versioning support) and for databricks returns `try_cast({{field}} as {{type}})`?

macros/utils/data_types/data_type.sql
```diff
@@ -41,6 +41,7 @@
 {%- endmacro -%}
 
 {%- macro default__type_long_string() -%}
+    {# Snowflake, Bigquery, Databricks #}
```
IDoneShaveIt, line 44: NIT: Redundant (Default if for everything that is not specified) 🙂

macros/utils/table_operations/merge_sql.sql
```diff
@@ -2,7 +2,7 @@
     {{ return(adapter.dispatch('merge_sql', 'elementary')(target_relation, tmp_relation, unique_key, dest_columns)) }}
 {%- endmacro %}
 
-{# Snowflake and Bigquery #}
+{# Snowflake, Bigquery and Databricks #}
```
IDoneShaveIt, line 5: NIT: Redundant (Default if for everything that is not specified) 🙂

IDoneShaveIt (commented):
Looking good 😎 !
Left some small comments

macros/edr/system/system_utils/full_names.sql
Maayan-s: There are cases when Databricks has a database, depending on user configuration.
Also wanted to be more platform agnostic thinking of future integrations.

macros/edr/tests/on_run_end/handle_tests_results.sql
Maayan-s: The thing is adapter would limit us to config per integration, and Databricks might have a database name.

macros/edr/tests/on_run_end/handle_tests_results.sql
Maayan-s: done

IDoneShaveIt (approved):
LGTM

PR #95, opened 2022-09-05
artifacts timing - replace run start time with the current time utc

IDoneShaveIt:
replaced run start time with the current time utc

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_exposures.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_exposures.sql
@@ -58,7 +58,7 @@
         'package_name': node_dict.get('package_name'),
         'original_path': node_dict.get('original_file_path'),
         'path': node_dict.get('path'),
-        'generated_at': elementary.current_timestamp_utc_as_string()
+        'generated_at': elementary.datetime_now_utc_as_string()
       }%}
     {{ return(flatten_exposure_metadata_dict) }}
 {% endmacro %}
--- a/macros/edr/dbt_artifacts/upload_dbt_exposures.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_exposures.sql
@@ -58,7 +58,7 @@
         'package_name': node_dict.get('package_name'),
         'original_path': node_dict.get('original_file_path'),
         'path': node_dict.get('path'),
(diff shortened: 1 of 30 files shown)


macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -58,7 +58,7 @@
         'package_name': node_dict.get('package_name'),
         'original_path': node_dict.get('original_file_path'),
         'path': node_dict.get('path'),
-        'generated_at': elementary.get_run_started_at().strftime('%Y-%m-%d %H:%M:%S')
+        'generated_at': modules.datetime.datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S')
```
oravi, line 61: Maybe put this in a macro - 'current_timestamp_utc_as_string' (you can even get the format and set it to default '%Y-%m-%d %H:%M:%S'). The 'formatted_current_timestamp_column' could be current_timestamp_utc_as_timestamp.

IDoneShaveIt: Moved the modules datetime to a macro.
Didn't rename the formatted_current_timestamp_column macro because this name describes it better.
This is a current timestamp column that is formatted. this is not a current timestamp as timestamp.

macros/edr/system/system_utils/timestamp_column.sql
```diff
@@ -4,4 +4,8 @@
 
 {% macro current_timestamp_column() %}
     cast ({{elementary.current_timestamp_in_utc()}} as {{ elementary.type_timestamp() }})
-{% endmacro %}
\ No newline at end of file
+{% endmacro %}
+
+{% macro formatted_current_timestamp_column() %}
```
oravi, line 9: Why not renaming this as well and moving it to times.sql?

IDoneShaveIt: Because this method is specific for creating columns and use inside of a query (just like run_start_column and current_timestamp_column).
It is not a generic time macro that is used to get the current time as datetime/str.

This is why the name fits.
It is current_timestamp_column with a format.

oravi (approved):

PR #96, opened 2022-09-12
handle updated tests

IDoneShaveIt:
Created test results view that joins between `dbt_tests` and `elementary_test_results` tables, causing it show only the test results of the current project tests

### DIFF
--- a/models/edr/run_results/test_results.sql
+++ b/models/edr/run_results/test_results.sql
@@ -1,0 +1,49 @@
+{{
+  config(
+    materialized = 'view',
+    bind=False
+  )
+}}
+
+with elementary_test_results as (
+    select * from {{ ref('elementary_test_results') }}
+),
+
+dbt_tests as (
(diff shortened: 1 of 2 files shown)


models/edr/run_results/test_results.sql
```diff
@@ -0,0 +1,49 @@
...
+
+with elementary_test_results as (
+    select * from {{ ref('elementary_test_results') }}
+),
+
+dbt_tests as (
+    select * from {{ ref('dbt_tests') }}
+),
+
+firt_time_test_occurred as (
```
elongl, line 16: Typo

models/edr/run_results/test_results.sql
```diff
@@ -0,0 +1,49 @@
+{{
```
elongl, line 1: Choose a different name for this model that conveys on the fact that it only applies to currently existing tests (like "latest", "current") and use that name throughout.

elongl (approved):

PR #97, opened 2022-09-12
Artifacts improvements

IDoneShaveIt:
- Made long string type configurable (config the size of a long field column)
- Added a validation on compiled sql - if it is bigger than the long string size, we return "Query is too long" instead.
- Added failures to elementary run results and elementary test results tables - will be used instead of parsing the message.


macros/edr/system/system_utils/empty_table.sql
```diff
@@ -1,5 +1,5 @@
...
+    {{ elementary.empty_table([('id','long_string'),('data_issue_id','string'),('test_execution_id','long_string'),('test_unique_id','long_string'),('model_unique_id','long_string'),('detected_at','timestamp'),('database_name','string'),('schema_name','string'),('table_name','string'),('column_name','string'),('test_type','string'),('test_sub_type','string'),('test_results_description','long_string'),('owners','string'),('tags','string'),('test_results_query','long_string'),('other','string'),('test_name','long_string'),('test_params','long_string'),('severity','string'),('status','string'),('failures', 'bigint')]) }}
```
Maayan-s, line 2: What about compiled_sql?

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -28,7 +28,8 @@
       'column_string': ['min_length', 'max_length', 'average_length', 'missing_count', 'missing_percent'],
       'column_numeric': ['min', 'max', 'zero_count', 'zero_percent', 'average', 'standard_deviation', 'variance']
     },
-    'time_format': '%Y-%m-%d %H:%M:%S'
+    'time_format': '%Y-%m-%d %H:%M:%S',
+    'long_string_size': 16384
```
Maayan-s, line 32: Add to docs

macros/utils/graph/get_compiled_sql_from_node.sql
```diff
@@ -0,0 +1,9 @@
+{% macro get_compiled_sql_from_node(node) %}
```
Maayan-s, line 1: Shouldn't this be generic for any long string value we retrieve? 
Now that the max size is configurable, users can reduce it and values other than 'compiled_query' will fail as well. 
Maybe it's premature and we should do it on another PR, but I think it's easy. No?

Maayan-s (commented):
LMK what you think of my suggestion on long string

macros/utils/graph/get_compiled_sql_from_node.sql
IDoneShaveIt: I agree with you, but I think that maybe it should be another PR 🤔 
Anyhow, as long as we don't document it, it is not really configurable and have less of a chance to happen.

Let me see how much effort it will be to replace every place that uses long string

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -308,6 +313,8 @@
     {%- else -%}
         {% set test_name = test_long_name %}
     {%- endif -%}
+    {# Becasue we use compiled sql for fetching data sampling, we currently not using the macro get_model_sql_from_node #}
```
oravi, line 316: I think this comment is not relevant as the macro is called `get_model_sql_from_node` and here it's a test

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -275,23 +276,27 @@
...
-        'status': elementary.insensitive_get_dict_value(run_result_dict, 'status')
+        'status': elementary.insensitive_get_dict_value(run_result_dict, 'status'),
+        'failures': elementary.insensitive_get_dict_value(run_result_dict, 'failures')
     } %}
     {{ return(test_result_dict) }}
 {% endmacro %}
 
 {% macro get_schema_change_test_result(run_result_dict, schema_change_dict, test_node) %}
     {% set test_result_dict = schema_change_dict %}
+    {# Becasue we use compiled sql for fetching data sampling, we currently not using the macro get_model_sql_from_node #}
```
oravi, line 287: I think this comment is not relevant as the macro is called get_model_sql_from_node and here it's a test

macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -112,6 +112,7 @@
     {%- endif -%}
 
     {% set original_file_path = node_dict.get('original_file_path') %}
+    {# Becasue we use compiled sql for fetching data sampling, we currently not using the macro get_model_sql_from_node #}
```
oravi, line 115: I think this comment is not relevant as the macro is called get_model_sql_from_node and here it's a test

oravi (approved):

PR #99, opened 2022-09-15
Add test alias and test node name to elementary test results

IDoneShaveIt:
dbt allows you to configure a custom name for test.
In this PR I added test alias and test node name to elementary test results table so we would be able to extract the custom name in the CLI.

The graph doesn't contains the custom name param, and the way to retrieve it is to see if the alias and the name fields of the test node are equal.
If so, it means that the user provided a custom name for the test.


macros/edr/system/system_utils/empty_table.sql
```diff
@@ -1,5 +1,5 @@
...
+    {{ elementary.empty_table([('id','long_string'),('data_issue_id','string'),('test_execution_id','long_string'),('test_unique_id','long_string'),('model_unique_id','long_string'),('detected_at','timestamp'),('database_name','string'),('schema_name','string'),('table_name','string'),('column_name','string'),('test_type','string'),('test_sub_type','string'),('test_results_description','long_string'),('owners','string'),('tags','string'),('test_results_query','long_string'),('other','string'),('test_name','long_string'),('test_params','long_string'),('severity','string'),('status','string'),('custom_name', 'string')]) }}
```
IDoneShaveIt, line 2: Put the custom_name column at the end because the way incremental tables with append_column strategy is working is by adding the new column at the end.

macros/edr/system/system_utils/empty_table.sql
```diff
@@ -1,5 +1,5 @@
...
+    {{ elementary.empty_table([('id','long_string'),('data_issue_id','string'),('test_execution_id','long_string'),('test_unique_id','long_string'),('model_unique_id','long_string'),('detected_at','timestamp'),('database_name','string'),('schema_name','string'),('table_name','string'),('column_name','string'),('test_type','string'),('test_sub_type','string'),('test_results_description','long_string'),('owners','string'),('tags','string'),('test_results_query','long_string'),('other','string'),('test_name','long_string'),('test_params','long_string'),('severity','string'),('status','string'),('test_node_name', 'string'),('test_alias', 'string')]) }}
```
oravi, line 2: I think that having both test_node_name and test_name is confusing - we need to either change test_name to be test_short_name and then test_name will be the long but then we will need to think how to support backwards. That's the ideal solution (we can do it by checking if the test_short_name exists use it and if not use the test_name). If that's a big hassle we can maybe rename test_node_name to test_long_name

IDoneShaveIt: I am really against giving names that doesn't match anything in the graph to our fields (like test long name).
It is much more confusing and you can't understand why it is called like that.
Test node name is pretty obvious that is the name of the test node (node.name) IMO.

We can start and make changes in the existing tables and call it short_name and name, but that might cost us more (but if that is the right way then that is the right way).

PR #101, opened 2022-09-17
Issue #300 enhance: add config to not notify skipped models

fredmny:
Adding the option to disable Slack notification for skipped models through the `disable_skipped_model_alerts` var. 

This PR answers the issue # 300 ([link](https://github.com/elementary-data/elementary/issues/300)) of the main repo.

### DIFF
--- a/macros/edr/system/system_utils/get_config_var.sql
+++ b/macros/edr/system/system_utils/get_config_var.sql
@@ -16,7 +16,7 @@
     'disable_model_alerts': false,
     'disable_test_alerts': false,
     'disable_run_results': false,
-    'disable_skipped_model_alerts': false,
+    'disable_skipped_model_alerts': true,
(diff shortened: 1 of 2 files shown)


elongl:
Looks great Fred! Have you tested it and verified that it solves your problem?

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -16,6 +16,7 @@
     'disable_model_alerts': false,
     'disable_test_alerts': false,
     'disable_run_results': false,
+    'disable_skipped_model_alerts': false,
```
elongl, line 19: Could you please change the default value to be `true`?
We've figured out that it just confuses users more than it helps.

elongl (changes requested):

PR #106, opened 2022-09-22
Changed the query that returns the known columns and tables.

elongl:
Fixed a bug in the query that gets the latest known schema of a table.

The query filters by `detected_at` is applied globally rather than to the tested table.

Impact: When concurrent `schema_changes` tests run, if columns were added to the table, there's a chance the alert will not fail.


Maayan-s (approved):
LGTM

PR #109, opened 2022-09-23
Spark profile

Maayan-s:
Tested on dbt cloud


elongl (commented):
Change the adapter type?

Maayan-s:
No, it's intentional. 
We want the local deployment to run with dbt-databricks not dbt-spark (or else the CLI needs to install that adapter).

oravi (approved):

PR #113, opened 2022-10-02
Added source-freshness run results.

elongl:
Currently users need to manually run `dbt run-operation elementary.upload_source_freshness_results` as the `source-freshness` command doesn't trigger an `on-run-end` hook which allows us to upload its results.


macros/edr/dbt_artifacts/upload_source_freshness_results.sql
```diff
@@ -0,0 +1,28 @@
+{% macro upload_source_freshness_results() %}
+  {% set source_freshness_results_relation = ref('elementary', 'source_freshness_results') %}
```
oravi, line 2: cosmetics - just to follow our convention of dbt_ - wdyt about calling it dbt_source_freshness?

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
```diff
@@ -0,0 +1,28 @@
...
+  {% set sources_json_path = ref.config.target_path ~ '/sources.json' %}
```
oravi, line 3: Just to be on the safe side - what do you think about validating the target_path is not null and only then doing this concatenation?

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
```diff
@@ -0,0 +1,28 @@
...
+    {{- return(adapter.dispatch('flatten_source_freshness', 'elementary')) -}}
+{%- endmacro -%}
+
+{%- macro flatten_source_freshness(node_dict) -%}
+    {{- return(adapter.dispatch('flatten_source_freshness', 'elementary')(node_dict)) -}}
+{%- endmacro -%}
+
+{% macro default__flatten_source_freshness(node_dict) %}
+    {% set flatten_source_freshness_dict = {
+         'source_freshness_execution_id': [invocation_id, node_dict.get('unique_id')] | join('.'),
```
oravi, line 20: Add `generated_at`?

macros/edr/system/system_utils/empty_table.sql
```diff
@@ -2,6 +2,10 @@
     {{ elementary.empty_table([('id','long_string'),('data_issue_id','string'),('test_execution_id','long_string'),('test_unique_id','long_string'),('model_unique_id','long_string'),('detected_at','timestamp'),('database_name','string'),('schema_name','string'),('table_name','string'),('column_name','string'),('test_type','string'),('test_sub_type','string'),('test_results_description','long_string'),('owners','string'),('tags','string'),('test_results_query','long_string'),('other','string'),('test_name','long_string'),('test_params','long_string'),('severity','string'),('status','string'),('failures', 'bigint')]) }}
 {% endmacro %}
 
+{% macro empty_source_freshness_results() %}
+    {{ elementary.empty_table([('source_freshness_execution_id','string'),('unique_id','string'),('max_loaded_at','string'),('snapshotted_at','string'),('max_loaded_at_time_ago_in_s','int'),('status','string')]) }}
+{% endmacro %}
```
oravi, line 7: Make sure the types are the same as the types in the json?

models/edr/alerts/alerts_dbt_source_freshness.sql
```diff
@@ -0,0 +1,15 @@
...
+  config(
+    materialized = 'view',
+    bind=False
+  )
+}}
+
+select source_freshness_execution_id as alert_id,
+       unique_id,
+       max_loaded_at
+       snapshotted_at as detected_at,
```
oravi, line 11: 1. Consider adding the join with sources here
2. Add `generated_at`
3. The alias for snapshotted_at should not be detected_at as it's gonna be confusing with `generated_at`. So here I'd use the `snapshotted_at` and in the alerts itself let's call it `source freshness test time`

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
elongl: I don't understand why check a condition that cannot occur.
How would `target_path` be `null`?
Even if you remove `target-path` from `dbt_project.yml`, it still defaults to `target`.

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
elongl: You're comparing it to an artifact (`dbt_`), but it's a run result (`_results`).

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
```diff
@@ -0,0 +1,37 @@
+{% macro upload_source_freshness_results() %}
+  {% set source_freshness_results_relation = ref('source_freshness_results') %}
```
oravi, line 2: Maybe for consistency call it `dbt_source_freshness`?

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
```diff
@@ -0,0 +1,37 @@
...
+       'source_freshness_execution_id': [invocation_id, node_dict.get('unique_id')] | join('.'),
+       'unique_id': node_dict.get('unique_id'),
+       'max_loaded_at': node_dict.get('max_loaded_at'),
+       'snapshotted_at': node_dict.get('snapshotted_at'),
+       'max_loaded_at_time_ago_in_s': node_dict.get('max_loaded_at_time_ago_in_s'),
+       'status': node_dict.get('status'),
+       'error': node_dict.get('error'),
+       'generated_at': elementary.datetime_now_utc_as_string(),
+       'invocation_id': invocation_id,
+       'compile_started_at': node_dict.get('timing/0/started_at'),
```
oravi, line 31: Is this working when reading from json and not a csv?

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
elongl: But this is not an artifact, it's a results table.
It grows with time (incremental) and doesn't represent the latest state like the artifacts do.

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
elongl: Yes

macros/edr/dbt_artifacts/upload_source_freshness_results.sql
oravi: so maybe just add the prefix dbt (dbt_source_freshness_results)? like dbt_run_results?

PR #120, opened 2022-10-13
Seed artifacts

elongl:


macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -18,4 +18,48 @@
...
+  {% endif %}
+
+  {% if value is not string and value is not number %}
+    {{ return(tojson(value)) }}
+  {% else %}
+    {{ return(value) }}
+  {% endif %}
+{% endmacro %}
+
+{% macro are_artifacts_equal(first_artifacts_agate, second_artifacts_agate) %}
```
elongl, line 62: Not yet sure how performant is that.

The problem is that the artifacts contain a timestamp (`generated_at`).
This is guaranteed to change and therefore we cannot compare the file hashes which have to contain that field.

macros/utils/table_operations/seed.sql
```diff
@@ -0,0 +1,21 @@
+{% macro seed_elementary_model(relation, csv_path) %}
+  {% do adapter.dispatch('seed_elementary_model', 'elementary')(relation, csv_path) %}
+{% endmacro %}
+
+{% macro default__seed_elementary_model(relation, csv_path) %}
+  {% set table = elementary.get_agate_table().from_csv(csv_path) %}
+  {% set model = elementary.get_node('model.elementary.' ~ relation.identifier.lower()) %}
+  {% do context.update({'this': relation}) %}
```
elongl, line 8: Don't really like this.
This is why it's needed.

image

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
elongl: Not the most accurate science but seems to be fast.

Created sample data with 25,000 rows where each row is 500 characters:

```py
    with open('raw_data.csv', 'w') as f:
    writer = csv.writer(f)
    writer.writerow(['hello', 'world', 'test'])
    for _ in range(25000):
        writer.writerow([os.urandom(100), os.urandom(200), os.urandom(300)])
```

Both loading the table and and comparing the rows (`t1.rows == t2.rows`) were very fast.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -18,4 +18,60 @@
...
+
+{% macro upload_csv_artifacts_to_table(table_relation, artifacts, flatten_artifact_callback) %}
+    {% set output_csv_path = elementary.get_target_path(table_relation.identifier ~ '.csv') %}
+    {% set flatten_artifact_dicts = [] %}
+    {% for artifact in artifacts %}
+        {% set flatten_artifact_dict = flatten_artifact_callback(artifact) %}
+        {% set flatten_artifact_rendered_dict = {} %}
+        {% for key in flatten_artifact_dict %}
+          {% set value = flatten_artifact_dict[key] %}
+          {% if value %}
```
elongl, line 31: Don't really like this, but `flatten_test()` is currently used for both `elementary_test_results` and `dbt_tests`.
This is a problem since these tables don't share all the columns and therefore certain columns, such as `compiled_code`, will try to insert into `dbt_tests` in which it doesn't exist.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -18,4 +18,61 @@
...
+
+  {% if value is not string and value is not number %}
+    {{ return(tojson(value)) }}
+  {% else %}
+    {{ return(value) }}
+  {% endif %}
+{% endmacro %}
+
+{% macro are_artifacts_equal(first_artifacts_agate, second_artifacts_agate) %}
+  {% set shared_column_names = set(first_artifacts_agate.column_names).intersection(second_artifacts_agate.column_names) %}
```
oravi, line 72: I think that you should start by comparing the columns, if they are not equal just return False. This will protect us if we add more fields to the artifacts

oravi, line 72: I also think set is not supported on all dbt versions

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -1,8 +1,15 @@
 {%- macro upload_dbt_exposures() -%}
+    {% set identifier = 'dbt_exposures' %}
+    {% if results and elementary.get_result_node('model.elementary.%s' % identifier) %}
```
oravi, line 3: cosmetics - I'd get the unique id from the node of dbt_exposures

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -1,8 +1,15 @@
...
+      {{ elementary.debug_log('[%s] Artifacts already ran.' % identifier) }}
+      {{ return(none) }}
+    {% endif %}
+
+    {% set relation = elementary.get_elementary_relation(identifier) %}
     {% set edr_cli_run = elementary.get_config_var('edr_cli_run') %}
     {% if execute and not edr_cli_run %}
         {% set exposures = graph.exposures.values() | selectattr('resource_type', '==', 'exposure') %}
-        {% do elementary.upload_artifacts_to_table(this, exposures, elementary.get_flatten_exposure_callback()) %}
+        {% do elementary.upload_csv_artifacts_to_table(relation, exposures, elementary.get_flatten_exposure_callback()) %}
```
oravi, line 12: Do we want to have a flag that uses the previous artifacts uploading?

dbt_project.yml
```diff
@@ -23,4 +23,5 @@ on-run-start:
   - "{{ elementary.create_elementary_tests_schema() if flags.WHICH in ['test', 'build'] }}"
 on-run-end:
   - "{{ elementary.handle_tests_results(results) if flags.WHICH in ['test', 'build'] }}"
-  - "{{ elementary.upload_run_results(results) }}"
+  - "{{ elementary.upload_run_results(results) if not elementary.get_config_var('disable_run_results') }}"
+  - "{{ elementary.upload_dbt_artifacts() if not elementary.get_config_var('disable_artifacts') }}"
```
oravi, line 27: cosmetics - rename to disable_dbt_artifacts / disable_uploading_dbt_artifacts

macros/edr/system/system_utils/target_path.sql
```diff
@@ -0,0 +1,3 @@
+{% macro get_target_path(filename) %}
+  {{ return(flags.Path("%s/elementary/%s" % (ref.config.target_path, filename))) }}
```
oravi, line 2: use path.join so this will work on all platforms?

macros/utils/graph/get_package_database_and_schema.sql
```diff
@@ -1,11 +1,7 @@
 {% macro get_package_database_and_schema(package_name='elementary') %}
```
oravi, line 1: The previous code supported all packages and the new code supports only Elementary. The new code returns only the Elementary db and schema (note that we can provide a different package name). 
I also don't really like relying on a specific model name in the package as this could break over time if we will remove / rename this model.

macros/utils/table_operations/agate.sql
```diff
@@ -0,0 +1,11 @@
+{% macro get_agate_table() %}
+  {% do return(flags.os.sys.modules['agate'].Table) %}
```
oravi, line 2: Let's talk about this

macros/utils/table_operations/insert.sql
```diff
@@ -9,31 +9,52 @@
...
     {% endfor %}
     {%- if should_commit -%}
         {% do adapter.commit() %}
     {%- endif -%}
 {% endmacro %}
 
-{% macro get_insert_dicts_query(table_relation, columns, dicts) -%}
-    {% set insert_dicts_query %}
+
+{% macro insert_table(table_relation, table, chunk_size=10000, should_commit=False) %}
```
oravi, line 22: Not sure why we need this, could we maybe use insert dicts and pass the agate rows to insert_dicts?

macros/utils/table_operations/insert.sql
```diff
@@ -9,31 +9,52 @@
         {{ elementary.edr_log('Could not extract columns for table - ' ~ table_name ~ ' (might be a permissions issue)') }}
         {{ return(none) }}
     {% endif %}
-    {% set dicts_chunks = elementary.split_list_to_chunks(dicts, chunk_size) %}
-    {% for dicts_chunk in dicts_chunks %}
-        {% set insert_dicts_query = elementary.get_insert_dicts_query(table_relation, columns, dicts_chunk) %}
+    {% for dicts_chunk in dicts | batch(chunk_size) %}
```
oravi, line 12: Not sure what's the benefit here in replacing the exsiting split_list_to_chunks to batch(chunk_size)?

macros/utils/table_operations/seed.sql
```diff
@@ -0,0 +1,20 @@
+{% macro seed_elementary_model(relation, csv_path) %}
```
oravi, line 1: cosmetics - seed has a different meaning in dbt. I'd rename to upload_csv_to_relation

macros/utils/table_operations/seed.sql
```diff
@@ -0,0 +1,20 @@
...
+  {% do adapter.dispatch('seed_elementary_model', 'elementary')(relation, csv_path) %}
+{% endmacro %}
+
+{% macro default__seed_elementary_model(relation, csv_path) %}
+  {% set table = elementary.get_agate_table().from_csv(csv_path) %}
+  {% do adapter.truncate_relation(relation) %}
```
oravi, line 7: This is a bit dangerous as we can end up with an empty table, can we make it atomic? for example we could create a temp table, insert to this table, then do a drop and rename in one transaction or something like that

macros/utils/table_operations/seed.sql
```diff
@@ -0,0 +1,20 @@
...
+  {% do elementary.insert_table(relation, table) %}
+{% endmacro %}
+
+{% macro bigquery__seed_elementary_model(relation, csv_path) %}
+  {% set kwargs = {
+    'allow_quoted_newlines': true,
+    'source_format': 'CSV',
+    'autodetect': true,
```
oravi, line 15: What does the autodetect means if the table exists already?

models/edr/dbt_artifacts/dbt_exposures.sql
```diff
@@ -1,6 +1,6 @@
 {{
   config(
-    materialized='table',
+    materialized='incremental',
```
oravi, line 3: Why?

models/edr/dbt_artifacts/dbt_run_results.sql
```diff
@@ -1,3 +1,8 @@
+{#
```
oravi, line 1: See my comment about it above, I don't think it's a good enough solution

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -18,4 +18,61 @@
...
+      {% do elementary.debug_log("[%s] Artifacts were not changed. Skipping upload." % table_relation.identifier) %}
+      {% do return(none) %}
+    {% endif %}
+    {% do elementary.debug_log("[%s] Uploading artifacts." % table_relation.identifier) %}
+    {% do elementary.seed_elementary_model(table_relation, output_csv_path) %}
+    {% do flatten_artifacts_agate.to_csv(output_csv_path) %}
+{% endmacro %}
+
+{% macro get_cached_artifacts_agate(table_relation, cache_csv_path) %}
+  {% if cache_csv_path.exists() %}
```
oravi, line 49: Not sure that this macro should be responsible for saving to disk the returned artifacts from the db.. wdyt about returning here the cached artifacts and saving it to local disk only if you need to upload (meaning there's a change)

elongl: Interestingly, that was what I originally did.
The thing is that file-caching is much faster than db-caching,
if an environments runs those models but it already has artifacts (say a new engineer joined),
he'll never get the file-caching until he'll change something in the project.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
elongl: `flatten_test` is currently used for both `dbt_tests` and `elementary_test_results` even though they don't share a column, `compiled_code`, which is bad. This becomes a problem because it attempts to `.select(..., compiled_code)` and then it doesn't find it.

A better solution would be either:

1. In `flatten_test`, check if `compiled_code` exists (it only exists on `RunResult`, not `graph`), only if so add it to the artifact dict.
2. Split `flatten_test` to `flatten_test_artifact` and `flatten_test_result`.

What do you think?

elongl: Implementing the above solution would remove this need.

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
elongl: Well, this is the same logic, if it failed why would it succeed on the second attempt?

> I'd get the unique id from the node of dbt_exposures

Sorry, could you please clarify what do you mean? This is the `unique_id`.

image

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
elongl: The thing I'm not quite sure how to solve in terms of the compatibility, is that I've made the artifact models `incremental` so that their contents won't necessarily be deleted when the model is run and by that force the re-upload of the same data. A common case is that the artifacts don't change at all.

So the questions is - Is there a way to make a `table` materialization that doesn't delete its contents? (I'll look into it). Another solution is to just make the previous artifacts uploading delete from the table as well.

macros/utils/graph/get_package_database_and_schema.sql
elongl: I'll revert the changes.

The rational was that we use this macro a lot throughout the code,
I'm not sure if this is micro-optimization, but iterating the graph each time to find an Elementary model and extract its package and schema seems wasteful when you can do direct access, especially if the dbt project is big.

macros/utils/table_operations/insert.sql
elongl: Why re-implement functionality that already exists natively in your runtime?

macros/utils/table_operations/insert.sql
elongl: We duplicate your memory and waste compute by iterating over an agate's rows and copying them into dicts?
This is only a thin wrapper, it does the same thing.

models/edr/dbt_artifacts/dbt_exposures.sql
elongl: I've made the artifact models incremental so that their contents won't necessarily be deleted when the model is run and by that force the re-upload of the same data. A common case is that the artifacts don't change at all.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
oravi: Got it but I'd still write it outside of this macro (like you can move line 45 to be before the `are_artifacts_equal` validation

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
oravi: Not sure I understand.. you should be comparing the same type of artifact (current vs what's saved in file / DWH) no?

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
oravi: I know but if tomorrow dbt will change the format this will longer be valid so I'd rather find the node and get the unique id from the node (you rely on their format)

models/edr/dbt_artifacts/dbt_exposures.sql
oravi: I don't think it's a problem, most of our users don't run it that often unless we explicitly tell them to do so. And in that case it's ok that it will take the additional upload time IMO

macros/utils/table_operations/insert.sql
oravi: It's just that it exists already and the logic is the same and batch is something that is not documented so I don't see a real value relying on it if we already have this logic. wdyt?

macros/utils/table_operations/insert.sql
oravi: Are you sure that passing the rows to insert_dicts means copying the dicts?

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
oravi: See comment on the materialization below

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
elongl: That is also a problem because this could case a discrepancy if the upload failed.
You'd overwrite the cache before you know you successfully made the update.

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
elongl: I can change the signature to be `get_result_node(package_name, identifier)`, think that's more appropriate? Might be a better API in general.

However I'd say that I don't think changing the unique id format is something they're likely to do out of the blue without prior notice.

macros/utils/table_operations/insert.sql
elongl: In order to use `insert_dicts()` you need to use `agate_to_dicts` which clones the same data and by that duplicates it in memory, so yes.

elongl: I now see that `insert_dicts` works with `agate.Table`.
Let's just rename `insert_dicts` to `insert_rows` to indicate that it works with all row-iterables.
What do you think?

models/edr/dbt_artifacts/dbt_exposures.sql
elongl: This means that if a user has `dbt run -s elementary` as part of his pipeline, it'll necessarily re-upload all the artifacts and will never utilize the cache. Are we okay with that?

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -9,7 +15,8 @@
     {%- set flatten_artifacts_len = flatten_artifact_dicts | length %}
     {% if flatten_artifacts_len > 0 %}
         {{ elementary.debug_log('Inserting ' ~ flatten_artifacts_len ~ ' rows to table ' ~ table_relation) }}
-        {% do elementary.insert_dicts(table_relation, flatten_artifact_dicts, elementary.get_config_var('dbt_artifacts_chunk_size'), should_commit) %}
+        {% do dbt.truncate_relation(table_relation) %}
```
elongl, line 18: This is not atomic, but I think it's inevitable given that we're using batched inserts.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -1,4 +1,10 @@
 {% macro upload_artifacts_to_table(table_relation, artifacts, flatten_artifact_callback, should_commit=False) %}
+    {% set artifact_identifier = table_relation.identifier | lower %}
+    {% if results and elementary.get_result_node(artifact_identifier) %}
```
oravi, line 3: I think this should be in the upload_dbt_artifacts on-run-end

macros/utils/graph/get_result_node.sql
```diff
@@ -0,0 +1,14 @@
+{% macro get_result_node(identifier, package_name='elementary') %}
+  {% for result in results %}
+    {% if result.node.identifier == identifier %}
```
oravi, line 3: Are you sure all node types have an identifier? in `get_model_graph_node` we checked identifier / name / alias to be on the safe side

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,75 @@
...
+    {% for insert_query in insert_rows_queries %}
+      {% do elementary.debug_log("[%d/%d] Running insert query." % (loop.index, insert_rows_queries | length)) %}
+      {% do dbt.run_query(insert_query) %}
+    {% endfor %}
+    {%- if should_commit -%}
+        {% do adapter.commit() %}
+    {%- endif -%}
+{% endmacro %}
+
+{% macro get_insert_rows_queries(table_relation, columns, rows, max_query_size=1000000) -%}
```
oravi, line 24: Can we make the max_query_size configurable? I'd call it insert_query_max_size or something along these lines

macros/utils/graph/get_result_node.sql
oravi, line 3: Let's make sure we can rely only on identifier?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,79 @@
...
+    {% set columns = adapter.get_columns_in_relation(table_relation) -%}
+    {% if not columns %}
+        {% set table_name = elementary.relation_to_full_name(table_relation) %}
+        {{ elementary.edr_log('Could not extract columns for table - ' ~ table_name ~ ' (might be a permissions issue)') }}
+        {{ return(none) }}
+    {% endif %}
+
+    {% set insert_rows_queries = elementary.get_insert_rows_queries(table_relation, columns, rows) %}
+    {% for insert_query in insert_rows_queries %}
+      {% do elementary.debug_log("[%d/%d] Running insert query." % (loop.index, insert_rows_queries | length)) %}
```
oravi, line 16: cosmetics - let's calculate the length outside of the loop?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,79 @@
...
+      {% do dbt.run_query(insert_query) %}
+    {% endfor %}
+    {%- if should_commit -%}
+        {% do adapter.commit() %}
+    {%- endif -%}
+{% endmacro %}
+
+{% macro get_insert_rows_queries(table_relation, columns, rows, query_max_size=none) -%}
+    {% if not query_max_size %}
+      {% set query_max_size = elementary.get_config_var('query_max_size') %}
```
oravi, line 26: cosmetics - why not using this as the default?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,79 @@
...
+    {% endif %}
+
+    {% set columns = adapter.get_columns_in_relation(table_relation) -%}
+    {% if not columns %}
+        {% set table_name = elementary.relation_to_full_name(table_relation) %}
+        {{ elementary.edr_log('Could not extract columns for table - ' ~ table_name ~ ' (might be a permissions issue)') }}
+        {{ return(none) }}
+    {% endif %}
+
+    {% set insert_rows_queries = elementary.get_insert_rows_queries(table_relation, columns, rows) %}
```
oravi, line 14: What do you think about adding a feature flag to use the previous method?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,79 @@
...
+    {% for row in rows %}
+      {% set rendered_column_values = [] %}
+      {% for column in columns %}
+        {% set column_value = elementary.insensitive_get_dict_value(row, column.name) %}
+        {% do rendered_column_values.append(elementary.render_value(column_value)) %}
+      {% endfor %}
+      {% set row_sql = "(%s)" % (rendered_column_values | join(",")) %}
+      {% set query_with_row = current_query.data + ("," if not loop.first else "") + row_sql %}
+
+      {% if query_with_row | length > query_max_size %}
```
oravi, line 47: A small edge case here - if the query_max_size is smaller than the insert_query + first row - insert_queries will contain an invalid query with only the boiler plate. This is an edge case but could maybe happen if you have a very long first row so might be better to fix it I think.

dbt_project.yml
```diff
@@ -23,4 +23,5 @@ on-run-start:
   - "{{ elementary.create_elementary_tests_schema() if flags.WHICH in ['test', 'build'] }}"
 on-run-end:
   - "{{ elementary.handle_tests_results(results) if flags.WHICH in ['test', 'build'] }}"
-  - "{{ elementary.upload_run_results(results) }}"
+  - "{{ elementary.upload_run_results(results) if not elementary.get_config_var('disable_run_results') }}"
+  - "{{ elementary.upload_dbt_artifacts() if not elementary.get_config_var('disable_uploading_dbt_artifacts') }}"
```
oravi, line 27: cometics - disable_autoupload_dbt_artifacts?

macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
```diff
@@ -1,3 +1,20 @@
...
+      "dbt_tests": elementary.upload_dbt_tests,
+      "dbt_sources": elementary.upload_dbt_sources,
+      "dbt_snapshots": elementary.upload_dbt_snapshots,
+      "dbt_metrics": elementary.upload_dbt_metrics,
+      "dbt_exposures": elementary.upload_dbt_exposures,
+      }
+    %}
+    {% for artifacts_model, upload_artifacts_func in model_upload_func_map.items() %}
+      {% if not elementary.get_result_node(artifacts_model) %}
+        {% do upload_artifacts_func() %}
```
oravi, line 14: I think we should truncate the relation here and not as part of the upload since you call the upload also from the artifact models so no need to do it there (also it's more clear that upload does only the upload)

oravi: I also think that we should pass the should_commit=True to all the upload functions if you call it from the on-run-end because in redshift it doesn't commit automatically during an on-run-end for every query that you run

dbt_project.yml
oravi, line 27: What do you think about doing this first? then if for some reason the code in upload_run_results and handle_test_results is relying on fresh metadata it will work

macros/utils/table_operations/insert_rows.sql
elongl: I don't understand the renaming.
A warehouse has a max size for a query, the type of query doesn't matter.

macros/utils/graph/get_result_node.sql
elongl: image

All parsed nodes have an `identifier` property.

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -1,8 +1,9 @@
 {%- macro upload_dbt_exposures() -%}
+    {% set relation = elementary.get_elementary_relation('dbt_exposures') %}
     {% set edr_cli_run = elementary.get_config_var('edr_cli_run') %}
-    {% if execute and not edr_cli_run %}
+    {% if execute and relation and not edr_cli_run %}
         {% set exposures = graph.exposures.values() | selectattr('resource_type', '==', 'exposure') %}
-        {% do elementary.upload_artifacts_to_table(this, exposures, elementary.get_flatten_exposure_callback()) %}
+        {% do elementary.upload_artifacts_to_table(relation, exposures, elementary.flatten_exposure, should_commit=true) %}
```
oravi, line 6: I think that you should pass the should_commit=True only if it's being called from the on-run-end

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -14,9 +14,10 @@
     'disable_test_alerts': false,
     'disable_source_freshness_alerts': false,
     'disable_run_results': false,
+    'disable_autoupload_dbt_artifacts': false,
     'disable_skipped_model_alerts': true,
     'disable_skipped_test_alerts': true,
-    'dbt_artifacts_chunk_size': 50,
+    'insert_rows_batch_size': 5000,
```
oravi, line 20: Let's keep it with the same name and value to keep backwards if someone defined it already?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,108 @@
...
+    {% if insert_rows_method == 'max_query_size' %}
+      {% set insert_rows_queries = elementary.get_insert_rows_queries(table_relation, columns, rows) %}
+      {% set queries_len = insert_rows_queries | length %}
+      {% for insert_query in insert_rows_queries %}
+        {% do elementary.debug_log("[%d/%d] Running insert query." % (loop.index, queries_len)) %}
+        {% do dbt.run_query(insert_query) %}
+      {% endfor %}
+    {% elif insert_rows_method == 'batch' %}
+      {% set chunk_size = elementary.get_config_var('insert_rows_batch_size') %}
+      {% for rows_chunk in rows | batch(chunk_size) %}
```
oravi, line 24: I think it's a good practice to use the exact same code if you use the flag to "revert" to the previous method

macros/utils/table_operations/insert_rows.sql
```diff
@@ -0,0 +1,108 @@
...
+      {% for column in columns %}
+        {% set column_value = elementary.insensitive_get_dict_value(row, column.name) %}
+        {% do rendered_column_values.append(elementary.render_value(column_value)) %}
+      {% endfor %}
+      {% set row_sql = "(%s)" % (rendered_column_values | join(",")) %}
+      {% set query_with_row = current_query.data + ("," if not loop.first else "") + row_sql %}
+
+      {% if query_with_row | length > query_max_size %}
+        {% if loop.first %}
+          {% do exceptions.raise_compiler_error("First row to be inserted exceeds the 'query_max_size'.") %}
```
oravi, line 58: cosmetics - you can recommend increasing the query_max_size

macros/edr/system/system_utils/get_config_var.sql
elongl: It's just that this makes _specifically_ the artifacts uploading aware of the insertion method.
I'd need to accept a `chunk_size` on `insert_rows()` even though most of the time it will not be used, and make `upload_artifacts_to_table()` pass that var.

It's not the worst, but I also don't like it that much.
What do you think?

macros/utils/table_operations/insert_rows.sql
elongl: The only difference is using Jinja's built-in `batch()` instead of `split_list_to_chunks()`.
It's also better because it doesn't store in-memory all the chunks duplicated but is rather a generator.
Here's the [source code](https://github.com/pallets/jinja/blob/52843b5cbf635b37a82ac0b6c901921a8ee076ff/src/jinja2/filters.py#L1070).

PR #121, opened 2022-10-13
Created Elementary's internal information schema.

elongl:


dbt_project.yml
```diff
@@ -26,3 +26,4 @@ on-run-end:
   - "{{ elementary.upload_dbt_artifacts() if not elementary.get_config_var('disable_dbt_artifacts_autoupload') }}"
   - "{{ elementary.upload_run_results(results) if not elementary.get_config_var('disable_run_results') }}"
   - "{{ elementary.handle_tests_results(results) if flags.WHICH in ['test', 'build'] }}"
+  - "{{ elementary.upload_information() }}"
```
oravi, line 29: It seems like we are starting to have lots of on-run-end hooks and I think you mentioned this slows us down because it opens connection to the db. Maybe worth uniting them at this point? If you do so add debug logs so we can monitor their timings?

macros/edr/system/system_utils/empty_table.sql
```diff
@@ -20,6 +20,10 @@
     ]) }}
 {% endmacro %}
 
+{% macro empty_information_table() %}
+    {{ elementary.empty_table([('key', 'string'), ('value', 'string')]) }}
```
oravi, line 24: cosmetics - I don't like having generic columns as it's less simple. I'd explicitly call them dbt_version, elementary_version

macros/edr/system/system_utils/get_elementary_package_version.sql
```diff
@@ -0,0 +1,10 @@
+{# Only works if package is pulled from the registry. #}
```
oravi, line 1: What do you mean?

macros/edr/system/system_utils/get_elementary_package_version.sql
```diff
@@ -0,0 +1,10 @@
...
+
+{% macro get_elementary_package_version() %}
+  {% for pkg in ref.config.packages.packages %}
```
oravi, line 4: Is this the only way?

models/edr/system/information.sql
```diff
@@ -0,0 +1,9 @@
+{{
```
oravi, line 1: cosmetics - maybe call it elementary_versions so it's more explicit?

models/edr/system/information.sql
```diff
@@ -0,0 +1,9 @@
...
+  config(
+    materialized = 'table',
+    bind=False,
```
oravi, line 4: bind is only relevant to views AFAIK

macros/edr/system/system_utils/upload_information.sql
```diff
@@ -0,0 +1,26 @@
...
+  {% endif %}
+
+  {% set identifier = 'information' %}
+  {% set relation = elementary.get_elementary_relation(identifier) %}
+  {% if not relation %}
+    {{ return('') }}
+  {% endif %}
+
+  {% if results %}
+    {% if elementary.get_result_node(identifier) %}
```
oravi, line 14: Is there a reason why we will enable uploading it not only from an on-run-end hook?

dbt_project.yml
```diff
@@ -26,3 +26,4 @@ on-run-end:
   - "{{ elementary.upload_dbt_artifacts() if not elementary.get_config_var('disable_dbt_artifacts_autoupload') }}"
```
oravi, line 26: General comment - what do you think about just adding the elementary package version to the new dbt invocations table instead of using a new table? This way you will have history of the upgrades side by side with the dbt versions and you won't need to add additional step to the on-run-end which becomes bigger and bigger.

dbt_project.yml
elongl: https://github.com/elementary-data/dbt-data-reliability/pull/124 😉

macros/edr/system/system_utils/get_elementary_package_version.sql
elongl: You have to use `package` and not `git` or `local` in `packages.yml`.
I'll update the comment.

macros/edr/system/system_utils/get_elementary_package_version.sql
elongl: Unfortunately yes. dbt doesn't provide a native way of doing so.

macros/edr/system/system_utils/empty_table.sql
elongl: I thought of something like [Snowflake's parameters](https://docs.snowflake.com/en/sql-reference/parameters.html) in which we'll store relatively generic information rather than solely the versions.

models/edr/system/information.sql
elongl: I thought of something more generic in which we'll store information rather than solely the versions.
What do you think?

macros/edr/system/system_utils/upload_information.sql
elongl: [Not anymore](https://github.com/elementary-data/dbt-data-reliability/pull/121/commits/b8f7c745f04f2e051f44ad9436190be52f98574f).

dbt_project.yml
elongl: Sounds pretty good. I'll do that.
I'll have to update Dor as well.

elongl:
Replaced with https://github.com/elementary-data/dbt-data-reliability/pull/129.

PR #128, opened 2022-10-27
Hotfix dbt utils

elongl:


macros/utils/cross_db_utils/surrogate_key.sql
```diff
@@ -0,0 +1,30 @@
...
+  {%- set field_sqls = [] -%}
+  {%- for field in fields -%}
+    {%- do field_sqls.append(
+        "coalesce(cast(" ~ field ~ " as " ~ type_string() ~ "), '" ~ default_null_value  ~"')"
+    ) -%}
+    {%- if not loop.last %}
+        {%- do field_sqls.append("'-'") -%}
+    {%- endif -%}
+  {%- endfor -%}
+  {{ hash(concat(field_sqls)) }}
```
elongl, line 29: What are those `hash` and `concat` in the Jinja context?
I think those are macros of `dbt-utils` so maybe need to add package scoping.

PR #129, opened 2022-10-30
Versions in invocations

elongl:


macros/edr/system/hooks/on_run_end.sql
```diff
@@ -0,0 +1,22 @@
+{% macro on_run_end() %}
```
oravi, line 1: Add debug logs for timings?

macros/edr/system/hooks/on_run_end.sql
```diff
@@ -0,0 +1,22 @@
...
+
+  {% if not elementary.get_config_var('disable_dbt_artifacts_autoupload') %}
+    {{ elementary.upload_dbt_artifacts() }}
+  {% endif %}
+
+  {% if not elementary.get_config_var('disable_run_results') %}
+    {{ elementary.upload_run_results(results) }}
+  {% endif %}
+
+  {% if flags.WHICH in ['test', 'build'] %}
```
oravi, line 19: Add flag to disable that as well?

macros/edr/system/hooks/on_run_end.sql
elongl: I'm thinking of adding logs to the inner macros - `upload_dbt_invocation`, `upload_dbt_artifacts`, `upload_run_results`, etc so that the users would understand what happens at which step. What do you think?

PR #133, opened 2022-11-03
Added selectors to invocations.

elongl:


macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -33,6 +61,14 @@
       ('dbt_version', 'string'),
       ('elementary_version', 'string'),
 	  ('full_refresh', 'boolean'),
-      ('vars', 'long_string')
+      ('cli_vars', 'long_string'),
```
smitsrr, line 64: can this be formatted as an array - or does that break across databases?

oravi: Unfortunately it can't be an array because each variable has a name and a value so we have to store it as a dictionary

smitsrr: oh right, of course.

PR #134, opened 2022-11-03
Storing tests' results.

elongl:


macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,8 @@
+{% materialization test, default %}
+    {% set test_sample_sql %}
+        select * from ({{ sql }}) limit {{ elementary.get_config_var('test_sample_row_count') }}
+    {% endset %}
+    {% set test_result = dbt.run_query(test_sample_sql) %}
+    {% do graph["elementary"]["test_results"].update({model.unique_id: test_result}) %}
```
oravi, line 6: cosmetics - rename to test_results_sample ?

macros/edr/system/system_utils/empty_table.sql
```diff
@@ -1,5 +1,5 @@
...
+    {{ elementary.empty_table([('id','long_string'),('data_issue_id','string'),('test_execution_id','long_string'),('test_unique_id','long_string'),('model_unique_id','long_string'),('detected_at','timestamp'),('database_name','string'),('schema_name','string'),('table_name','string'),('column_name','string'),('test_type','string'),('test_sub_type','string'),('test_results_description','long_string'),('owners','string'),('tags','string'),('test_results_query','long_string'),('other','string'),('test_name','long_string'),('test_params','long_string'),('severity','string'),('status','string'),('failures', 'bigint'), ('samples', 'long_string')]) }}
```
oravi, line 2: cosmetics - rename to test_results_sample?

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -335,22 +338,18 @@
...
-    {% if test_compiled_code %}
-        {% if sample_limit %}
-            {% set test_compiled_code = test_compiled_code ~ ' limit ' ~ sample_limit %}
-        {% endif %}
-        {% set test_table_agate = run_query(test_compiled_code) %}
-        {% set test_row_dicts = elementary.agate_to_dicts(test_table_agate) %}
+{% macro get_test_result_rows_as_dicts(flatten_test_node) %}
+    {% set test_result = graph["elementary"]["test_results"].get(flatten_test_node.unique_id) %}
+    {% if not test_result %}
+      {{ return(none) }}
```
oravi, line 350: maybe return empty list here so the code that parses it could just read a json on an empty list?

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,8 @@
+{% materialization test, default %}
+    {% set test_sample_sql %}
+        select * from ({{ sql }}) limit {{ elementary.get_config_var('test_sample_row_count') }}
+    {% endset %}
+    {% set test_result = dbt.run_query(test_sample_sql) %}
```
oravi, line 5: do you want to keep is as an agate table here or covert to dict / json?

macros/edr/tests/on_run_end/handle_tests_results.sql
elongl: I thought it would make more sense that a failed test would have `null` rather than an empty list which is like a passed test.

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -335,22 +338,25 @@
...
-    {% if test_compiled_code %}
-        {% if sample_limit %}
-            {% set test_compiled_code = test_compiled_code ~ ' limit ' ~ sample_limit %}
-        {% endif %}
-        {% set test_table_agate = run_query(test_compiled_code) %}
-        {% set test_row_dicts = elementary.agate_to_dicts(test_table_agate) %}
+{% macro get_test_sample(flatten_test_node) %}
+    {% set test_sample = graph["elementary"]["test_samples"].get(flatten_test_node.unique_id) %}
+    {% if not test_sample %}
+      {{ return([]) }}
```
elongl, line 350: Doesn't it make more sense to store `null` if the test errored?
Just like how `test_results_query` and `failures` are `null`.

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,230 @@
...
+
+  {% set test_result_metrics_rows = {} %}
+  {% set anomaly_scores_rows = elementary.query_test_result_rows() %}
+  {% for anomaly_score_row in anomaly_scores_rows %}
+    {% do test_result_metrics_rows.setdefault(anomaly_score_row.metric_name, []) %}
+    {% do test_result_metrics_rows[anomaly_score_row.metric_name].append(anomaly_score_row) %}
+  {% endfor %}
+
+  {% set elementary_test_results_rows = [] %}
+  {% for metric_rows in test_result_metrics_rows.values() %}
```
elongl, line 28: Not sure that this is the correct logic.

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,235 @@
...
+  {%- endset -%}
+  {% do return(elementary.agate_to_dicts(dbt.run_query(query))) %}
+{% endmacro %}
+
+{% macro cache_elementary_test_results_rows(elementary_test_results_rows) %}
+  {% do elementary.get_cache("elementary_test_results").update({model.unique_id: elementary_test_results_rows}) %}
+{% endmacro %}
+
+
+{% macro handle_anomaly_test(flattened_test) %}
```
elongl, line 13: This is presumably the most important part.

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,238 @@
...
+    {% set sample_row = anomaly_scores_rows[0] %}
+    {% do elementary.cache_elementary_test_results_rows([elementary.get_anomaly_test_result_row(flattened_test, sample_row, anomaly_scores_rows)]) %}
+  {% else %}
+    {% set test_result_metrics_rows = {} %}
+    {% for anomaly_score_row in anomaly_scores_rows %}
+      {% do test_result_metrics_rows.setdefault(anomaly_score_row.metric_name, []) %}
+      {% do test_result_metrics_rows[anomaly_score_row.metric_name].append(anomaly_score_row) %}
+    {% endfor %}
+
+    {% set elementary_test_results_rows = [] %}
```
elongl, line 31: This is not compatible with the [current logic](https://github.com/elementary-data/dbt-data-reliability/blob/3ec6eef10fb4004dbea68561b8395b394f983fe2/macros/edr/tests/on_run_end/handle_tests_results.sql#L28) in which we create a row for each failure or for the latest anomaly scores.

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,238 @@
...
+              lag(min_metric_value) over (partition by full_table_name, column_name, metric_name, dimension, dimension_value order by bucket_end)
+             else min_metric_value end as min_value,
+             case when is_anomalous = TRUE then
+              lag(max_metric_value) over (partition by full_table_name, column_name, metric_name, dimension, dimension_value order by bucket_end)
+              else max_metric_value end as max_value,
+             bucket_start as start_time,
+             bucket_end as end_time,
+             is_anomalous,
+             dimension,
+             dimension_value,
```
haritamar, line 143: Where are you using the added `dimension` and `dimension_value` fields? (didn't see them in `handle_test_result`)

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,243 @@
...
+    {% set test_result_metrics_rows = {} %}
+    {% for anomaly_score_row in anomaly_scores_rows %}
+      {% do test_result_metrics_rows.setdefault(anomaly_score_row.metric_name, []) %}
+      {% do test_result_metrics_rows[anomaly_score_row.metric_name].append(anomaly_score_row) %}
+    {% endfor %}
+
+    {% set elementary_test_results_rows = [] %}
+    {% for metric_rows in test_result_metrics_rows.values() %}
+      {% for metric_row in metric_rows %}
+        {% do elementary_test_results_rows.append(elementary.get_anomaly_test_result_row(flattened_test, metric_row, metric_rows)) %}
```
haritamar, line 34: Should there actually be more than one row per metric?
(a bit strange to me that we add multiple rows for the same metric with the same full sample, but I'm guessing that doesn't really happen)

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,243 @@
...
+      'test_sub_type': metric_name,
+      'test_results_description': test_results_description,
+      'other': elementary.insensitive_get_dict_value(elementary_test_row, 'anomalous_value'),
+      'owners': elementary.insensitive_get_dict_value(flattened_test, 'model_owners'),
+      'tags': elementary.insensitive_get_dict_value(flattened_test, 'model_tags'),
+      'test_results_query': test_results_query,
+      'test_name': elementary.insensitive_get_dict_value(flattened_test, 'short_name'),
+      'test_params': elementary.insensitive_get_dict_value(flattened_test, 'test_params'),
+      'severity': elementary.insensitive_get_dict_value(flattened_test, 'severity'),
+      'result_rows': elementary.render_result_rows(anomaly_score_rows)
```
haritamar, line 181: Shouldn't we limit `anomaly_score_rows` here (or in `render_result_rows`)? because if I understand correctly in `handle_anomaly_test` we query the rows without limit (which makes sense).

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,238 @@
...
+    "anomaly_detection": elementary.handle_anomaly_test,
+    "schema_change": elementary.handle_schema_changes_test
+  } %}
+  {% set test_type_handler = test_type_handler_map.get(test_type, elementary.handle_dbt_test) %}
+  {% do test_type_handler(flattened_test) %}
+{% endmacro %}
+
+{% materialization test, default %}
+  {% do elementary.materialize_test() %}
+  {{ return(dbt.materialization_test_default()) }}
```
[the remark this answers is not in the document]
elongl: I'm not doing this because I'm not sure which default materialization to call,
I'm doing it because I myself need to override the default materialization.
If I don't define it, my code wouldn't be called _at all_ on Snowflake.

macros/edr/materializations/tests/test.sql
elongl: The query itself is limited to 5 rows on a dbt test,
for an Elementary test, I'd rather either include all the results (only up to `days_back`),
or just have `null` so that we query with `test_results_query`.

macros/edr/materializations/tests/test.sql
elongl: The `all_columns_anomalies` test includes a row per column per metric, in which case it's necessary to do so.

macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
```diff
@@ -21,6 +21,6 @@
         {% do elementary.debug_log('[{}] Artifacts already ran.'.format(artifacts_model)) %}
       {% endif %}
     {% endfor %}
-    {% do elementary.edr_log("Uploaded dbt artifacts successfully.") %}
+    {% do elementary.debug_log("Uploaded dbt artifacts successfully.") %}
```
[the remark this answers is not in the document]
elongl: Agreed, applied.

macros/edr/materializations/tests/test.sql
haritamar: 👍

macros/edr/materializations/tests/test.sql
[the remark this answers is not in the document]
elongl: Correct, I'm not aware of a way to do it right now.
If the need arises, might make some research to tackle that.

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,242 @@
...
+
+  {% set anomaly_scores_rows = elementary.query_test_result_rows() %}
+  {% if anomaly_scores_rows and flattened_test.short_name == "dimension_anomalies" %}
+    {% set sample_row = anomaly_scores_rows[0] %}
+    {% do elementary.cache_elementary_test_results_rows([elementary.get_anomaly_test_result_row(flattened_test, sample_row, anomaly_scores_rows)]) %}
+  {% else %}
+    {% set test_result_metrics_rows = {} %}
+    {% for anomaly_score_row in anomaly_scores_rows %}
+      {% do test_result_metrics_rows.setdefault(anomaly_score_row.metric_name, []) %}
+      {% do test_result_metrics_rows[anomaly_score_row.metric_name].append(anomaly_score_row) %}
```
IDoneShaveIt, line 28: Naming is a bit confusing.
The final dict called `test_result_metrics_rows` but it contains anomaly score rows and not metric rows 🤔

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,242 @@
...
+
+{% macro handle_anomaly_test(flattened_test) %}
+  {% set metrics_tables_cache = elementary.get_cache("tables").get("metrics") %}
+  {% set metrics_table = elementary.get_elementary_test_table(flattened_test.name, 'metrics') %}
+  {% if metrics_table %}
+    {% do metrics_tables_cache.append(metrics_table) %}
+  {% endif %}
+
+  {% set anomaly_scores_rows = elementary.query_test_result_rows() %}
+  {% if anomaly_scores_rows and flattened_test.short_name == "dimension_anomalies" %}
```
IDoneShaveIt, line 21: I think it will be good to add a comment (or even split into smaller macros) to explain why we handle dimension like that, and why we handle other anomalies different.

This is the most important part as you said, and we should make it as easy as possible for someone to understand the logic behind it 🙂

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,242 @@
...
+    {% set sample_row = anomaly_scores_rows[0] %}
+    {% do elementary.cache_elementary_test_results_rows([elementary.get_anomaly_test_result_row(flattened_test, sample_row, anomaly_scores_rows)]) %}
+  {% else %}
```
IDoneShaveIt, line 24: See comment on the if part

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,242 @@
...
+  {% endif %}
+  {% do return("dbt_test") %}
+{% endmacro %}
+
+{% macro materialize_test() %}
+  {% set flattened_test = elementary.flatten_test(model) %}
+  {% set test_type = elementary.get_elementary_test_type(flattened_test) %}
+  {% set test_type_handler_map = {
+    "anomaly_detection": elementary.handle_anomaly_test,
+    "schema_change": elementary.handle_schema_changes_test
```
IDoneShaveIt, line 80: I think it will better to add the dbt_test type to the type handler map as well.
You want the map to contain all the supported types there is (even if you use the dbt test type as default from the outside).

macros/edr/materializations/tests/test.sql
```diff
@@ -0,0 +1,242 @@
...
+  } %}
+  {% set test_type_handler = test_type_handler_map.get(test_type, elementary.handle_dbt_test) %}
```
IDoneShaveIt, line 82: Thinking out loud:
To add a new type of test, we need to manually create a new one.
Don't we want to make it mandatory to add a new handler and extend the map?
Maybe we should throw an error if a new type is added and it has no reference in the type handler map (instead of trying to run a default handler)? That way we would make sure that we don't miss any step when adding a new test type, and it will help us understand what is missing as well while developing.
WDYT?

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -1,373 +1,44 @@
...
-    {% endif %}
-    {% set recent_anomaly_sample = most_recent_anomalies_scores[0] if most_recent_anomalies_scores else {} %}
-    {% if anomalous_dimensions %}
-        {% set test_results_description = "There are " ~ anomalous_dimensions | length ~ " anomalous values for '" ~ recent_anomaly_sample.get('dimension') ~ "'" %}
-    {% else %}
-        {% set test_results_description = none %}
+    {# Materializing the test failed and therefore was not added to the cache. #}
+    {% if not elementary_test_results_rows %}
+      {% set flattened_test = elementary.flatten_test(result.node) %}
+      {% set elementary_test_results_rows = [elementary.get_dbt_test_result_row(flattened_test)] %}
```
IDoneShaveIt, line 29: mmmm how is it going to work on tests that are not dbt_test?
How is it going to affect the report when an anomaly test returns results in the format of a dbt test?

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -1,373 +1,44 @@
...
-            {% endif %}
-        {% endfor %}
-    {% endif %}
-    {% set recent_anomaly_sample = most_recent_anomalies_scores[0] if most_recent_anomalies_scores else {} %}
-    {% if anomalous_dimensions %}
-        {% set test_results_description = "There are " ~ anomalous_dimensions | length ~ " anomalous values for '" ~ recent_anomaly_sample.get('dimension') ~ "'" %}
-    {% else %}
-        {% set test_results_description = none %}
+    {# Materializing the test failed and therefore was not added to the cache. #}
+    {% if not elementary_test_results_rows %}
```
IDoneShaveIt, line 27: Do we want to log that something went wrong 🤔?

IDoneShaveIt (commented):
Really like this change!

macros/edr/tests/on_run_end/handle_tests_results.sql
elongl: dbt logs it automatically when a test fails to materialize.

```
10:24:11  Completed with 1 error and 0 warnings:
10:24:11
10:24:11  Database Error in model error_model (models/error_model.sql)
10:24:11    001003 (42000): SQL compilation error:
10:24:11    syntax error line 4 at position 16 unexpected 'as'.
10:24:11    compiled Code at target/run/elementary_integration_tests/models/error_model.sql
10:24:11
```

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -1,372 +1,44 @@
...
-        {{ elementary.merge_data_monitoring_metrics(database_name, schema_name, test_metrics_tables) }}
-        {{ elementary.merge_schema_columns_snapshot(database_name, schema_name, test_columns_snapshot_tables) }}
-        {% if elementary_test_results %}
-            {%- set elementary_test_results_relation = adapter.get_relation(database=database_name,
-                                                                            schema=schema_name,
-                                                                            identifier='elementary_test_results') -%}
-            {%- do elementary.insert_rows(elementary_test_results_relation, elementary_test_results, should_commit=True) -%}
-        {% endif %}
+    {{ elementary.edr_log("Handling test results.") }}
+    {% set cached_elementary_test_results = elementary.get_cache("elementary_test_results") %}
```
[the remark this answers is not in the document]
elongl: Definitely agree.
What do you think about `state` or `storage`?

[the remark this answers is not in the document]
haritamar: (Wrote on slack too but adding here for documentation)

Between the two I'd go with state, not sure how clear storage is (sounds like an adapter for writing files or something :slightly_smiling_face: )
Maybe even get_var/set_var

haritamar (approved):

PR #137, opened 2022-11-06
Add test node name and test alias to elementary test results

IDoneShaveIt:
dbt allows you to configure a custom name for test.
In this PR I added test alias and test node name to elementary test results table so we would be able to extract the custom name in the CLI.

The graph doesn't contains the custom name param, and the way to retrieve it is to see if the alias and the name fields of the test node are equal.
If so, it means that the user provided a custom name for the test.


macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -172,11 +172,13 @@
         'owners': elementary.insensitive_get_dict_value(test_node, 'model_owners'),
         'tags': elementary.insensitive_get_dict_value(test_node, 'model_tags'),
         'test_results_query': test_results_query,
-        'test_name': elementary.insensitive_get_dict_value(test_node, 'short_name'),
+        'test_name': elementary.insensitive_get_dict_value(test_node, 'name'),
```
haritamar, line 175: Is there a compatibility issue here? 
E.g. if people update the DBT package but not the CLI this can cause a behavior change, no?
Though in any case you mentioned we're increasing the minor version.

IDoneShaveIt: There is no problem with it because `name` is a field that is already on the `test_node` object, and we just replaced how we create the `test_name` key (without creating any new thing).

And yes, the version bumping will make us safe anyway!

haritamar (approved):

PR #140, opened 2022-11-08
remove compiled_code from dbt_tests

nzewail:
I noticed that there is a `compiled_code` column defined in `dbt_tests` model that doesn't show up in the model SQL

### DIFF
--- a/models/dbt_artifacts.yml
+++ b/models/dbt_artifacts.yml
@@ -174,10 +174,6 @@ models:
         data_type: string
         description: ""
 
-      - name: compiled_code
-        data_type: string
-        description: ""
-
       - name: path
         data_type: string
         description: ""


elongl:
Thanks a lot for your contribution! Merging.

nzewail:
Do we need to cut a new release for `0.5.6`?

elongl:
I think we can leave it at `master` for the time being as long as it doesn't affect the package in terms of usability,
and release it on the next version.

@nzewail 
What do you think?

nzewail:
> I think we can leave it at `master` for the time being as long as it doesn't affect the package in terms of usability, and release it on the next version.
> 
> @nzewail What do you think?

That seems fine! I found this because we hav a library internally that checks for these things and so our CI broke but yeah I agree it can wait for the next release

PR #143, opened 2022-11-09
Fix bug: model_run_results.sql

eKatia:
The index for the last and first execution of the dat was not working properly because the sorting was based on the timestamp, not date.


elongl:
Thanks for fixing this. Merging.

PR #147, opened 2022-11-14
DRAFT: Add dbt_invocations

smitsrr:
Adding the documentation for `dbt_invocations` model.

### DIFF
--- a/models/run_results.yml
+++ b/models/run_results.yml
@@ -70,6 +70,10 @@ models:
         data_type: string
         description: The selected resources in the dbt command. While this is a string in the database, this can easily be converted to an array.
 
+      - name: selected_resources
+        data_type: string
+        description: Deprecated and replaced with Selected. Run a full-refresh to remove this column (history will be lost). The selected resources in the dbt command. While this is a string in the database, this can easily be converted to an array.
+
       - name: yaml_selector
         data_type: string
         description: The yaml selector that was passed in this invocation.
--- a/models/run_results.yml
+++ b/models/run_results.yml
@@ -16,24 +16,24 @@ models:
               to: ref('dbt_run_results')
 
       - name: run_started_at
-        data_type:
-        description:
+        data_type: timestamp
+        description: Timestamp the invocation was started.
 
       - name: run_completed_at
-        data_type:
-        description:
+        data_type: timestamp
+        description: Timestamp the invocation was completed
 
       - name: generated_at
-        data_type:
-        description:
+        data_type: timestamp
+        description: Timestamp the invocation was completed and generated.
 
       - name: command
-        data_type:
-        description:
+        data_type: string
+        description: dbt command that was used. For example, run.
 
       - name: dbt_version
-        data_type:
-        description:
+        data_type: string
+        description: Version of dbt that was used in this invocation.
 
       - name: elementary_version
         data_type:
@@ -44,16 +44,16 @@ models:
         description: Whether or not this invocation was executed with a full-refresh flag
 
       - name: cli_vars
-        data_type:
-        description:
+        data_type: string
+        description: Dictionary of the variables (and values) that were declared in the invocation.
 
       - name: vars
-        data_type:
-        description:
+        data_type: string
+        description: Dictionary of all variables (and values) in the dbt project. If none were declared at runtime, these are the variables declared in dbt_project yml
 
       - name: target_name
-        data_type:
-        description:
+        data_type: string
+        description: Name of the target used in this invocation.
 
       - name: target_database
         data_type: string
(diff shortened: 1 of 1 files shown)


models/run_results.yml
```diff
@@ -72,6 +122,12 @@ models:
         data_type: string
         description: The compiled code (SQL / Python) executed against the database.
 
+  - name: elementary_test_results
```
smitsrr, line 125: Alphabetized the models by name.

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+        data_type: boolean
+        description: Whether or not this invocation was executed with a full-refresh flag
+
+      - name: cli_vars
+        data_type: string
+        description: Dictionary of the variables (and values) that were declared in the invocation.
+
+      - name: vars
+        data_type: string
+        description: Dictionary of all variables (and values) in the dbt project. If none were declared at runtime, these are the variables declared in dbt_project yml
```
smitsrr, line 52: For future reference, 
```
parse_json(vars) as vars
```
returns the variables as json dictionary. (in snowflake)

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
 models:
+  - name: dbt_invocations
+    description: >
+      Attributes associated with each dbt invocation. Inserted at the end of each invocation.
+
+    columns:
+      - name: invocation_id
+        data_type: string
+        tests:
+          - not_null
```
smitsrr, line 12: I'm not sure if you want tests in these models, but it is useful for users trying to understand the relationships

oravi: Yeah it's definitely useful! the reason we didn't add tests here so far was that we don't want to ask our users to actively exclude these tests when they run `dbt test`. So unless we have a good solution for that for now I'd avoid adding it here

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+          - unique
+          - relationships:
+              field: invocation_id
+              to: ref('dbt_run_results')
+
+      - name: run_started_at
+        data_type: timestamp
```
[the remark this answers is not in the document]
oravi: I know it's a bit confusing and we could definitely improve it down the road but currently the type is string (varchar).. it was easier to support multiple platforms this way and also very easy to convert to timestamp if needed

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+        description: Timestamp the invocation was started.
+
+      - name: run_completed_at
```
oravi, line 22: Type is string as well

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+        data_type: timestamp
+        description: Timestamp the invocation was completed
+
+      - name: generated_at
```
oravi, line 26: Type is string as well. The description could also be - the time this invocation was uploaded to the database.

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+      - name: command
+        data_type: string
+        description: dbt command that was used. For example, run.
+
+      - name: dbt_version
+        data_type: string
+        description: Version of dbt that was used in this invocation.
+
+      - name: elementary_version
+        data_type:
```
oravi, line 39: Data type is string as well. Description could be - Version of the elementary package that was used in this invocation.

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+        data_type: string
+        description: Dictionary of all variables (and values) in the dbt project. If none were declared at runtime, these are the variables declared in dbt_project yml
+
+      - name: target_name
+        data_type: string
+        description: Name of the target used in this invocation.
+
+      - name: target_database
+        data_type: string
+        description: Name of the target schema that was used in this invocation.
```
oravi, line 60: Small fix here - Name of the target *database* that was used in this invocation.

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+
+      - name: target_profile_name
+        data_type: string
+        description: Name of the dbt profile that was used in this invocation.
+
+      - name: threads
+        data_type: integer
+        description: ""
```
oravi, line 72: Description could be - Number of threads that were used to run this dbt invocation.
(this number could impact the performance of a dbt invocation).

models/run_results.yml
```diff
@@ -1,6 +1,85 @@
...
+
+      - name: selected
+        data_type: string
+        description: While this is a string in the database, this can easily be converted to an array.
```
oravi, line 76: I'd add to the description - 'the selected resources in the dbt command. While this is a string in the database, this can easily be converted to an array.'

Issue #152, opened 2022-11-22
On-run-end macros for storing dbt_tests throws an error


jgrinaveckis:
Hello,

Recently decided to try and use elemenary-data for observability, but stumbled upon problem that BigQuery throws me an error **_Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex._**

Which could be expected, because as of now we have ~1600 tests and trying to insert all that metadata about each of them seems to be too much for BigQuery. 

I was wondering maybe we could retrieve test resources from manifest.json based on name thats provided after `-s` flag. There should be multiple checks I suppose, based on different possibilities for model selection (concrete name, tag, etc.).

Do you think that could be a viable solution for monorepo lovers out here?

Many thanks,
Justas

oravi:
Hi @jgrinaveckis ! we discussed this over Slack but for the sake of documentation I'll also share the suggested workaround here - 

By default Elementary tries to upload the dbt metadata to the DWH with less insert queries (to make the upload faster), To achieve that we use the maximum possible query size based on the limitation of each platform. For some reason, specifically on BigQuery our current default is probably too big and it creates a query that could be problematic in terms of size. This is why you received the error above. 

We do support controlling the size of these insert queries if needed by the adding the following vars - 
 
```
vars:
  insert_rows_method: chunk
  dbt_artifacts_chunk_size: 500
```

This means that each insert query of dbt artifacts will only upload 500 rows. If this is still not working for you, you can reduce the chunk size further up to a point where the generated query doesn't exceed BigQuery's limits. 
For a more robust solution we will change the defaults of our BigQuery limits so potentially this won't happen for more users. 

I think that once we change the defaults in BigQuery we can close this issue.

jgrinaveckis:
Hey @oravi,

Yes, I think we are pretty much clear on this and keeping this open as this still is relevant for BigQuery users.
Thanks for information!

oravi:
Thanks @jgrinaveckis ! 
Just updating that we have opened a PR to change the default method and chunk size in BigQuery.

elongl:
We've released a fix that should resolve this issue.

PR #153, opened 2022-11-22
Bigquery insert query defaults

oravi:


oravi:
Fixes #152

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -30,12 +38,17 @@
...
     'query_max_size': 1000000,
     'insert_rows_method': 'max_query_size'
   } %}
+  {{- return(default_config) -}}
+{%- endmacro -%}
 
-  {{ return(var(var_name, default_config.get(var_name))) }}
-{% endmacro %}
+{%- macro bigquery__get_default_config() -%}
+    {% set default_config = elementary.default__get_default_config() %}
```
elongl, line 51: Not that important as we can't have the same dbt process run on two different adapters,
but you're overriding the default config for all adapters.

```suggestion
    {% set default_config = elementary.default__get_default_config().copy() %}
```

If we'd have logic that compares different config values among adapters, this would break it.

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -20,6 +27,7 @@
     'disable_skipped_model_alerts': true,
     'disable_skipped_test_alerts': true,
     'dbt_artifacts_chunk_size': 5000,
+    'test_sample_row_count': 5,
```
elongl, line 30: Those are unrelated changes, any reason they're on this PR?

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -30,12 +38,17 @@
       'column_numeric': ['min', 'max', 'zero_count', 'zero_percent', 'average', 'standard_deviation', 'variance']
     },
     'time_format': '%Y-%m-%d %H:%M:%S',
-    'long_string_size': 16384,
+    'long_string_size': 65535,
```
elongl, line 41: Same as above.

macros/edr/system/system_utils/get_config_var.sql
oravi: I don't think that it's important, also don't want to pay the copy for every var access

PR #161, opened 2022-11-28
Added jobs.

elongl:


macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -41,6 +47,54 @@
...
+{% macro get_env() %}
+    {% set env = elementary.get_config_var("env") %}
+    {% if env %}
+        {{ return(env) }}
+    {% endif %}
+    {{ return(elementary.get_first_not_none_env_var(["DBT_ENV"])) }}
+{% endmacro %}
+
+{% macro get_job_id() %}
+    {% set job_id = elementary.get_config_var("job_id") %}
```
elongl, line 59: @oravi 
Do you think those should also be just env vars like `JOB_ID`?
Most likely that if they'd add a var it'll just be
```yml
vars:
  job_id: "{{ env_var('JOB_ID') }}"
```

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -25,6 +26,11 @@
       'selected': elementary.get_invocation_select_filter(),
       'yaml_selector': elementary.get_invocation_yaml_selector(),
       'project_name': elementary.get_project_name(),
+      'env': elementary.get_env(),
+      'run_reason_category': elementary.get_run_reason_category(),
+      'run_reason': elementary.get_run_reason(),
+      'pull_request_id': elementary.get_pr_id(),
+      'git_sha': elementary.get_git_sha(),
```
elongl, line 33: Anything else we should add?
Here's [GitHub's envs](https://docs.github.com/en/actions/learn-github-actions/environment-variables).

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -41,6 +51,46 @@
     {% do return(config.project_name) %}
 {% endmacro %}
 
+{% macro get_env() %}
```
IDoneShaveIt, line 54: Isn't a function that returns functionB(var) is a bit redundant?
I can see why making those macros is good when we want to use them in other cases as well, which in that case I think they should be under env_utils or something like that 🙂

IDoneShaveIt (commented):
Looking good 🙂 
Left one small comment

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
elongl: Actually yeah, I agree.
These functions were longer beforehand but now that it's only a single expression,
makes more sense. I'll change.

PR #167, opened 2022-12-01
Support dbt expectations

IDoneShaveIt:
- Added a subtype for dbt_expectations tests - expectation
- Added a default description for dbt expectations tests


macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -133,12 +141,88 @@
...
         {% set test_type = 'generic' %}
     {%- elif 'tests/' in test_path -%}
         {% set test_type = 'singular' %}
     {%- endif -%}
     {{- return(test_type) -}}
-{%- endmacro -%}
\ No newline at end of file
+{%- endmacro -%}
+
+{% macro get_default_description(short_name, test_namespace = none) %}
```
IDoneShaveIt, line 156: Not sure if this is the right place for this macro 🤔 
It seems OK because it is only used here but LMK if you think of a better place to put it!

PR #171, opened 2022-12-04
Schema changes from baseline

haritamar:
This PR adds support for testing schema changes from a baseline. 
This complements the existing `schema_changes` tests, but allows comparing against a baseline defined as a part of a source's or model's definition.

Example of how it looks like:
```
sources:
  - name: tcph_sf1
    database: "SNOWFLAKE_SAMPLE_DATA"
    schema: "TPCH_SF1"
    tables:
      - name: orders
        columns:
          - name: o_clerk
            data_type: TEXT
          - name: o_custkey
            data_type: NUMBER
        tests:
          - elementary.schema_changes_from_baseline:
             fail_on_added: false              # Optional - if set to True, we'll fail on added columns (default is False)
             enforce_types: false              # Optional - if set to True, we'll cause a test compilation error if columns are defined without types
```

We also have a supplied macro that can help to generate these tests automatically. They can be generated either for a particular source / model, or for all of them. Examples:
```
dbt run-operation elementary.generate_schema_baseline_test --args '{"name": "orders"}'                      # Generate for a model / source named "orders"
dbt run-operation elementary.generate_schema_baseline_test --args '{"name": "orders", strict: true}'        # Generate for a model / source named "orders" in strict mode (fail on added columns)
dbt run-operation elementary.generate_schema_baseline_test                                                  # Generate a test for all sources / models
dbt run-operation elementary.generate_schema_baseline_test --args '{"include_models": false}'               # Generate a test for all sources, but exclude models
```

The generated results have the same format as the existing `schema_changes` test, and shown in the report in the same way:
image


integration_tests/models/schema.yml
```diff
@@ -234,13 +234,31 @@ sources:
     schema: "{{ target.schema }}"
     tables:
       - name: groups
+        columns:
+          - name: group_a
+            data_type: "{{ 'STRING' if target.type == 'bigquery' else 'character varying' if target.type == 'redshift' else 'TEXT' }}"
```
haritamar, line 239: I need the data type to be the exact type in the information schema for each platform :/

elongl: Have you considered `data_type: "{{ elementary.type_string() }}"`?

haritamar: `type_string` won't be enough for me as it's not exactly the same as what I'll find in INFORMATION_SCHEMA.
But extracting to a macro absolutely makes sense, I'll do it.

haritamar: Hmm seems like macros are not supported in yamls and I think you can't add "{% %}" sections there either (I think the yaml is parsed before dbt expands specific settings with jinja)
If you have another idea that's less ugly than what I currently did will be happy to hear :)

elongl: Apologies for misleading, I recalled this is supported.

elongl (commented):
Looks great!
Very small suggestions, just to add package scoping.

macros/commands/generate_schema_baseline_test.sql
```diff
@@ -0,0 +1,85 @@
...
+  {% endif %}
+{% endmacro %}
+
+{% macro generate_schema_baseline_test_for_all_nodes(include_sources=True, include_models=True, fail_on_added=False, enforce_types=False) %}
+  {% set nodes = get_nodes_from_graph() %}
+  {% for node in nodes %}
+    {% if node.package_name != 'elementary' and
+          ((include_sources and node.resource_type == 'source') or
+           (include_models and node.resource_type == 'model')) %}
+      {% do print("Generating schema changes from baseline test for " ~ node.resource_type ~ " '" ~ node.name ~ "':") %}
```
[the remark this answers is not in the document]
haritamar: Yup I agree, changing

PR #179, opened 2022-12-07
Postgres integration

kouridis:


macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -46,7 +46,7 @@
                     dimension_value,
                     1 as joiner
                 from filtered_monitored_table
-            )
+            ) rs
```
elongl, line 49: What are those?

kouridis: In postgres subqueries in FROM must have an alias

elongl: I see. What does `rs` stand for?

kouridis: result set. I use it in my queries

elongl:
Merging this so we can continue the work on this branch!

PR #181, opened 2022-12-11
Uploading artifacts according to the command that was ran.

elongl:


macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
```diff
@@ -1,23 +1,24 @@
...
-      "dbt_tests": elementary.upload_dbt_tests,
-      "dbt_sources": elementary.upload_dbt_sources,
-      "dbt_snapshots": elementary.upload_dbt_snapshots,
-      "dbt_metrics": elementary.upload_dbt_metrics,
-      "dbt_exposures": elementary.upload_dbt_exposures,
-      }
-    %}
+    {% set artifacts = [
+      {"model": "dbt_models", "which": "run", "handler": elementary.upload_dbt_models},
+      {"model": "dbt_sources", "which": "run", "handler": elementary.upload_dbt_sources},
```
elongl, line 5: Maybe `sources`, `metrics`, and `exposures` should not be uploaded by default?

IDoneShaveIt: Yeah I think that we talked about it when we planned the task.
We can only upload the specific models that ran / tests that were executed / etc.
But maybe for now this is enough

IDoneShaveIt (approved):
LGTM!

PR #186, opened 2022-12-13
Json schema validation

haritamar:
This PR adds support for validating JSON schemas on a column of a model / source. The implementation is based on our Python tests capability.

Example of how it looks like:
```yaml
models:
  - name: my_first_dbt_model
    columns:
      - name: bla
        tests:
          - elementary.json_schema:
              properties:
                a:
                  type: string
              required:
                - a
              type: object
```

We also have a supplied macro that can help to generate these tests automatically for a particular node and column name (right now I didn't implement generating for an entire model / project)
In order to generate the schema, we use the [genson](https://pypi.org/project/genson/) python library.

For example:
```
>>> dbt run-operation elementary.generate_json_schema_test --args '{"node_name": "my_first_dbt_model", "column_name": "bla"}'

09:07:19  Running with dbt=1.3.0
Please add the following test to your model configuration:
----------------------------------------------------------

columns:
  - name: bla
    tests:
      - elementary.json_schema:
          properties:
            a:
              type: integer
          required:
          - a
          type: object
``` 

Unfortunately though, BigQuery currently does not support installing python packages from PyPi automatically so it requires the user to install `genson` manually in their Dataproc cluster :/ (and it can't work for the serverless submission method)

The PR also contains some changes to the Python infra:
1. The ability to run a custom function that is not a test (needed for the generation).
2. Added a fix to support python tests for sources.


macros/edr/tests/test_json_schema.sql
```diff
@@ -0,0 +1,43 @@
+{% test test_json_schema(model, column, json_schema) %}
```
elongl, line 1: I think the test should be just called `json_schema`, there's no need to re-specify that it's a test.

elongl (commented):
Looks great!
Left one comment.

macros/commands/generate_json_schema_test.sql
```diff
@@ -0,0 +1,101 @@
...
+
+    {% set gen_json_schema_func = elementary.generate_json_schema_py_func(column_name) %}
+    {% set node_relation = node_relation.quote(false, false, false) %}
+    {% set compiled_py_code = adapter.dispatch('compile_py_code', 'elementary')(node_relation, gen_json_schema_func,
+                                                                                output_table, code_type='function') %}
+
+    {% do elementary.run_python(node, compiled_py_code) %}
+    {% set json_schema = elementary.result_value('select result from {}'.format(output_table)) %}
+    {% if json_schema == 'genson_not_installed' %}
+      {% do print("ERROR - the 'genson' python library is missing from your warehouse.") %}
```
elongl, line 39: Maybe use `exceptions.raise_compiler_error()` here?
Probably thinking too long into the future, but we'd probably want the macro to return non-zero exit code so that it fails if it's in CI or something.

macros/commands/generate_json_schema_test.sql
```diff
@@ -0,0 +1,101 @@
...
+    import genson
+except ImportError:
+    genson = None
+
+def get_column_name_in_df(df, column):
+    matching = [col for col in df.columns if col.lower() == column.lower()]
+    if len(matching) == 0:
+        raise Exception("No column with the name: {}".format(col))
+    elif len(matching) > 1:
+        # If there's more than one match, then case matters, so return as-is
```
elongl, line 79: Not sure if it matters, but technically this still doesn't mean that there was a match.
I can search for `heLLo` and `Hello` and `hellO` matched, and I'd return `heLLo` regardless.

elongl (approved):

macros/commands/generate_json_schema_test.sql
haritamar: Hmm yeah you're right
Maybe I should just fail here

haritamar: Or I can do a case-sensitive search in that case

haritamar: But it's an edge case hh

elongl: Yeah, generally I wouldn't have pointed this out, just sharing since we already consider the casing.
I think we can just look for a lower-case match and that should be enough for now.

haritamar: Fixed to simpler logic that is more correct (I think)

macros/commands/generate_json_schema_test.sql
haritamar: done

PR #187, opened 2022-12-14
fix databricks create table as bug

IDoneShaveIt:
dbt-spark doesn't have an adapter for `create_table_as`.
This means that every time we run `create_or_replace` macro on spark, we use the default `create_table_as` which only uses create.
This causing a bug when running a test twice, we try and create an existing table and gets an error.
The fix is to drop the table if exists before creating it (same as we do in redshift)

### DIFF
--- a/integration_tests/run_e2e_tests.py
+++ b/integration_tests/run_e2e_tests.py
@@ -393,7 +393,7 @@ def e2e_tests(target, test_types, clear_tests) -> TestResults:
         test_results.extend(results)
 
     if "create_table" in test_types:
-        # If there is a ptoblem with create_or_replace macro, it will crash the test.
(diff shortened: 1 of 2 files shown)


elongl (approved):

integration_tests/run_e2e_tests.py
```diff
@@ -392,6 +392,11 @@ def e2e_tests(target, test_types, clear_tests) -> TestResults:
         ]
         test_results.extend(results)
 
+    if "create_table" in test_types:
+        # If there is a ptoblem with create_or_replace macro, it will crash the test.
```
elongl, line 396: Typo!

elongl, line 396:
```suggestion
        # If there is a problem with create_or_replace macro, it will crash the test.
```

PR #191, opened 2022-12-18
Handling case in which invocation vars contain non-serializable data.

elongl:
Ideally, I would've used `json.dumps(..., default=str)`.
However, this is not available in the Jinja context and therefore we have to replicate its logic.


IDoneShaveIt (approved):
LGTM

PR #192, opened 2022-12-18
Installing dbt-utils based on dbt version.

elongl:


packages.yml
```diff
@@ -1,3 +1,9 @@
 packages:
   - package: dbt-labs/dbt_utils
-    version: [">=0.8.0", "<2.0.0"]
+    version: |
+      {%- set minor_to_utils_range_map = {
+        "0": [">=0.8.0", "<0.9.0"],
+        "1": [">=0.8.0", "<0.9.0"],
+        "2": [">=0.9.0", "<1.0.0"],
+      } -%}
+      {{- minor_to_utils_range_map.get(dbt_version.split('.')[1], [">=1.0.0", "<2.0.0"]) -}}
```
IDoneShaveIt, line 9: We want to support lower versions in case the user has a different package with lower dbt utils, so the lower version should always be >=0.8.0

IDoneShaveIt:
```suggestion
      {{- minor_to_utils_range_map.get(dbt_version.split('.')[1], [">=0.8.0", "<2.0.0"]) -}}
```

packages.yml
```diff
@@ -1,3 +1,9 @@
 packages:
   - package: dbt-labs/dbt_utils
-    version: [">=0.8.0", "<2.0.0"]
+    version: |
+      {%- set minor_to_utils_range_map = {
+        "0": [">=0.8.0", "<0.9.0"],
+        "1": [">=0.8.0", "<0.9.0"],
+        "2": [">=0.9.0", "<1.0.0"],
```
IDoneShaveIt, line 7:
```suggestion
        "2": [">=0.8.0", "<1.0.0"],
```

PR #200, opened 2022-12-20
add invocation id to elementary test results

IDoneShaveIt:


macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -34,6 +34,7 @@
       {% set status = "pass" if failures == 0 else result.status %}
       {% do elementary_test_results_row.update({'status': status, 'failures': failures}) %}
```
elongl, line 35: Maybe put it in the `update()`?
The `setdefault` implies that it could be that the materialization would already place this key.

PR #202, opened 2022-12-22
Anomaly tests granular time buckets

elongl:


elongl:
Here's [the documentation](https://github.com/elementary-data/elementary/pull/490).

macros/edr/system/system_utils/buckets_cte.sql
```diff
@@ -0,0 +1,58 @@
+{% macro complete_buckets_cte(time_bucket) %}
+    {{ adapter.dispatch('complete_buckets_cte','elementary')(time_bucket) }}
+{% endmacro %}
```
haritamar, line 3: Just adding a note to make sure these methods are well tested on all platforms including Databricks

macros/edr/system/system_utils/buckets_cte.sql
```diff
@@ -0,0 +1,58 @@
...
+        where edr_bucket_end < {{ elementary.cast_as_timestamp(elementary.quote(elementary.get_max_bucket_end())) }}
+    {%- endset %}
+    {{ return(complete_buckets_cte) }}
+{% endmacro %}
+
+{% macro redshift__complete_buckets_cte(time_bucket) %}
+    {%- set complete_buckets_cte %}
+      with integers as (
+        select (row_number() over (order by 1)) - 1 as num
+        from pg_catalog.pg_class
```
haritamar, line 48: Hh what is this sorcery? ;)
The trick is nice but what happens if pg_class doesn't have enough rows?

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -45,15 +45,15 @@
...
     'long_string_size': 65535,
     'collect_model_sql': true,
     'model_sql_max_size': 10240,
     'query_max_size': 1000000,
     'insert_rows_method': 'max_query_size',
     'cache_artifacts': true,
     'project_name': none,
-    'elementary_full_refresh': false
+    'elementary_full_refresh': false,
+    'training_set_size': 14
```
haritamar, line 56: Is this currently used?

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -104,14 +104,14 @@
...
-                count(metric_value) over (partition by metric_name, full_table_name, column_name, dimension, dimension_value order by edr_daily_bucket asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_set_size,
-                last_value(bucket_end) over (partition by metric_name, full_table_name, column_name, dimension, dimension_value order by edr_daily_bucket asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) training_end,
-                first_value(bucket_end) over (partition by metric_name, full_table_name, column_name, dimension, dimension_value order by edr_daily_bucket asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_start
-            from daily_buckets left join
-                grouped_metrics on (edr_daily_bucket = bucket_end)
+                edr_bucket_end,
+                avg(metric_value) over (partition by metric_name, full_table_name, column_name, dimension, dimension_value order by edr_bucket_end asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_avg,
```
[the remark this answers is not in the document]
haritamar: Unless you intended to use `training_set_size` here? not sure

models/edr/data_monitoring/anomaly_detection/metrics_anomaly_score.sql
```diff
@@ -32,16 +26,13 @@ time_window_aggregation as (
...
-    {{ dbt_utils.group_by(13) }}
-
+        avg(metric_value) over (partition by metric_name, full_table_name, column_name order by bucket_start asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_avg,
+        stddev(metric_value) over (partition by metric_name, full_table_name, column_name order by bucket_start asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_stddev,
+        count(metric_value) over (partition by metric_name, full_table_name, column_name order by bucket_start asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_set_size,
+        last_value(bucket_end) over (partition by metric_name, full_table_name, column_name order by bucket_start asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) training_end,
+        first_value(bucket_end) over (partition by metric_name, full_table_name, column_name order by bucket_start asc rows between {{ elementary.get_config_var('days_back') }} preceding and current row) as training_start
```
haritamar, line 33: Same comment regarding `days_back`

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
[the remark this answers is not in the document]
elongl: You're absolutely right!

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -1,20 +1,31 @@
...
+      floor({{ elementary.datediff(min_bucket_start, elementary.cast_as_timestamp(timestamp_column), time_bucket.period) }} / {{ time_bucket.count }}) * {{ time_bucket.count }}
+    {% endset %}
+
     {% if timestamp_column %}
-        with filtered_monitored_table as (
+        with buckets as (
+          select
+            edr_bucket_start,
+            edr_bucket_end,
+            1 as joiner
```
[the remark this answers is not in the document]
elongl: I also didn't quite understand initially so maybe we should change that.
This is needed in order to join every dimension with every bucket (many-to-many).

macros/edr/system/system_utils/buckets_cte.sql
elongl: It would be problematic.
However, I looked up various [system catalogs](https://www.postgresql.org/docs/current/catalogs.html) and this one seemed to have the biggest row count.

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
```diff
@@ -1,21 +1,30 @@
...
-            , {{ elementary.null_timestamp() }} as edr_bucket
+            , {{ elementary.null_timestamp() }} as start_bucket_in_data
             {%- endif %}
         from {{ monitored_table_relation }}
         where
         {% if timestamp_column -%}
-            {{ elementary.cast_as_timestamp(timestamp_column) }} >= {{ elementary.cast_as_timestamp(min_bucket_start) }}
-            and {{ elementary.cast_as_timestamp(timestamp_column) }} <= {{ elementary.cast_as_timestamp(max_bucket_end) }}
+            {{ elementary.cast_as_timestamp(timestamp_column) }} >= (select min(edr_bucket_start) from buckets)
+            and {{ elementary.cast_as_timestamp(timestamp_column) }} < (select max(edr_bucket_end) from buckets)
```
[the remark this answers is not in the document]
elongl: It's a bit harder to calculate the `max_complete_bucket_end` because in order to do that you'd have to start the bucketing from `min_bucket_start` whereas previously you just did today at 00:00.

Technically it should still be possible to do it with Jinja, like getting the min and max value of [this](https://github.com/elementary-data/dbt-data-reliability/blob/7fceb254d82343fc93901505fe98af158ce1b219/integration_tests/run_e2e_tests.py#L21), but I liked how it's contained within the buckets CTE and there's no need to re-calculate it.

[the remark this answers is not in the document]
haritamar: OK that makes sense to me, didn't think about it
Didn't realize that the "today at 00:00" logic is not correct anymore

macros/edr/system/system_utils/buckets_cte.sql
haritamar: Can we maybe add a simple check and log a warning if that's the case? (seems like it should be rare so maybe it's good enough, but just so we'll know)

elongl: Another concern is that a user might not have access to query the `pg_catalog` schema but I think that's rather unlikely. Kind of absurd that Redshift doesn't have an easy way of doing `range(start, end)`.

haritamar (approved):

macros/edr/system/system_utils/buckets_cte.sql
haritamar: sababa we can start with that and learn
another thing you could do I guess is to create a temp table which has the range, but maybe overkill

PR #203, opened 2022-12-22
remove selected_resources from dbt_invocations yaml

nzewail:
Noticed that `selected_resources` isn't a column in the `dbt_invocations` model


elongl:
The description does say that it's deprecated, however, it was deprecated rather quickly so it makes sense to me that we'd remove it. Merging! Thanks.

PR #214, opened 2023-01-01
E2e tests no seed

elongl:


.gitignore
```diff
@@ -4,9 +4,9 @@ logs/
 scripts/
 
 .idea
+.DS_Store
 
 edr.log
-integration_tests/data
```
elongl, line 9: Not sure if we should upload all the data into the repo.
This is better because it creates clear fixtures.
The only downside is that `dbt deps` downloads the entire repo.

IDoneShaveIt (approved):
LGTM!

PR #219, opened 2023-01-03
Freshness v2

haritamar:
This PR revamps our existing freshness test, and splits it into two tests:
1. `freshness_anomalies` - the goal of this test is to test the **expected time between updates**, and alert if the freshness during the detection period deviates from it.
2. `event_freshness_anomalies` - the goal of this test is to monitor freshness of event data. Since this is continuous / streaming data normally, monitoring time between updates as we do in the first test is not useful. So instead, we monitor the freshness as the time since the event timestamp, where there are two options:

- If an update timestamp is supplied, the freshness is defined as the maximum value of (`update_timestamp` - `event_timestamp`).
- Otherwise, we continuously collect the "current freshness" - e.g. `current_timestamp() - max(event_timestamp)`

### DIFF
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -24,7 +24,7 @@ vars:
   disable_dbt_artifacts_autoupload: true
 
 seeds:
-  +schema: test_seeds_freshness_v2
+  +schema: test_seeds_v2
 
 models:
   elementary:
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -24,7 +24,7 @@ vars:
   disable_dbt_artifacts_autoupload: true
 
 seeds:
-  +schema: test_seeds
+  +schema: test_seeds_freshness_v2
 
 models:
   elementary:
--- a/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
@@ -185,20 +185,13 @@
         order by 1
     ),
 
-    consecutive_updates as (
-        select
-            timestamp_val as update_timestamp,
-            lag(timestamp_val) over (order by timestamp_val) as prev_timestamp,
-        from unique_timestamps
-        where timestamp_val >= (select min(edr_bucket_start) from buckets)
-    ),
-
     -- compute freshness for every update as the time difference from the previous update
     consecutive_updates_freshness as (
         select
-            update_timestamp,
-            {{ elementary.timediff('second', 'prev_timestamp', 'update_timestamp') }} as freshness
-        from consecutive_updates
+            timestamp_val as update_timestamp,
+            {{ elementary.timediff('second', 'lag(timestamp_val) over (order by timestamp_val)', 'timestamp_val') }} as freshness
+        from unique_timestamps
+        where timestamp_val >= (select min(edr_bucket_start) from buckets)
     ),
 
     -- divide the freshness metrics above to buckets
--- a/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
@@ -196,7 +196,7 @@
     bucketed_consecutive_updates_freshness as (
         select
             edr_bucket_start, edr_bucket_end, update_timestamp, freshness
-        from buckets, consecutive_updates_freshness
+        from buckets cross join consecutive_updates_freshness
         where update_timestamp >= edr_bucket_start AND update_timestamp < edr_bucket_end
     ),
 
--- a/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
@@ -200,15 +200,15 @@
         where update_timestamp >= edr_bucket_start AND update_timestamp < edr_bucket_end
     ),
 
-    -- we also want to record the freshness at the end of each bucket as an additional point (as it might be larger,
-    -- or even the only point if no data was updated at all in a particular bucket)
+    -- we also want to record the freshness at the end of each bucket as an additional point. By this we mean
+    -- the time that passed since the last update in the bucket and the end of the bucket.
     bucket_end_freshness as (
         select
             edr_bucket_start,
             edr_bucket_end,
             max(timestamp_val) as update_timestamp,
             {{ elementary.timediff('second', elementary.cast_as_timestamp('max(timestamp_val)'), "least(edr_bucket_end, {})".format(elementary.current_timestamp_column())) }} as freshness
-        from buckets, unique_timestamps
+        from buckets cross join unique_timestamps
         where timestamp_val < edr_bucket_end
         group by 1,2
     ),
@@ -225,7 +225,7 @@
     bucket_freshness_ranked as (
         select
             *,
-            rank () over (partition by edr_bucket_end order by freshness desc) as row_number
+            rank () over (partition by edr_bucket_end order by freshness desc) as rank
         from bucket_all_freshness_metrics
     )
 
@@ -236,7 +236,7 @@
         {{ elementary.cast_as_string('update_timestamp') }} as source_value,
         freshness as metric_value
     from bucket_freshness_ranked
-    where row_number = 1
+    where rank = 1
 {% else %}
     {# Update freshness test not supported when timestamp column is not provided #}
     {# TODO: We can enhance this test for models to use model_run_results in case a timestamp column is not defined #}
--- a/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
@@ -191,3 +191,18 @@
     {% do return(none) %}
 {% endif %}
 {% endmacro %}
+
+{% macro sla_metric_query(metric_args, timestamp_column) %}
+{% set data_time_column = metric_args.data_time_column %}
+{% set insertion_time_column = metric_args.insertion_time_column %}
+
+{% if timestamp_column %}
+{% else %}
+
+    select
+        {{ elementary.const_as_string('sla') }} as metric_name,
+        {{ elementary.timediff('second', elementary.cast_as_timestamp("max('{}')".format(metric_args.data)), elementary.current_timestamp_column()) }} as metric_value
+    from monitored_table
+    group by 1
+{% endif %}
+{% endmacro %}
--- a/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
@@ -1,10 +1,6 @@
-{% macro table_monitoring_query(monitored_table_relation, timestamp_column, min_bucket_start, table_monitors, freshness_column, where_expression, time_bucket) %}
+{% macro table_monitoring_query(monitored_table_relation, timestamp_column, min_bucket_start, table_monitors, where_expression, time_bucket, metric_args) %}
 
     {% set full_table_name_str = elementary.quote(elementary.relation_to_full_name(monitored_table_relation)) %}
-
-    {% set metric_args = {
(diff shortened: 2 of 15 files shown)


macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -1,78 +1,33 @@
...
-        union_metrics as (
-
-        select * from row_count
-        union all
-        select * from table_freshness
-
+        metrics as (
+            {{ elementary.get_unified_metrics_query(metrics=table_monitors,
+                                                    metric_args=metric_args,
+                                                    timestamp_column=timestamp_column) }}
```
haritamar, line 30: I did a bit of a refactor here - instead of putting the queries for the different metrics in the body of the function, each is implemented in its own macro below and `get_unified_metrics_query` returns the union of them.

I felt it's needed since there is an additional metric, and the existing freshness metric has a more complex query now.

macros/edr/tests/test_utils/get_anomaly_query.sql
```diff
@@ -50,3 +50,13 @@
...
+
+{%- macro is_score_anomalous_condition(sensitivity) -%}
+    {%- set spikes_only_metrics = ['freshness', 'event_freshness'] -%}
+
+    case when metric_name IN {{ elementary.strings_list_to_tuple(spikes_only_metrics) }} then
+            anomaly_score > {{ sensitivity }}
+         else
+            abs(anomaly_score) > {{ sensitivity }}
+         end
+{%- endmacro -%}
```
haritamar, line 62: For freshness, a lower metric value is a good thing.
So we only want to detect spikes (positive anomaly score beyond threshold)

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -1,78 +1,33 @@
-{% macro table_monitoring_query(monitored_table_relation, timestamp_column, min_bucket_start, table_monitors, freshness_column, where_expression, time_bucket) %}
+{% macro table_monitoring_query(monitored_table_relation, timestamp_column, min_bucket_start, table_monitors, where_expression, time_bucket, metric_args) %}
```
haritamar, line 1: The main changes are in this file:

1. I refactored the function a bit, so each metric implementation is in its own macro below - I felt it's needed as I added more logic here (in the beginning I thought to do an even bigger refactor but decided to wait with it...)
2. There are 3 metrics defined below - `row_count`, `freshness` (modified metric), and `event_freshness` (a new metric)

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -144,4 +91,183 @@
...
+
+{% if update_timestamp_column %}
+    select
+        edr_bucket_start,
+        edr_bucket_end,
+        {{ elementary.const_as_string('event_freshness') }} as metric_name,
+        {{ elementary.cast_as_string('max({})'.format(event_timestamp_column)) }} as source_value,
+        {{ 'coalesce(max({}), {})'.format(
+                elementary.timediff('second', elementary.cast_as_timestamp(event_timestamp_column), elementary.cast_as_timestamp(update_timestamp_column)),
+                elementary.timediff('second', 'edr_bucket_start', 'edr_bucket_end')
```
haritamar, line 262: I was a bit unsure what should I do for the `event_freshness` metric if a particular bucket is empty, since the calculation is local per row.
I thought that maybe saying the freshness is the entire duration of the bucket is good enough (for event data, it's like "infinity"), but maybe there's a better option.

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -144,4 +91,183 @@
...
+
+    {%- set metric_query = metric_macro(metric_args, timestamp_column=timestamp_column) %}
+    {%- if not metric_query %}
+        {%- do return(none) %}
+    {%- endif %}
+
+    {{ metric_query }}
+{% endmacro %}
+
+{% macro row_count_metric_query(metric_args, timestamp_column=none) %}
```
haritamar, line 146: Same logic as before here, just moved.

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -1,78 +1,33 @@
-{% macro table_monitoring_query(monitored_table_relation, timestamp_column, min_bucket_start, table_monitors, freshness_column, where_expression, time_bucket) %}
+{% macro table_monitoring_query(monitored_table_relation, timestamp_column, min_bucket_start, table_monitors, where_expression, time_bucket, metric_args) %}
 
     {% set full_table_name_str = elementary.quote(elementary.relation_to_full_name(monitored_table_relation)) %}
 
+    with monitored_table as (
+        select * from {{ monitored_table_relation }}
+        {% if where_expression %}
+        where {{ where_expression }}
+        {% endif %}
```
haritamar, line 9: I fixed an unrelated bug here (where_expression was only implemented when a timestamp is defined)

.github/workflows/test-warehouse.yml
```diff
@@ -101,4 +101,4 @@ jobs:
         working-directory: ${{ env.TESTS_DIR }}
         run: |
           dbt deps
-          python run_e2e_tests.py -t "${{ inputs.warehouse-type }}" --clear-tests "${{ inputs.clear-tests }}"
+          python run_e2e_tests.py -g -t "${{ inputs.warehouse-type }}" --clear-tests "${{ inputs.clear-tests }}"
```
haritamar, line 104: Temporarily doing this because I need to regenerate the seeds for this branch (I added a column)

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -144,4 +91,183 @@
...
+        from monitored_table
+        order by 1
+    ),
+
+    -- compute freshness for every update as the time difference from the previous update
+    consecutive_updates_freshness as (
+        select
+            timestamp_val as update_timestamp,
+            lag(timestamp_val) over (order by timestamp_val) as prev_timestamp,
+            {{ elementary.timediff('second', 'prev_timestamp', 'update_timestamp') }} as freshness
```
rantibi, line 193: This command doesn't works in BQ.
I get an error `Unrecognized name: prev_timestamp at [121:20]`
Seems like you can't reference to field that defined in the same select.
changing the query to the following solved the issue
`timestamp_diff(lag(timestamp_val) over (order by timestamp_val), timestamp_val, second)`

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -1,78 +1,33 @@
...
-        with buckets as (
+        buckets as (
             select edr_bucket_start, edr_bucket_end from ({{ elementary.complete_buckets_cte(time_bucket) }})
             where edr_bucket_start >= {{ elementary.cast_as_timestamp(min_bucket_start) }}
         ),
 
-        filtered_monitored_table as (
+        time_filtered_monitored_table as (
             select *,
                    {{ elementary.get_start_bucket_in_data(timestamp_column, min_bucket_start, time_bucket) }} as start_bucket_in_data
```
rantibi, line 20: in BQ the complied CTE that I get is:
```
ime_filtered_monitored_table as (
            select *,
                   cast(

        datetime_add(
            cast( '2023-01-02 00:00:00+00:00' as datetime),
        interval cast(
      floor(

    datetime_diff(
        cast(cast(timestamp as TIMESTAMP) as datetime),
        cast('2023-01-02 00:00:00+00:00' as datetime),
        day
    )
```
and it raise the following error in runtime:
`Invalid datetime string "2023-01-02 00:00:00+00:00"`
removing `+00:00` manually solved the issue.
It relevant only for `datetime` casting, for timestamp no change required

integration_tests/dbt_project.yml
```diff
@@ -24,7 +24,7 @@ vars:
   disable_dbt_artifacts_autoupload: true
 
 seeds:
-  +schema: test_seeds
+  +schema: test_seeds_v2
```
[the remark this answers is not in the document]
haritamar: I planned to revert this, but thought to do it after merging (so PRs won't conflict with each other before).
WDYT?

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
haritamar: Thanks, this is a great idea.
I thought about doing a larger refactor here but decided to hold off to another PR (and also break down the other anomaly queries while we're at it)

One thing that was in my mind though is that we may want at some point to be able to calculate metrics regardless of the tests being defined (e.g. for a data quality dashboard).

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -150,4 +91,182 @@
...
+            {{ elementary.timediff('second', 'lag(timestamp_val) over (order by timestamp_val)', 'timestamp_val') }} as freshness
+        from unique_timestamps
+        where timestamp_val >= (select min(edr_bucket_start) from buckets)
+    ),
+
+    -- divide the freshness metrics above to buckets
+    bucketed_consecutive_updates_freshness as (
+        select
+            edr_bucket_start, edr_bucket_end, update_timestamp, freshness
+        from buckets, consecutive_updates_freshness
```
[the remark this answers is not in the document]
haritamar: 👍 
I'll change it

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -150,4 +91,182 @@
...
+        union all
+        select * from bucket_end_freshness
+    ),
+
+    -- get all the freshness values, ranked by size (we use partition by and not group by, because we also want to have
+    -- the associated timestamp as source value)
+    bucket_freshness_ranked as (
+        select
+            *,
+            rank () over (partition by edr_bucket_end order by freshness desc) as row_number
```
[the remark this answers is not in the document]
haritamar: Ah you're right
I'll change

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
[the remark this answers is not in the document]
haritamar: It's technically a "cross join" since multiple timestamps can match per row

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -150,4 +91,182 @@
...
+            edr_bucket_start,
+            edr_bucket_end,
+            max(timestamp_val) as update_timestamp,
+            {{ elementary.timediff('second', elementary.cast_as_timestamp('max(timestamp_val)'), "least(edr_bucket_end, {})".format(elementary.current_timestamp_column())) }} as freshness
+        from buckets, unique_timestamps
+        where timestamp_val < edr_bucket_end
+        group by 1,2
+    ),
+
+    -- create a single table with all the freshness values
```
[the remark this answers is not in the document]
haritamar: Should this comment be under the `consecutive_updates_freshness` CTE? just to make sure I understand it correctly.

In any case - not 100% sure about it because technically the lag can be much larger then the entire duration of the bucket (e.g. if the bucket is on the 16th and the last update was on the 14th, the freshness metric will be 2.5 days).

But maybe it's not consistent with what I did in `event_freshness` - maybe we can discuss quickly in the office

integration_tests/dbt_project.yml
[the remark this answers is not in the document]
elongl: Sorry, not sure I understand, why would they conflict?
Anyhow, as long as you'd revert it, I don't mind 😄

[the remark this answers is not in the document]
haritamar: Actually now that I'm thinking about it, I'm not sure 😅 
I remember that when I changed the seeds it caused other PRs to break, I thought it's related to the order of the columns but now I'm not sure that's the case.

PR #226, opened 2023-01-06
Added support to define a zero for the test result limit

oravi:


elongl (approved):

elongl:
There's a problem here that `dimension_anomalies` would still sample the data.

PR #227, opened 2023-01-07
Atomic dbt artifacts upload

haritamar:


models/edr/dbt_artifacts/dbt_exposures.sql
```diff
@@ -1,8 +1,8 @@
 {{
   config(
-    materialized='table',
+    materialized='incremental',
     transient=False,
-    post_hook='{{ elementary.upload_dbt_exposures() }}'
+    post_hook=after_commit('{{ elementary.upload_dbt_exposures() }}')
```
haritamar, line 5: `after_commit` is needed because I'm using transactions (affects only Redshift, as it's the only adapter that actually uses transactions)

models/edr/dbt_artifacts/dbt_exposures.sql
```diff
@@ -1,8 +1,8 @@
 {{
   config(
-    materialized='table',
+    materialized='incremental',
```
haritamar, line 3: I checked, and the default strategy for all adapters is either append or merge - which means it should be safe concurrently

macros/utils/table_operations/replace_data.sql
```diff
@@ -0,0 +1,43 @@
+{% macro replace_data(relation, sql_query) %}
+    {{ return(adapter.dispatch('replace_data', 'elementary')(relation, sql_query)) }}
+{% endmacro %}
+
+{% macro replace_data_with_table_contents(relation, replacing_relation) %}
+    {% do return(elementary.replace_data(relation, 'select * from {}'.format(replacing_relation))) %}
+{% endmacro %}
+
+{# Databricks / Spark (non-atomic implementation) #}
+{% macro default__replace_data(temporary, relation, sql_query) %}
```
haritamar, line 10: No transactions in databricks, and in any case I thought it's good to have a generic default implementation, even if it's not atomic

macros/edr/dbt_artifacts/get_dbt_artifacts_materialization.sql
```diff
@@ -0,0 +1,14 @@
+{% macro get_dbt_artifacts_materialized() %}
+    {{ return(adapter.dispatch('get_dbt_artifacts_materialized', 'elementary')()) }}
+{% endmacro %}
+
+{% macro default__get_dbt_artifacts_materialized() %}
+    {# Incremental - to avoid dropping and creating the table #}
+    {% do return('incremental') %}
```
haritamar, line 7: I verified - and the default incremental strategy is append or merge - so it should be safe on concurrent runs

macros/utils/table_operations/replace_data.sql
```diff
@@ -0,0 +1,43 @@
...
+    {% do return(elementary.replace_data(relation, 'select * from {}'.format(replacing_relation))) %}
+{% endmacro %}
+
+{# Databricks / Spark (non-atomic implementation) #}
+{% macro default__replace_data(relation, sql_query) %}
+    {% do dbt.truncate_relation(relation) %}
+    {% do dbt.run_query(elementary.insert_as_select(relation, sql_query)) %}
+{% endmacro %}
+
+{% macro snowflake__replace_data(relation, sql_query) %}
```
haritamar, line 15: Not sure if it's critical, but technically in both Snowflake and BQ there's a MERGE statement - that may allow doing this without a transaction.
I thought maybe this is simpler and more clear though.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -1,18 +1,39 @@
-{% macro upload_artifacts_to_table(table_relation, artifacts, flatten_artifact_callback, truncate_if_on_run_end, should_commit=False) %}
+{% macro upload_artifacts_to_table(table_relation, artifacts, flatten_artifact_callback, append=False, should_commit=False) %}
     {% set flatten_artifact_dicts = [] %}
     {% for artifact in artifacts %}
         {% set flatten_artifact_dict = flatten_artifact_callback(artifact) %}
-        {% if flatten_artifact_dict is not none %}
+        {% if flatten_artifact_dict is not none and elementary.dict_has_nonempty_values(flatten_artifact_dict) %}
```
elongl, line 5: What is this check needed for?
When would `dict_has_nonempty_values()` be `true`?

models/edr/dbt_artifacts/dbt_exposures.sql
```diff
@@ -1,6 +1,6 @@
...
```
elongl, line 3: Why is there a need to change the materialization?

elongl (commented):
Looks great, just two questions if you will.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -1,18 +1,39 @@
...
     {% endif %}
+{% endmacro %}
+
+{% macro overwrite_artifacts_table(table_relation, flatten_artifact_dicts, should_commit) %}
+    {% do return(adapter.dispatch('overwrite_artifacts_table', 'elementary')(table_relation, flatten_artifact_dicts, should_commit)) %}
+{% endmacro %}
+
+{% macro default__overwrite_artifacts_table(table_relation, flatten_artifact_dicts, should_commit) %}
+    {# First upload everything to a temp table #}
+    {% set temp_table_suffix = modules.datetime.datetime.utcnow().strftime('__tmp_%y%m%d%H%M%S%f') %}
```
elongl, line 24: Why not just `__tmp`?
The reason I'm asking is that Postgres has a maximum of 63 characters for table names so we'd generally rather avoid long table names.
It caused a bit of problems with the anomaly tests because certain models resolved to the same table names after truncation (`<TRUNCATED_TEST_ID>__metrics`) even though they weren't the same test.

```py
In [203]: len(datetime.datetime.utcnow().strftime('__tmp_%y%m%d%H%M%S%f'))
Out[203]: 24 (0x18)
```

models/edr/dbt_artifacts/dbt_exposures.sql
haritamar: So my reasoning is that the table materialization actually drops the table, where by default the incremental materialization only does [append or merge](https://docs.getdbt.com/docs/build/incremental-models#about-incremental_strategy) - which in this case is equivalent to doing nothing.

So changing to incremental helps to make it atomic, as basically all the logic happens in `upload_artifact_to_table` (which is also why I always truncate now, regardless of whether it runs from a post-hook or the on-run-end hook)

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
haritamar: I think you're probably right, just did it as an extra measure of safety.
Also IIRC `dbt.make_temp_relation` in postgres already adds a date so there's no sense in adding it twice...

So I'll remove it.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -1,18 +1,38 @@
...
+{% endmacro %}
+
+{% macro default__overwrite_artifacts_table(table_relation, flatten_artifact_dicts, should_commit) %}
+    {# First upload everything to a temp table #}
+    {% set temp_table_relation = dbt.make_temp_relation(table_relation) %}
+    {% do elementary.create_table_like(temp_table_relation, table_relation, temporary=True) %}
+    {% do elementary.insert_rows(temp_table_relation, flatten_artifact_dicts, should_commit, elementary.get_config_var('dbt_artifacts_chunk_size')) %}
+
+    {# Now atomically replace the data #}
+    {% do elementary.replace_data_with_table_contents(table_relation, temp_table_relation) %}
```
elongl, line 29: Why replace the data and not rename the relation like the materialization does?
It's much more cheap.

haritamar: @elongl so I read somewhere that recreating the table might cause views defined over the table to stop working. But actually I didn't verify that.
WDYT?

haritamar: It may be better though to rename like the materialization does, just need to make sure I can rename under a transaction in all platforms (as it requires two renames)

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
haritamar: I'm guessing you mean when it would be `false`?
I'm not sure to be honest
I added it to replace `remove_empty_rows` which did the same logic in SQL after the table was created.

elongl: Maybe let's remove this if we don't understand it?
I don't like how we have "paranoid" code that we can't give a concrete reason for its existence.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
elongl: Given that's what dbt do for materlializing models, I feel comfortable doing the same.
I think that `replace_data` would be more expensive in terms of performance and storage (data duplication) on larger projects.

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
haritamar: 👍 
I'll change

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
haritamar: So what I'm worried about is that unlike dbt, I'm not doing this as a part of a materialization
It can happen in any dbt run
So not sure I want to risk that views depending on the artifacts will stop working...
I can check though that this is indeed what happens.

haritamar: (I mean that with regular models it's usually clear that if you change a model you have to run models that depend on it too)

PR #229, opened 2023-01-11
Extracting db and schema name of tested table from model if possible

oravi:


macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -112,6 +111,19 @@
...
+
+    {% set primary_test_model_database = none %}
+    {% set primary_test_model_schema = none %}
+    {%- if primary_test_model_id -%}
+        {% set tested_model_node = elementary.get_node(primary_test_model_id) %}
+        {% set primary_test_model_database = tested_model_node.get('database') %}
+        {% set primary_test_model_schema = tested_model_node.get('schema') %}
+    {%- endif -%}
+
+    {%- if primary_test_model_database is none or primary_test_model_schema is none -%}
```
oravi, line 123: The reason I have this safety is because for Singular tests sometimes we have more than one model so in that case I'm taking it from the test node and not model node

PR #230, opened 2023-01-12
Postgres integration

elongl:


macros/edr/alerts/anomaly_detection_description.sql
```diff
@@ -9,21 +9,21 @@
 {% endmacro %}
 
 {% macro freshness_description() %}
-    'Last update was at ' || anomalous_value || ', ' || abs(round(metric_value/3600,2)) || ' hours ago. Usually the table is updated within ' || abs(round(training_avg/3600,2)) || ' hours.'
+    'Last update was at ' || anomalous_value || ', ' || abs(round({{ elementary.cast_as_numeric('metric_value/3600') }}, 2)) || ' hours ago. Usually the table is updated within ' || abs(round({{ elementary.cast_as_numeric('metric_value/3600') }}, 2)) || ' hours.'
```
haritamar, line 12: Guessing it's a mistake - the second should be `training_avg` right?

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -10,7 +10,7 @@
             edr_bucket_start,
             edr_bucket_end,
             1 as joiner
-          from ({{ elementary.complete_buckets_cte(time_bucket) }})
+          from ({{ elementary.complete_buckets_cte(time_bucket) }}) rs
```
haritamar, line 13: What does "rs" stand for?

macros/edr/metadata_collection/get_tables_from_information_schema.sql
```diff
@@ -68,3 +67,28 @@
...
+    )
+
+    select
+        {{ elementary.full_table_name() }} as full_table_name,
+        upper(database_name || '.' || schema_name) as full_schema_name,
+        database_name,
+        schema_name,
+        table_name
+    from information_schema_tables
+    )
```
haritamar, line 92: Why is the query wrapped with parantheses?

macros/edr/tests/test_utils/get_compiled_test_id.sql
```diff
@@ -0,0 +1,6 @@
+{% macro get_compiled_test_id() %}
+    {% set test_node = model %}
+    {% set test_hash = test_node.unique_id.split(".")[-1] %}
+    {% set test_name = test_node.name %}
+    {% do return("{}_{}".format(test_name, test_hash)) %}
```
haritamar, line 5: Can add context on why this was needed?
I'm slightly worried about adding another type of ID for tests to our code, it's confusing :)

macros/utils/data_types/cast_column.sql
```diff
@@ -6,6 +6,10 @@
     cast({{ column }} as {{ elementary.type_float() }})
 {%- endmacro -%}
 
+{%- macro cast_as_numeric(column) -%}
+    cast({{ column }} as {{ elementary.type_numeric() }})
+{%- endmacro -%}
```
haritamar, line 11: Can add context on why casting to `type_numeric` was needed?

macros/utils/table_operations/create_or_replace.sql
```diff
@@ -4,17 +4,24 @@
...
+    {% do dbt.drop_relation_if_exists(relation) %}
+    {% do run_query(dbt.create_table_as(temporary, relation, sql_query)) %}
     {% do adapter.commit() %}
 {% endmacro %}
 
+{% macro postgres__create_or_replace(temporary, relation, sql_query) %}
+    {% do run_query("BEGIN") %}
+    {% do dbt.drop_relation_if_exists(relation) %}
+    {% do run_query(dbt.create_table_as(temporary, relation, sql_query)) %}
+    {% do run_query("COMMIT") %}
```
haritamar, line 20: I'm wondering if the Postgres implementation specifically should be in a transaction if the rest are not, WDYT?

macros/utils/table_operations/table_name_with_suffix.sql
```diff
@@ -2,7 +2,7 @@
     {% set relation_max_name_length = elementary.get_relation_max_name_length()  %}
     {% if relation_max_name_length %}
         {% set suffix_length = suffix | length %}
-        {% set table_name_with_suffix = table_name[:relation_max_name_length - suffix_length] ~ suffix %}
+        {% set table_name_with_suffix = table_name[-(relation_max_name_length - suffix_length):] ~ suffix %}
```
haritamar, line 5: Why?

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
elongl: Result-set.
I was also unaware at first, the naming is a community contribution.
Should we change it to `results`?

macros/utils/data_types/cast_column.sql
elongl: The `round()` function of Postgres expects this data type as its first argument.

```sql
Database Error
  function round(double precision, integer) does not exist
  LINE 16:                 else round(execution_time, 1)
                                ^
  HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

macros/utils/table_operations/create_or_replace.sql
elongl: I don't have an alternative really.
Calling `adapter.commit()` fails because it says that I haven't started a transaction.
And not doing so just doesn't commit.

The docs says that it's okay to do so however.
image

macros/edr/metadata_collection/get_tables_from_information_schema.sql
elongl: I think the contributor just followed the others' pattern.
There doesn't seem to be a need in those parentheses.
Removed.

macros/edr/tests/test_utils/get_compiled_test_id.sql
elongl: The problem is conflicting names for Elementary's test tables.
The current table names are `{flattened_test.name}__metrics` or `{flattened_test.name}__anomaly_scores`.
Here's an example of the problem with the current situation:
```py
ipdb> t1 = ctx['graph']['nodes']['test.elementary_integration_tests.elementary_source_freshness_anomalies_training_any_type_column_anoma
lies_training_.5733415eda']
ipdb> t2 = ctx['graph']['nodes']['test.elementary_integration_tests.elementary_source_freshness_anomalies_training_string_column_anomali
es_training_.4640cc98af']
ipdb> t1['name']
'elementary_source_freshness_anomalies_training_any_type_column_anomalies_training_'
ipdb> t2['name']
'elementary_source_freshness_anomalies_training_string_column_anomalies_training_'
ipdb> prefix = '__anomaly_scores'
ipdb> max_length = 63
ipdb> t1["name"][:max_length - len(prefix)] + prefix
'elementary_source_freshness_anomalies_training___anomaly_scores'
ipdb> t2["name"][:max_length - len(prefix)] + prefix
'elementary_source_freshness_anomalies_training___anomaly_scores'
```

As you can see, two different tests (`t1`, `t2`) yield the same table name and therefore override each other and error upon running `dbt test`.

The solution is to include the test's hash into the table so that it's unique.
In addition to that, the table's name is now truncated from the end rather than from the start in order to include the hash. Alternatively we could've placed the hash first but then we also need to add dummy characters because the hash might start with digits, but table names need to start with letters.

Please let me know what you think.
Thanks.

macros/utils/table_operations/table_name_with_suffix.sql
elongl: Explanation [here](https://github.com/elementary-data/dbt-data-reliability/pull/230/files#r1070637099).

.github/workflows/test-all-warehouses.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        warehouse-type: [ snowflake, bigquery, redshift, databricks ]
+        warehouse-type: [ postgres, snowflake, bigquery, redshift, databricks ]
```
Maayan-s, line 15: Regardless of this PR, should we add Spark here?

integration_tests/README.md
```diff
@@ -0,0 +1,40 @@
+# dbt-data-reliability dbt Package Tests
```
Maayan-s, line 1: Cool!
I imagine this is for contributors? 
Should we update here as well? 
https://docs.elementary-data.com/general/contributions

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
elongl: Decided to rename it.

Maayan-s: He called it result set -> rs . 
I would prefer to use a name that describes what the result set is, makes the code more readable

integration_tests/README.md
elongl: Generally yes, but at the moment I don't think so because we have a task this sprint to improve our testing infrastructure in which hopefully we'll have something more systematic than our current script.

macros/edr/metadata_collection/get_columns_from_information_schema.sql
```diff
@@ -33,4 +32,19 @@
...
+    {%- set database_name, schema_name = schema_tuple %}
+
+    select
+        upper(table_catalog || '.' || table_schema || '.' || table_name) as full_table_name,
+        upper(table_catalog) as database_name,
+        upper(table_schema) as schema_name,
+        upper(table_name) as table_name,
+        upper(column_name) as column_name,
+        data_type
+    from information_schema.columns
```
Maayan-s, line 47: Can we change to relay on dbt here like on the default query? feels safer

macros/edr/metadata_collection/get_tables_from_information_schema.sql
```diff
@@ -7,7 +7,7 @@
     {%- set database_name, schema_name = schema_tuple %}
     {% set schema_relation = api.Relation.create(database=database_name, schema=schema_name).without_identifier() %}
 
-    (with information_schema_tables as (
+    with information_schema_tables as (
```
Maayan-s, line 10: Why was this removed? 
I think it was there for a reason

macros/edr/metadata_collection/get_tables_from_information_schema.sql
```diff
@@ -39,15 +39,12 @@
...
-    )
-
 {% endmacro %}
 
-
 {% macro redshift__get_tables_from_information_schema(schema_tuple) %}
     {%- set database_name, schema_name = schema_tuple %}
 
-    (with information_schema_tables as (
+    with information_schema_tables as (
```
Maayan-s, line 47: same

macros/edr/metadata_collection/get_tables_from_information_schema.sql
```diff
@@ -65,6 +62,27 @@
         schema_name,
         table_name
     from information_schema_tables
+{% endmacro %}
+
+{% macro postgres__get_tables_from_information_schema(schema_tuple) %}
+    {%- set database_name, schema_name = schema_tuple %}
+
+    with information_schema_tables as (
```
Maayan-s, line 70: Might be missing "(" on the beginning

macros/edr/tests/test_utils/get_compiled_test_id.sql
haritamar: Hmm I see, makes sense
Any chance you can give it a more direct name like "get_elementary_test_table_name"? (and avoid using it in logs etc)
Is it needed anywhere else?

haritamar: Also - one problem I see with truncating from the end is that you might have a problematic character in the middle of a test name (e.g. a digit)
So you may need to add something at the beginning.

Or alternatively - truncate from the beginning but make sure the hash is added anyway.

macros/utils/data_types/cast_column.sql
haritamar: 👍

macros/utils/table_operations/create_or_replace.sql
haritamar: OK got it
Didn't realize it was needed by the implementation

macros/edr/metadata_collection/get_tables_from_information_schema.sql
haritamar: I think it was because of my comment (I didn't realize why it's needed)

elongl: Moved to the [macro itself](https://github.com/elementary-data/dbt-data-reliability/pull/230/commits/6fc9b19c4d3d0e0ed1afe972b8795cab44350d30).

macros/edr/tests/test_utils/get_compiled_test_id.sql
elongl: Thanks, that's great feedback.
I changed the implementation [here](https://github.com/elementary-data/dbt-data-reliability/pull/230/commits/90500cc2bc4c7a6bc3eb026dcdb3065e427d8972).
One thing that I'm not too crazy about is that singular tests don't have a hash at their unique ID.
Generally it doesn't matter because this macro is only called on Elementary's generic tests, but I wonder if we can find a different and more stable identifier for a test that is short. Ideally I'd hash the `unique_id` but I can't because `local_md5` is only introduced in v1.4.0. This might be good enough though.

haritamar (approved):

PR #233, opened 2023-01-15
Fixed the test's description.

elongl:


IDoneShaveIt:
I think that this is not a bug.
There is dbt description, and the description elementary uses from the meta.
When showing the test description at the UI / alert, we extract it from the meta and don't use the description field.

elongl:
@IDoneShaveIt 
This is a bug in the `dbt_tests` table.

PR #235, opened 2023-01-18
add sum column anomaly check

tc-chrisbui:


Maayan-s:
Adding a consideration here - 

In the past we didn't add `sum` to the default monitors for numeric columns, as sum of very large numbers returns a result that is larger than the max int, and the DB returns an error. 
This happens on columns like `id` for example where the numbers are big and `sum` doesn't make sense.

elongl:
Hi @tc-chrisbui, could you please also add the `sum` column to the test's data in [this file](https://github.com/elementary-data/dbt-data-reliability/blob/b0bdff12136a2275aa310911df79b6981fcbc47a/integration_tests/generate_data.py#L126)?

Continuing what @Maayan-s mentioned,
I think maybe for the time being it shouldn't be part of the default monitors (change in `get_config_var`).
What do you think?

Final note, care to also contribute the documentation to that as well?
You can do so by clicking the edit button on the top on [this page](https://docs.elementary-data.com/guides/add-elementary-tests).

Thanks.

tc-chrisbui:
Thanks, @Maayan-s  and @elongl.  That makes sense to me.  I will make the changes.

elongl:
Hi @tc-chrisbui !
Thanks a lot for contributing this, and apologies for the wait.
If you're interested, I've made a couple of small changes on top of your PR [here](https://github.com/elementary-data/dbt-data-reliability/pull/277/files). Here's the gist of them:

1. We renamed `type_float` to `edr_type_float`.
2. Previously the `edr_monitors` var defined all the possible options of monitors so you couldn't have a monitor that is not activated by default, I changed this logic in `column_monitors_by_type`.
3. I added `sum` to the `numeric_column_anomalies` model so that we'd be able to test it.
4. Because the `sum` isn't included by default, I added the test argument of `column_anomalies: sum`.

PR #239, opened 2023-01-21
Package integration tests

haritamar:


macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -225,7 +225,7 @@
     bucket_freshness_ranked as (
         select
             *,
-            rank () over (partition by edr_bucket_end order by freshness desc) as rank
+            row_number () over (partition by edr_bucket_end order by freshness is null, freshness desc) as row_number
```
haritamar, line 228: Actually found a couple of bugs in the new freshness test:
1. When ordering only by `freshness desc`, if they are nulls they have the top rank.
2. rank is not good actually because it allows duplicate values to return (if several values share the highest freshness).

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -263,7 +263,7 @@
 {% else %}
     select
         {{ elementary.const_as_string('event_freshness') }} as metric_name,
-        {{ elementary.timediff('second', elementary.cast_as_timestamp("max({})".format(event_timestamp_column)), elementary.current_timestamp_column()) }} as metric_value
+        {{ elementary.timediff('second', elementary.cast_as_timestamp("max({})".format(event_timestamp_column)), elementary.quote(elementary.get_run_started_at())) }} as metric_value
```
haritamar, line 266: Wasn't really a bug, but using `get_run_started_at` makes it more testable as I can override it with `custom_run_started_at`.

integration_tests/integration_tests/test_table_monitoring_query.py
```diff
@@ -0,0 +1,138 @@
+from datetime import datetime
+from dbt_osmosis.core.osmosis import DbtProject
+from parametrization import Parametrization
```
elongl, line 3: What's the benefit of `pytest-parametrization` over Pytest's built-in functionality?

integration_tests/integration_tests/test_table_monitoring_query.py
```diff
@@ -0,0 +1,138 @@
...
+    metric="row_count",
+    timestamp_column=None,
+    input_rows=[
+        {"name": "Jessica Jones", "updated_at": "2022-01-01 13:34:29"},
+        {"name": "Luke Cage", "updated_at": "2022-01-02 13:35:28"},
+        {"name": "Luke Cage", "updated_at": "2022-01-02 15:23:28"},
+        {"name": "Luke Cage", "updated_at": "2022-01-02 17:48:28"}
+    ],
+    expected_metrics={
+        datetime(2022, 1, 4, 0, 13, 42): 4
```
elongl, line 41: Could you define a `RUN_STARTED_AT` const at the top and use it here, and also pass it to `custom_run_started_at`? I was a bit confused a bit as to how does this work.

integration_tests/integration_tests/test_table_monitoring_query.py
```diff
@@ -0,0 +1,138 @@
+from datetime import datetime
+from dbt_osmosis.core.osmosis import DbtProject
```
elongl, line 2: Why use `dbt-osmosis` over `dbt-core`?
Does it have a compatibility layer for different versions of dbt?

haritamar: So yeah I think so - they created wrappers on top of the dbt-core objects for this purpose.
In general it looks like a relatively well maintained repo and the API they provided is nice and relatively simple (but not without caveats).

I admit though that I didn't try to use dbt-core directly so we can try that as well :)

integration_tests/integration_tests/test_table_monitoring_query.py
haritamar: It's actually a wrapper on top of pytest's functionality.
Makes it a bit easier to provide names for all the cases + the default parameters is nice.
Not a must but I find it a bit more readable than pytest's existing functionality

integration_tests/integration_tests/test_table_monitoring_query.py
haritamar: Will do, you're right 👍

integration_tests/integration_tests/dbt_project.py
```diff
@@ -0,0 +1,125 @@
...
+        self.manifest.nodes.pop(f"{NodeType.SqlOperation}.{self.config.project_name}.{name}", None)
+
+
+class Args(BaseModel):
+    """
+    Minimal mock to dbt config arguments
+    """
+    project_dir: str
+    target: Optional[str] = None
+    threads: Optional[int] = 1
```
elongl, line 125: Maybe set default `threads` to 8 / 32?

haritamar: I actually did it as a safety because I saw a potential issue with accessing the adapter I create here in multiple threads, but not 100% sure about it.
In any case though I think it doesn't really impact anything - because everything I'm invoking here is synchronous.

PR #241, opened 2023-01-22
Ele 97 improve on run end performance

elongl:
Changes:

1. Added `artifact_hash` column to artifacts.
2. Added `dbt_artifacts_state` view that hash a hash on all the hashes per artifact.
3. Upon `on-run-end`, the state is pulled and compared to the current state using `local_md5`.

This means that the caching will only take place from dbt 1.4.0 onwards.


macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -52,5 +54,8 @@
         'path': node_dict.get('path'),
         'generated_at': elementary.datetime_now_utc_as_string()
       }%}
+    {% set time_excluded_dict = flatten_exposure_metadata_dict.copy() %}
+    {% do time_excluded_dict.pop("generated_at") %}
+    {% do flatten_exposure_metadata_dict.update({"artifact_hash": local_md5(time_excluded_dict | string) if local_md5 else none}) %}
```
elongl, line 59: We could also define the `artifact_hash` within the warehouse (`md5(...) as artifact_hash`) and then it'd work even if you're not on dbt-v1.4.0, but I don't see why it'd be worth the trouble.

RoiTabach:
@elongl do you think we need to add a macro/action to clear existing artifacts? 
I also wanted to ask if you want to add an e2e test on this new functionality, something like calling twice and seeing the second time not running?

elongl:
@RoiTabach 

> @elongl do you think we need to add a macro/action to clear existing artifacts?
Could you elaborate a bit on why? What's the use-case?
> f you want to add an e2e test on this new functionality, something like calling twice and seeing the second time not running?

Yeah, definitely, I wanted to add tests today.

RoiTabach:
Why do we need to add an action to clear artifacts?
It'll allow me to force upload if that's something I want (if I care about the updated_at field or I want to debug something in the DWH)

elongl:
@RoiTabach You can force upload by running the models themselves (`dbt run -s dbt_models` or `dbt run -s edr.dbt_artifacts`) because the caching only takes place from `upload_dbt_artifacts()` that is only called in the `on-run-end`.

models/edr/dbt_artifacts/dbt_artifacts_state.sql
```diff
@@ -0,0 +1,15 @@
+{{
+  config(
+    materialized = 'view',
+    bind=False
+  )
+}}
+
+select
+  (select md5({{ dbt.listagg("artifact_hash", order_by_clause="order by artifact_hash") }}) from {{ ref("dbt_models") }}) as dbt_models,
```
elongl, line 9: Need to check if `listagg` can overflow.

models/edr/dbt_artifacts/dbt_artifacts_state.sql
```diff
@@ -0,0 +1,22 @@
+{{
+  config(
+    materialized = 'view',
+    bind=False
+  )
+}}
+
+{% set hash_agg_expr %}
+md5({{ elementary.listagg("artifact_hash") }})
```
elongl, line 9: From Snowflake's docs:
> If you do not specify the WITHIN GROUP (<orderby_clause>), the order of elements within each list is unpredictable. (An ORDER BY clause outside the WITHIN GROUP clause applies to the order of the output rows, not to the order of the list elements within a row.)

From BigQuery's docs:
> The ORDER BY clause is supported only for aggregate functions that depend on the order of their input. For those functions, if the ORDER BY clause is omitted, the output is nondeterministic.

It seems that not using `ORDER BY` results in nondeterministic order.
However, in practice, doing the `ORDER BY` on the source table seems to work.

The reason for not doing it within the `listagg()` is that Databricks doesn't support it.

models/edr/dbt_artifacts/dbt_artifacts_state.sql
```diff
@@ -0,0 +1,20 @@
+{{
+  config(
+    materialized = 'view',
+    bind=False,
+    enabled=(target.type not in ['databricks', 'spark'])
+  )
+}}
+
+{% set hash_agg_expr %}
+md5({{ elementary.listagg("artifact_hash", order_by_clause="order by artifact_hash") }})
```
elongl, line 10: image

This is problematic in Redshift.
`65535 / 32 (MD5 Size) = 2048`
Meaning it'll fail on 2048+ model count.

I think I'll change the implementation to be `unique_id`, `artifact_hash` and then just diff them in Jinja.
What do you think? @haritamar

elongl: Changed [here](https://github.com/elementary-data/dbt-data-reliability/pull/241/commits/656c8e6619bf028c08d591cdb1c221e2bd3514de).

models/edr/dbt_artifacts/dbt_artifacts_state.sql
```diff
@@ -0,0 +1,25 @@
...
+  "dbt_sources",
+  "dbt_snapshots",
+  "dbt_metrics",
+  "dbt_exposures",
+  "dbt_seeds",
+] %}
+
+{% for artifact_model in artifact_models %}
+select
+  '{{ artifact_model }}' as artifacts_model,
```
oravi, line 20: Instead of returning all the hashes to the memory - let's leverage the db for doing fast hash calculations. I suggest calculating an md5 on all the hashes here (ordered by the hash itself) instead of returning the list of hashes per artifacts table

models/edr/dbt_artifacts/dbt_exposures.sql
```diff
@@ -2,7 +2,10 @@
   config(
     materialized='incremental',
     transient=False,
-    post_hook='{{ elementary.upload_dbt_exposures() }}'
+    post_hook='{{ elementary.upload_dbt_exposures() }}',
+    unique_key='unique_id',
+    on_schema_change='append_new_columns',
```
oravi, line 7: Let's talk with Itamamr about this? maybe we should use here the sync columns mode

macros/edr/system/system_utils/logs.sql
```diff
@@ -9,6 +9,8 @@
         {% set debug_logs_enabled = elementary.get_config_var('debug_logs') %}
         {% if debug_logs_enabled %}
             {{ elementary.edr_log(msg) }}
+        {% else %}
+            {{ elementary.edr_log(msg, info=False) }}
```
oravi, line 13: Are we sure about it? is it going to slow down our performance? this means all of our debug logs will be printed ot the log file right? not sure about it in terms of perf

macros/utils/run_queries/run_query.sql
```diff
@@ -0,0 +1,8 @@
+{% macro run_query(query) %}
```
oravi, line 1: the name here is a bit dangerous (I know we should use the "elementary"), I'd just create a macro called "lower_column_in_query_result" and call it on the result that gets back from the run_query macro

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -25,7 +25,9 @@
                                                                      ('package_name', 'string'),
                                                                      ('original_path', 'long_string'),
                                                                      ('path', 'string'),
-                                                                     ('generated_at', 'string')]) %}
+                                                                     ('generated_at', 'string'),
+                                                                     ('artifact_hash', 'string'),
```
oravi, line 29: cosmetics - what do you think about metadata_hash?

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -52,5 +54,6 @@
         'path': node_dict.get('path'),
         'generated_at': elementary.datetime_now_utc_as_string()
       }%}
+    {% do flatten_exposure_metadata_dict.update({"artifact_hash": elementary.get_artifact_hash(flatten_exposure_metadata_dict)}) %}
```
oravi, line 57: Why not doing it in the dict above? (all the comments here are relevant for all the different artifacts)

oravi (changes requested):

macros/edr/dbt_artifacts/get_artifact_hash.sql
```diff
@@ -0,0 +1,10 @@
+{% macro get_artifact_hash(artifact) %}
+  {% if not local_md5 %}
+    {% do return(none) %}
+  {% endif %}
+
+
+  {% set time_excluded_artifact = artifact.copy() %}
```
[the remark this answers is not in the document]
elongl: Only `copy()` since all we do is pop the timestamp value.

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
elongl: Because it's a dict literal, you can't access properties of a dict within the dict.

```py
In [118]: d = {
     ...: 'a': 1,
     ...: 'b': d['a'] + 1
     ...: }
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[118], line 3
      1 d = {
      2 'a': 1,
----> 3 'b': d['a'] + 1
      4 }

NameError: name 'd' is not defined

```

macros/utils/run_queries/run_query.sql
elongl: This is needed on nearly every query we're doing in order to support Snowflake,
so I think it's useful to have a macro that abstracts that.
I can rename the macro to convey that it's doing it, but not sure if it's a must.
(Our current way of dealing with it is by using `insensitive_get_dict_value` everywhere.)

What do you think?

models/edr/dbt_artifacts/dbt_artifacts_state.sql
```diff
@@ -0,0 +1,25 @@
+{{
```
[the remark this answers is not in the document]
elongl: Agreed!

models/edr/dbt_artifacts/dbt_artifacts_state.sql
elongl: That was my original implementation.
I changed it because of [this](https://github.com/elementary-data/dbt-data-reliability/pull/241#discussion_r1086511999) and [this](https://github.com/elementary-data/dbt-data-reliability/pull/241#discussion_r1085342913).

macros/edr/dbt_artifacts/get_artifact_hash.sql
```diff
@@ -0,0 +1,10 @@
...
+  {% do time_excluded_artifact.pop("generated_at") %}
+  {% do return(local_md5(time_excluded_artifact | string)) %}
```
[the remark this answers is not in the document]
elongl: Starting from Python 3.7, dicts are guaranteed to keep their order of declarations.

Sources:
- https://www.python.org/downloads/release/python-370/
- https://realpython.com/python37-new-features/#the-order-of-dictionaries-is-guaranteed
- https://stackoverflow.com/a/39537308/8182628

macros/edr/dbt_artifacts/upload_artifacts_to_table.sql
```diff
@@ -7,6 +7,14 @@
         {% endif %}
     {% endfor %}
 
+    {% if state_hashes and elementary.get_config_var("cache_artifacts") %}
```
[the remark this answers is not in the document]
elongl: I agree, this makes sense to me.
In that case, I'll also have to breakdown `upload_artifacts_to_table`.
Entire flow:

1. Build artifacts by flattening them.
2. Compare flattened artifacts to state hashes.
3. Upload the artifacts if there's a difference.

Note that step (1) is currently implemented in `upload_artifacts_to_table`.
Sounds good?

PR #242, opened 2023-01-23
Ele 130 synonymous data types varchar text arent

RoiTabach:
see description https://www.notion.so/elementary-data/Schema-changes-from-baseline-Test-Issues-92d5a368f32f4d2a929424990c89fb53

### DIFF
--- a/macros/utils/data_types/get_data_type_synonyms.sql
+++ b/macros/utils/data_types/get_data_type_synonyms.sql
@@ -47,3 +47,36 @@
    see https://spark.apache.org/docs/latest/sql-ref-datatypes.html #}
    {{return (exact_data_type) }}
 {% endmacro %}
+
+
+{% macro redshift__get_data_type_synonyms(exact_data_type) %}
+{# understanding Redshift data type synonyms:
+ https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html #}
+ {% set exact_data_type_to_data_type_returned_by_the_info_schema = {'INT2':	'SMALLINT',
+                                                                   'INT':	'INTEGER',
+                                                                   'INT4':	'INTEGER',
+                                                                   'INT8':	'BIGINT',
+                                                                   'NUMERIC':	'DECIMAL',
+                                                                   'FLOAT4':	'REAL',
+                                                                   'FLOAT8':	'DOUBLE PRECISION',
+                                                                   'FLOAT':	'DOUBLE PRECISION',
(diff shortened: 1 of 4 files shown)


macros/utils/data_types/get_data_type_synonyms.sql
```diff
@@ -0,0 +1,107 @@
+{% macro get_data_type_synonyms(exact_data_type) %}
+    {% set result = adapter.dispatch('get_data_type_synonyms','elementary')(exact_data_type) %}
```
haritamar, line 2: Maybe do here something like "if result is None return exact_data_type", rather than do an "if" in every implementation later?
(though maybe it's not the best practice to add logic below "adapter.dispatch", not sure, however you feel)

macros/utils/data_types/get_data_type_synonyms.sql
```diff
@@ -0,0 +1,107 @@
+{% macro get_data_type_synonyms(exact_data_type) %}
```
haritamar, line 1: Cosmetic note - "get_data_type_synonyms" sounds a bit like you return a list of synonyms, which is not what happens.
How about calling it something like "normalize_data_type"?

macros/utils/data_types/get_data_type_synonyms.sql
```diff
@@ -0,0 +1,107 @@
...
+
+{% macro bigquery__get_data_type_synonyms(exact_data_type) %}
+{# BigQuery has no concept of data type synonyms,
+ see https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types #}
+   {{return (exact_data_type) }}
+{% endmacro %}
+
+
+
+{% macro snowflake__get_data_type_synonyms(exact_data_type) %}
```
haritamar, line 18: Should we perhaps also convert to uppercase before looking up the type?

macros/utils/data_types/get_data_type_synonyms.sql
RoiTabach: In this case I think I'll not add this logic below adapter.dispatch.

macros/utils/data_types/get_data_type_synonyms.sql
RoiTabach: Good comment, changing

RoiTabach: renamed to get_normalized_data_type (left the "get_" )

PR #244, opened 2023-01-23
Ele 227 configuring a timestamp column that

Maayan-s:
Fixes https://github.com/elementary-data/elementary/issues/610

### DIFF
--- a/macros/edr/tests/test_utils/find_normalized_data_type_for_column.sql
+++ b/macros/edr/tests/test_utils/find_normalized_data_type_for_column.sql
@@ -3,7 +3,7 @@
 {% set columns_from_relation = adapter.get_columns_in_relation(model_relation) %}
 {% if columns_from_relation and columns_from_relation is iterable %}
     {% for column_obj in columns_from_relation %}
-        {% if column_obj.column | lower == column_name | lower %}
+        {% if column_obj.column | lower == column_name | lower or column_obj.column | lower == column_name[1:-1] | lower %}
             {{ return(elementary.normalize_data_type(column_obj.dtype)) }}
         {% endif %}
     {% endfor %}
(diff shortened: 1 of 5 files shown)


macros/edr/tests/test_utils/find_normalized_data_type_for_column.sql
```diff
@@ -3,7 +3,7 @@
 {% set columns_from_relation = adapter.get_columns_in_relation(model_relation) %}
 {% if columns_from_relation and columns_from_relation is iterable %}
     {% for column_obj in columns_from_relation %}
-        {% if column_obj.column | lower == column_name | lower %}
+        {% if column_obj.column | lower == column_name | lower or column_obj.column | lower == column_name[1:-1] | lower %}
```
RoiTabach, line 6: This is python, right? 
how about 
```suggestion
        {% if column_obj.column | lower == column_name | lower or column_obj.column | lower == column_name.replace('"','') | lower %}
```

Maayan-s: Made a change to be more explicit, thanks

macros/edr/tests/test_utils/find_normalized_data_type_for_column.sql
```diff
@@ -3,7 +3,7 @@
 {% set columns_from_relation = adapter.get_columns_in_relation(model_relation) %}
 {% if columns_from_relation and columns_from_relation is iterable %}
     {% for column_obj in columns_from_relation %}
-        {% if column_obj.column | lower == column_name | lower %}
+        {% if column_obj.column | lower == column_name | trim('\"') | trim('\'') | trim('\`') | lower %}
```
elongl, line 6:
```suggestion
        {% if column_obj.column | lower == column_name | trim('\'\"\`') | lower %}
```

elongl:
Have we tested that the rest of the logic works as expected?

Maayan-s:
@elongl 
I checked locally with a column as the user described and it worked. It still leaves the responsibility on the user to pass the timestamp column quoted.
But they do the same with dbt source freshness, so I think it's valid. 
The only place where we failed because of the quotes was here.

PR #248, opened 2023-01-25
replace_table_data: revert atomic upload for spark / databricks, and …

haritamar:
…do a plain truncate + replace instead


macros/utils/table_operations/replace_table_data.sql
```diff
@@ -46,7 +45,7 @@
     {% endif %}
 
     {% do elementary.create_table_like(int_relation, base_relation, temporary=temporary) %}
-    {% do elementary.insert_rows(int_relation, rows, should_commit, elementary.get_config_var('dbt_artifacts_chunk_size')) %}
+    {% do elementary.insert_rows(int_relation, rows, chunk_size=elementary.get_config_var('dbt_artifacts_chunk_size')) %}
```
RoiTabach, line 48: I think it would be more readable with expicit should_commit=False, something like this: WDYT? 

```suggestion
    {% do elementary.insert_rows(int_relation, rows, should_commit=false, chunk_size=elementary.get_config_var('dbt_artifacts_chunk_size')) %}
```

PR #251, opened 2023-01-25
Ele 206 small hourly time buckets on the anomaly

elongl:


macros/utils/data_types/data_size.sql
```diff
@@ -1,15 +1,23 @@
-{% macro get_column_size() %}
-  {{ return(adapter.dispatch('get_column_size', 'elementary')()) }}
+{% macro long_string_size() %}
```
elongl, line 1: Considering deleting this macro altogether (not used anymore) to discourage design in which this matters.
What do you think? @haritamar

macros/edr/tests/test_utils/backfill_result_rows.sql
```diff
@@ -0,0 +1,35 @@
...
+
+    {% set test_result_rows_rows = [] %}
+    {% set backfill_query %}
+        select
+        id,
+        detected_at,
+        result_rows
+        from {{ ref("elementary", "elementary_test_results") }}
+        where {{ elementary.datediff(elementary.cast_as_timestamp('detected_at'), elementary.current_timestamp(), 'day') }} < {{ elementary.get_config_var("days_back") }}
+        and result_rows is not null
```
elongl, line 20: Theoretically we should also query rows with `null` in `result_rows` and query their `test_results_query`.
However, there's a high likelihood that the test will no longer yield the same results, or even potentially crash.
In that regard, I prefer to exclude that case.

macros/edr/tests/test_utils/backfill_result_rows.sql
```diff
@@ -0,0 +1,35 @@
+{% macro backfill_result_rows() %}
```
haritamar, line 1: Some thoughts / questions about it:

- I assume we should tell in the release notes for people to execute `dbt run -s elementary` once, to create the new table?
- I'm not sure if we have performance concerns, though since it's limited by days_back it's probably fine.
- WDYT about concurrent executions of this - are we worried about it?

Not sure we should change anything, just wonder what you think about it.

elongl: 1. Yes, they'll need to run the new table. Do you think we should add a check at the `on-run-start` that the table exists? I don't want to break the pipeline but maybe it makes sense if they upgrade. Otherwise test results' will just be list.
2. Yes, that's mostly what I figured as well. Do you have an idea on how to optimize it further?
3. I don't think so because if there's a concurrent run, the materialization of the model itself would crash and therefore there wouldn't be a call to the `post-hook` (this macro). I'm validating that the table doesn't exist yet (`not is_incremental()`).

haritamar: 1. I think as long as it's clear what they need to do it may be fine (unless you think it's easy to do the on-run-start change).
2. Not sure it's needed, let's learn like Or says :)
3. Yup, makes sense to me.

PR #260, opened 2023-02-08
Fix the `DBT_ORCHESTRATOR` env var not being used.

andreqaugusto:
As the doc says ([link](https://docs.elementary-data.com/guides/modules-overview/dbt-package)), setting the `DBT_ORCHESTRATOR` environment variable should populate the `orchestrator` column in `invocations` table. However, per my experience, this is not happening.

This PR aims to fix this behaviour with a simple change.

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_invocation.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_invocation.sql
@@ -106,7 +106,7 @@
       {% do return(orchestrator) %}
     {% endif %}
   {% endfor %}
-  {% do return(none) %}
+  {% do return(elementary.get_first_env_var(["DBT_ORCHESTRATOR"])) %}
 {% endmacro %}
 
 {% macro get_dbt_invocations_empty_table_query() %}


macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -106,7 +106,7 @@
       {% do return(orchestrator) %}
     {% endif %}
   {% endfor %}
-  {% do return(none) %}
+  {% do return(elementary.get_first_env_var('DBT_ORCHESTRATOR')) %}
```
elongl, line 109: `get_first_env_var` receives a list rather than a string.

```suggestion
  {% do return(elementary.get_first_env_var(["DBT_ORCHESTRATOR"])) %}
```

elongl (changes requested):
Thanks a lot for your contribution!
Added a slight change suggestion.

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
andreqaugusto: yeah, an oversight from my part. Fixing that now!

Maayan-s:
Thank you for your contribution @andreqaugusto!!
Will be released in the next version 🎉

PR #272, opened 2023-02-19
Fixed a breaking change in dbt 1.5.0.

elongl:
`invocation_args_dict.vars` was changed from a `string` to a `dict`.

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_invocation.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_invocation.sql
@@ -60,7 +60,7 @@
         {{- return(config.args.select) -}}
     {%- else -%}
         {{- return([]) -}}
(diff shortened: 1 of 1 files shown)


macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -71,14 +71,18 @@
         {{- return(config.args.selector_name) -}}
     {%- else -%}
         {{- return([]) -}}
-    {%- endif -%})
+    {%- endif -%}
```
RoiTabach, line 74: this was an unrelated bug?

elongl: Not a bug, just shouldn't be there.

haritamar (approved):

PR #274, opened 2023-02-23
Created a state model.

elongl:
The model doesn't automatically update on the `on-run-end` because the main purpose is to learn that the Elementary models were ran.


models/edr/system/state.sql
```diff
@@ -0,0 +1,4 @@
+{{ config(materialized='table') }}
```
elongl, line 1: Not sure if `state` is necessarily the best name for it.
Perhaps `metadata` is better?

haritamar: I think `metadata` is better :)

haritamar (approved):

PR #283, opened 2023-03-02
feature/detailed output python - draft

Deathfireofdoom:
Hope it is okey, but just to give you an idea how I would like this feature.

1. Python test now takes `detailed_test_output: bool` as a parameter. 
2. For this to work, the output from the python-macro needs to be a dataframe(pandas or spark). If this is not the case, it will throw an error.
3. The outputed dataframe will be saved to a separate table.
4. Instead of just "fail_count" in the edr-report, a sql-query will have an sql-query for that table.   

This is just rough draft, but works. Things that could be nice to add is possibility to chose which columns that should be saved, and maybe a limit to avoid saving hundreds of rows.

image

https://github.com/elementary-data/elementary/issues/701


Deathfireofdoom:
Closing this, I just got what Elon said in the feature-request. Will come a new MR soon.

PR #285, opened 2023-03-06
Removing monitors that don't match the column's type.

elongl:


haritamar (approved):

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -39,12 +39,7 @@
     'edr_cli_run': false,
     'max_int': 2147483647,
     'custom_run_started_at': none,
-    'edr_monitors': {
```
Maayan-s, line 42: @elongl 
The reason this was originally a var was so users could change the defaults. 
I'm not sure this is a real need, but maybe @oravi should add his input.

elongl: It still is, I didn't change the logic here.

PR #290, opened 2023-03-14
add lower to avoid string matching error

sadahry:
To fix column_name matching error

when using `elementary.schema_changes_from_baseline` with `enforce_types: true`

before fix -> success
after fix -> error
```
cur:
  column_name: partition_hourly
  data_type: VARCHAR
pre:
  column_name: PARTITION_HOURLY
  data_type: TIMESTAMP_NTZ
```

### DIFF
--- a/macros/edr/data_monitoring/schema_changes/get_columns_changes_query.sql
+++ b/macros/edr/data_monitoring/schema_changes/get_columns_changes_query.sql
@@ -77,7 +77,7 @@
             pre.data_type as pre_data_type,
             pre.detected_at
         from cur inner join pre
-            on (cur.full_table_name = pre.full_table_name and lower(cur.column_name) = lower(pre.column_name))
+            on (lower(cur.full_table_name) = lower(pre.full_table_name) and lower(cur.column_name) = lower(pre.column_name))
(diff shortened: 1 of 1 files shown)


macros/edr/data_monitoring/schema_changes/get_columns_changes_query.sql
```diff
@@ -77,7 +77,7 @@
             pre.data_type as pre_data_type,
             pre.detected_at
         from cur inner join pre
-            on (cur.full_table_name = pre.full_table_name and cur.column_name = pre.column_name)
+            on (cur.full_table_name = pre.full_table_name and lower(cur.column_name) = lower(pre.column_name))
```
Maayan-s, line 80: Thanks @sadahry!
We should probably add the same safety to the full_table_name as well. 
Could you add to the PR?

sadahry: Sure! fixed https://github.com/elementary-data/dbt-data-reliability/pull/290/commits/9a405310e1a1d5c1baf1c9005c93ada1b20bbf65 .

Maayan-s:
Amazing, merging now! 
Thank you for contributing 🤩

PR #294, opened 2023-03-17
Fix - Decimal type databricks not recognized as numeric

Maayan-s:
Reported here -
https://elementary-community.slack.com/archives/C02CTC89LAX/p1679047291649449


RoiTabach (approved):

macros/utils/data_types/normalize_data_type.sql
```diff
@@ -1,4 +1,10 @@
 {% macro normalize_data_type(data_type) %}
+
+    {# decimal data type is in the format decimal(p,s) #}
```
RoiTabach, line 3:
```suggestion
    {# in case data type has precision info - e.g. decimal is in the format decimal(p,s) #}
```

PR #295, opened 2023-03-17
Fixed unsupported time buckets in BQ (week, month)

Maayan-s:
https://elementary-community.slack.com/archives/C02CTC89LAX/p1679055418315779


macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
```diff
@@ -1,31 +1,37 @@
 {% macro column_monitoring_query(monitored_table_relation, min_bucket_start, column_obj, column_monitors, metric_properties) %}
-    {% set full_table_name_str = elementary.edr_quote(elementary.relation_to_full_name(monitored_table_relation)) %}
+    {%- set full_table_name_str = elementary.edr_quote(elementary.relation_to_full_name(monitored_table_relation)) %}
+    {%- set min_bucket_start = elementary.edr_date_trunc(metric_properties.time_bucket.period, elementary.edr_cast_as_timestamp(min_bucket_start))%}
```
Maayan-s, line 3: date_trunc by period to the min_bucket start means the it will be rounded to the beginning of the bucket (if week -> first day of week).
It's important for week and month.

macros/edr/system/system_utils/buckets_cte.sql
```diff
@@ -1,75 +1,90 @@
 {% macro complete_buckets_cte(time_bucket) %}
-    {{ adapter.dispatch('complete_buckets_cte','elementary')(time_bucket) }}
+    {%- set edr_bucket_end_expr = elementary.edr_timeadd(time_bucket.period, time_bucket.count, 'edr_bucket_start') %}
+    {%- set edr_min_bucket_start_expr = elementary.edr_date_trunc(time_bucket.period, elementary.edr_cast_as_timestamp(elementary.edr_quote(elementary.get_min_bucket_start()))) %}
+    {%- set edr_max_bucket_end_expr = elementary.edr_cast_as_timestamp(elementary.edr_quote(elementary.get_max_bucket_end())) %}
```
Maayan-s, line 4: Moved the expressions out of the adapters to remove duplicated code. 
Changed the min bucket start to be truncated to the start of the bucket.

macros/edr/system/system_utils/buckets_cte.sql
```diff
@@ -1,75 +1,90 @@
...
+    {{ adapter.dispatch('complete_buckets_cte','elementary')(time_bucket, edr_bucket_end_expr, edr_min_bucket_start_expr, edr_max_bucket_end_expr) }}
 {% endmacro %}
 
-{% macro spark__complete_buckets_cte(time_bucket) %}
-    {% set edr_bucket_end_expr = elementary.edr_timeadd(time_bucket.period, time_bucket.count, 'edr_bucket_start') %}
+{% macro spark__complete_buckets_cte(time_bucket, edr_bucket_end_expr, edr_min_bucket_start_expr, edr_max_bucket_end_expr) %}
```
RoiTabach, line 8: really no need for the `edr_` prefix and the `_expr` suffix in the bucket_start, bucket_end, no?
Something like this would be cleaner?
```suggestion
{% macro spark__complete_buckets_cte(time_bucket, bucket_end, min_bucket_start, max_bucket_end) %}
```

integration_tests/models/schema.yml
```diff
@@ -40,6 +40,13 @@ models:
           config:
             severity: warn
           tags: ["table_anomalies"]
+      - elementary.volume_anomalies:
+          time_bucket:
+            period: week
+            count: 1
+          config:
+            severity: warn
+          tags: ["table_anomalies"]
```
RoiTabach, line 49: Changing in here requires changing in the python E2E test - but I didn't find another PR in the python.

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
RoiTabach: This means we could be taking a few days back extra in addition to what we previously would have taken. I think it's the better behaviour (better than rounding to the end) but maybe this can mean in some cases we take a lot more days - in case of `month` we can be adding upto 29 more days. 
I think we should add somewhere in the docs of how to use it, also how we round - so users expect this behaviour.

macros/edr/system/system_utils/buckets_cte.sql
Maayan-s: I kept the `edr_` as this was the convention before my changes. Can remove it probably.
`_exp` is important as these do not return a time, they return an SQL expression.

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
Maayan-s: If we don't round we will be comparing incomplete buckets to complete buckets (will cause false positives). 
I assume that you would configure month and week buckets expecting us to calculate for full months / weeks. The fact we don't do so is unexpected behavior.

macros/edr/data_monitoring/monitors_query/get_start_bucket_in_data.sql
```diff
@@ -1,6 +1,6 @@
 {% macro get_start_bucket_in_data(timestamp_column, min_bucket_start, time_bucket) %}
     {% set bucket_start_datediff_expr %}
-      floor({{ elementary.timediff(time_bucket.period, min_bucket_start, elementary.edr_cast_as_timestamp(timestamp_column)) }} / {{ time_bucket.count }}) * {{ time_bucket.count }}
+      floor({{ elementary.edr_datediff(min_bucket_start, elementary.edr_cast_as_timestamp(timestamp_column), time_bucket.period) }} / {{ time_bucket.count }}) * {{ time_bucket.count }}
```
RoiTabach, line 3: I have to say moving from `timediff` to `datediff` looked to me like it would truncate to dates but it doesn't, after checking in dbt-core

macros/edr/system/system_utils/buckets_cte.sql
```diff
@@ -1,75 +1,90 @@
...
-    {{ adapter.dispatch('complete_buckets_cte','elementary')(time_bucket) }}
+    {%- set bucket_end_expr = elementary.edr_timeadd(time_bucket.period, time_bucket.count, 'edr_bucket_start') %}
+    {%- set min_bucket_start_expr = elementary.edr_date_trunc(time_bucket.period, elementary.edr_cast_as_timestamp(elementary.edr_quote(elementary.get_min_bucket_start()))) %}
+    {%- set max_bucket_end_expr = elementary.edr_cast_as_timestamp(elementary.edr_quote(elementary.get_max_bucket_end())) %}
+    {{ adapter.dispatch('complete_buckets_cte','elementary')(time_bucket, bucket_end_expr, min_bucket_start_expr, max_bucket_end_expr) }}
 {% endmacro %}
 
-{% macro spark__complete_buckets_cte(time_bucket) %}
-    {% set edr_bucket_end_expr = elementary.edr_timeadd(time_bucket.period, time_bucket.count, 'edr_bucket_start') %}
+{% macro spark__complete_buckets_cte(time_bucket, bucket_end_expr, min_bucket_start_expr, max_bucket_end_expr) %}
```
RoiTabach, line 8: Cool, I didn't know we're allowed to have a different signature for the DWH__function functions in the adapter pattern

macros/utils/data_types/cast_column.sql
```diff
@@ -29,3 +29,16 @@
...
+
+{%- macro edr_cast_as_date(timestamp_field) -%}
+    {{ return(adapter.dispatch('edr_cast_as_date', 'elementary')(timestamp_field)) }}
+{%- endmacro -%}
+
+{%- macro default__edr_cast_as_date(timestamp_field) -%}
+    cast({{ timestamp_field }} as {{ elementary.edr_type_date() }})
+{%- endmacro -%}
+
+{# Bigquery (for some reason that is beyond me) can't cast a string as date if it's in timestamp format #}
```
RoiTabach, line 41: lol at the comment

macros/utils/cross_db_utils/timeadd.sql
```diff
@@ -9,7 +9,13 @@
 {% endmacro %}
 
 {% macro bigquery__edr_timeadd(date_part, number, timestamp_expression) %}
-    timestamp_add({{ elementary.edr_cast_as_timestamp(timestamp_expression) }}, INTERVAL {{ number }} {{ date_part }})
+    {%- if date_part | lower in ['second', 'minute', 'hour', 'day'] %}
+       timestamp_add({{ elementary.edr_cast_as_timestamp(timestamp_expression) }}, INTERVAL {{ number }} {{ date_part }})
+    {%- elif date_part | lower in ['week', 'month', 'quarter', 'year'] %}
+       date_add({{ elementary.edr_cast_as_date(timestamp_expression) }}, INTERVAL {{ number }} {{ date_part }})
```
RoiTabach, line 15: For BigQuery we cast as `date` and then add a week / month - would we lose the information on the hour-of-day ? Also - is this by intentional?

macros/utils/cross_db_utils/datediff.sql
```diff
@@ -1,7 +1,27 @@
-{% macro edr_datediff(first_date, second_date, datepart) %}
+{% macro edr_datediff(first_date, second_date, date_part) %}
```
RoiTabach, line 1: +1 for `date_part` instead of `datepart`

macros/utils/cross_db_utils/timeadd.sql
Maayan-s: 1. We have no choice, adding day/week/month to timestamp is not supported in BQ
2. We truncate days to 00:00:00 across the code anyway

integration_tests/models/schema.yml
Maayan-s: Consulted with Or and decided not to add specific tests here at the moment

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
RoiTabach: I think the end user possibly won't think this all the way through and would maybe be surprised. That's why I suggested adding it in the docs, not doing something else. 
I am thinking about some code running in prod each day with 30 days back, and on the 29th of the month we go back almost twice as much as we do on the 1st of the month. It _is_ a surprise IMO - but whatever you decide about it.

Maayan-s: It's just on the first execution, minor impact.

macros/edr/data_monitoring/monitors_query/get_start_bucket_in_data.sql
Maayan-s: Is this an actionable comment?

integration_tests/models/schema.yml
RoiTabach: What I meant here was - after adding this to the E2E we can expect that the report to look a bit different. We have a test in the E2E with very specific hard coded  expectations about what is going on in the package's E2E and it needs to be updated when the package's E2E changes. I ran this https://github.com/elementary-data/elementary/actions/runs/4524326381/jobs/7968070443 so I see there's no failures so no need to add anything.

RoiTabach (approved):

macros/utils/cross_db_utils/timeadd.sql
RoiTabach: Got it

macros/edr/data_monitoring/monitors_query/get_start_bucket_in_data.sql
RoiTabach: No - more like something for future readers wondering if it's truncating or not - it doesn't.

PR #299, opened 2023-03-22
support custom test name

Maayan-s:
Implementation based on @jtcohen6 suggestion in https://github.com/dbt-labs/dbt-core/issues/6470. 
Works for dbt generic tests and our tests on models and sources.

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_tests.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_tests.sql
@@ -188,7 +188,7 @@
     {% set generic_test_name = test_metadata.get('name') %} {# 'unique', 'relationships', 'volume_anomalies' etc #}
     {% set test_package_name = test_metadata.get('namespace') %}
     {% set test_instance_name = node_dict.get('name') %} {# Test custom name or dbt auto generated long name #}
-    {%- if test_instance_name and generic_test_name %}
+    {%- if generic_test_name %}
         {%- if test_package_name %}
             {% set test_short_name =
                 generic_test_name if (test_instance_name.startswith(test_package_name + '_' + generic_test_name) or test_instance_name.startswith(test_package_name + '_source_' + generic_test_name))
@@ -201,8 +201,6 @@
             %}
         {%- endif %}
         {{ return(test_short_name) }}
-    {%- elif generic_test_name %}
-        {{ return(generic_test_name) }}
     {%- else %}
         {{ return(test_instance_name) }}
     {%- endif %}
--- a/macros/edr/dbt_artifacts/upload_dbt_tests.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_tests.sql
@@ -47,7 +47,8 @@
 
     {% set test_metadata = elementary.safe_get_with_default(node_dict, 'test_metadata', {}) %}
     {% set test_namespace = test_metadata.get('namespace') %}
-    {% set test_short_name = test_metadata.get('name') %}
+    {% set test_short_name = elementary.get_test_short_name(node_dict, test_metadata) %}
+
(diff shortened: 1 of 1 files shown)


macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -174,6 +175,40 @@
...
+{% macro get_test_short_name(node_dict, test_metadata) %}
+    {#
+    If there is a custom name it overrides the dbt auto generated long name.
+    This is a best effort to extract custom names.
+    dbt automated name generation -
+        - Generic test long name starts with the generic name or source_generic
+        - Generic tests from packages long name starts with the package_generic or package_source_generic
+    #}
+
+    {% set generic_test_name = test_metadata.get('name') %} {# 'unique', 'relationships', 'volume_anomalies' etc #}
```
elongl, line 188: I think we should rename to `test_type_name`.

macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -174,6 +175,40 @@
...
+    {% set test_package_name = test_metadata.get('namespace') %}
+    {% set specific_test_name = node_dict.get('name') %} {# Test custom name or dbt auto generated long name #}
```
elongl, line 190: I think we should rename to `test_instance_name`.

macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -174,6 +175,40 @@
...
+    {%- if specific_test_name and generic_test_name and test_package_name %}
```
elongl, line 191: Only `test_package_name` can potentially be `None` here.
Maybe we should change the `if` only to it and move the non-package checks to the `elif` below that can be an `else`.

Roughly (pseudo):

```py
{% if test_package_name %}
  {{ return(generic_test_name if (specific_test_name.startswith(test_package_name + '_' + generic_test_name) or specific_test_name.startswith(test_package_name + '_source_' + generic_test_name))
              else specific_test_name) }}
{% else %}
  {{ return(generic_test_name if (specific_test_name.startswith(generic_test_name) or specific_test_name.startswith( 'source_' + generic_test_name))
              else specific_test_name) }}
{% endif %}
```

macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -174,6 +175,40 @@
...
+            %}
+        {%- else %}
+            {% set test_short_name =
+                generic_test_name if (test_instance_name.startswith(generic_test_name) or test_instance_name.startswith('source_' + generic_test_name))
+                else test_instance_name
+            %}
+        {%- endif %}
+        {{ return(test_short_name) }}
+    {%- elif generic_test_name %}
+        {{ return(generic_test_name) }}
```
elongl, line 205: I think this code will never get called.

macros/edr/dbt_artifacts/upload_dbt_tests.sql
```diff
@@ -174,6 +175,40 @@
...
+    This is a best effort to extract custom names.
+    dbt automated name generation -
+        - Generic test long name starts with the generic name or source_generic
+        - Generic tests from packages long name starts with the package_generic or package_source_generic
+    #}
+
+    {% set generic_test_name = test_metadata.get('name') %} {# 'unique', 'relationships', 'volume_anomalies' etc #}
+    {% set test_package_name = test_metadata.get('namespace') %}
+    {% set test_instance_name = node_dict.get('name') %} {# Test custom name or dbt auto generated long name #}
+    {%- if test_instance_name and generic_test_name %}
```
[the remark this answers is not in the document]
Maayan-s: Just to make the code clearer, but put it there before adding my comment so removing

PR #303, opened 2023-03-26
dump_table returns an empty list if table is missing.

elongl:


macros/utils/log_macro_results.sql
```diff
@@ -12,5 +12,7 @@
         {% do exceptions.raise_compiler_error("Received invalid macro name: {}".format(macro_name)) %}
     {% endif %}
     {% set results = macro(**macro_args) %}
-    {% do elementary.edr_log(tojson(results)) %}
+    {% if results is not none %}
+        {% do elementary.edr_log(tojson(results)) %}
```
elongl, line 16: For macros that still don't return and instead do `elementary.edr_log()` and return `None`,
this prevents the logging of the `None`.

haritamar (approved):

PR #304, opened 2023-03-27
Macro 'dump_table' supports excluding deprecatedly-marked columns.

elongl:


macros/commands/dump_table.sql
```diff
@@ -1,14 +1,24 @@
...
+            {% if deprecated_column_name in column_names %}
+                {% do column_names.remove(deprecated_column_name) %}
+            {% endif %}
+        {% endfor %}
+    {% endif %}
+
     {% set timestamp_column = node.meta.timestamp_column %}
     {% set query %}
-        select * from {{ node.relation_name }}
+        select {{ '\"' + column_names | join('\", \"') + '\"' }} from {{ relation }}
```
RoiTabach, line 21: we moved from querying `node.relation_name` to `adapter.get_relation` - why? I'm guessing that the new version is actually more stable just making sure

macros/commands/dump_table.sql
```diff
@@ -1,14 +1,24 @@
...
-    {% if not relation_exists %}
+    {% set relation = adapter.get_relation(database=node.database, schema=node.schema, identifier=node.alias) %}
+    {% if relation is none %}
         {% do print("Relation '{}' does not exist.".format(node.relation_name)) %}
         {% do return([]) %}
     {% endif %}
 
+    {% set column_names = adapter.get_columns_in_relation(relation) | map(attribute="name") | list %}
+    {% if exclude_deprecated_columns %}
+        {% set deprecated_column_names = node.columns.values()| selectattr("deprecated") | map(attribute="name") | list %}
```
RoiTabach, line 11: Did you test this on anything? What happens if `node.columns.values()` doesn't have an attribute `"deprecated"` (because it's empty?) 
Or alternatively - can we be sure we won't encounter this case?

elongl: Yes, tested.
We have only one column that does have it.

macros/commands/dump_table.sql
elongl: It doesn't really matter.
We already have the relation which is presumably more safe as we use it all other the place and is a more common property, but mostly doesn't matter.

RoiTabach: indeed more stable

macros/commands/dump_table.sql
```diff
@@ -1,14 +1,24 @@
...
+        {% for deprecated_column_name in deprecated_column_names %}
+            {% if deprecated_column_name in column_names %}
+                {% do column_names.remove(deprecated_column_name) %}
```
RoiTabach, line 14: instead of adding columns to the node and marking as deprecated, won't it make more sense to intersect the columns from the info schema with the columns from the node? And instead of deprecating - just deleting the column from the yaml?

RoiTabach: Guessing the answer is equivalent to the following question: Do we ever run this macro with `exclude_deprecated_columns`=False ?

elongl: I totally agree regarding the first point, the problem is that at the moment we don't maintain the columns list (some models don't even have it). It'd make sense to do it if we had a test that compares the table's state against the columns definition, but this will take some time and this is a blocker for a user.

Regarding the second question, no.

RoiTabach: Seems OK in this case to move forward

RoiTabach (approved):
Merge after running dump on the DWHs of our existing users, at least

PR #305, opened 2023-03-27
Raise an exception if user configures a timestamp column that doesn't exist

Maayan-s:
Today we just ignore and the test works without a timestamp.


elongl:
Don't we [already have this logic](https://github.com/elementary-data/dbt-data-reliability/blob/012dc165399b990f974127c9a863c8b5d51775b2/macros/edr/tests/test_table_anomalies.sql#L42)?

Maayan-s:
@elongl 
He saw the "column is not a timestamp" error and reached out in support as the column is a timestamp. 
But he had a typo in the column name. So the right error should be that there is no such column.

PR #307, opened 2023-03-29
Remove custom name from elementary tests - DONT MERGE

Maayan-s:


macros/edr/tests/test_utils/get_test_type.sql
```diff
@@ -0,0 +1,42 @@
+{% macro get_test_type(flattened_test) %}
+    {% if flattened_test.test_namespace == "elementary" %}
+        {%- set elementary_test_type = get_elementary_test_type(flattened_test) %}
+    {% endif %}
+
+    {%- if elementary_test_type and elementary_test_type != "python_test" %}
```
Maayan-s, line 6: I know python test is not dbt test, but we save this type to the test results. 
I wanted to maintain the current flow that saves these as dbt tests. 
WDYT?

PR #310, opened 2023-03-29
Ele 347 add new mode of day of week to our

RoiTabach:


macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -83,6 +83,11 @@
                 source_value,
                 bucket_start,
                 bucket_end,
+                {%- if metric_properties and metric_properties.seasonality and metric_properties.seasonality=='day_of_week' %}
+                    {{ elementary.edr_day_of_week_mod_7('bucket_end') }} as bucket_end_mod_7,
```
RoiTabach, line 87: Only open thing I'm wondering about - I have a column in `get_anomalies_scores_query` that has the meaning of "day_of_week(bucket_end) as an integer modulo 7" when we've got `seasonality: day_of_week` , but is just all `0` when we don't have seasonality. Wondering how to call it . Current name is `bucket_end_mod_7` .

Maayan-s: I think it's a pretty narrow name and implementation. 
If we want the flexibility to support other seasonality configs in the future, I suggest:

1. Naming the column `bucket_seasonality` 
I think it would also be much clearer to save there `Sunday` instead of 1, and null instead of 0. 

2. Set a var in the top like:
{%- set seasonality_expression = get_bucket_seasonality() %}
And do the logic of returning 0/null or an expression there (easier to expand later and also looks clearer, I think).

macros/utils/cross_db_utils/edr_day_of_week_mod_7.sql
```diff
@@ -0,0 +1,11 @@
+{% macro edr_day_of_week_mod_7(date_expr) %}
```
Maayan-s, line 1: I'm 90% sure that all the DBs have a syntax for extracting the day of week. 
Try this in Snowflake:
`SELECT DAYOFWEEK(cast('2022-08-02' as date))`
This will return a consistent number.

integration_tests/macros/e2e_tests/utils/table_utils.sql
```diff
@@ -0,0 +1,10 @@
+
```
Maayan-s, line 1: Rename file to table_assertions

integration_tests/models/schema.yml
```diff
@@ -301,6 +301,25 @@ models:
...
+    description: >
+      Volume anomalies on the training should fail when seasonality is Off and pass when seasonality is On. 
+      When Validation data is added, should fail when seasonality is On and pass when seasonality is Off.
+    tests:
+      - elementary.volume_anomalies:
+          alias: volume_anomalies_no_seasonality_anomalies
+          timestamp_column: "updated_at"
+          tags: ["seasonality_volume"]
+          sensitivity: 2
+          backfill_days: 14
```
Maayan-s, line 314: Why 14 backfill days? 
Also how would we have enough data points to calculate the anomaly? 
Won't we have only 2?

Maayan-s (changes requested):
Added some comments but a few bigger issues - 

1. This seasonality should only work with 'day' buckets. You don't validate that anywhere. I think the test should error if set with another bucket size. 

2. I don't understand how we will have enough data points to detect an anomaly. If I understand correctly the code, for 14 days back we would have 2 data points only.

integration_tests/models/schema.yml
RoiTabach: `backfill_days` means we predict on 14 days.
I'm running it with 210 `days_back` so we have 30 days back on each day of the week.

integration_tests/macros/e2e_tests/utils/table_utils.sql
RoiTabach: got it

RoiTabach: done

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
Maayan-s: @RoiTabach You didn't address this

macros/utils/cross_db_utils/edr_day_of_week_expression.sql
```diff
@@ -0,0 +1,34 @@
+{% macro edr_day_of_week_expression(date_expr) %}
```
Maayan-s, line 1: This works, but it's a mess and for no reason. 
You can cleanup all the `set` and `returns`, like in all the other croos_db_utils.

Maayan-s, line 1: Is the behavior consistent between all DWHs? 
Do they all return the exact same when running this on the same date?

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -83,6 +83,11 @@
                 source_value,
                 bucket_start,
                 bucket_end,
+                {%- if metric_properties and metric_properties.seasonality and metric_properties.seasonality=='day_of_week' %}
```
Maayan-s, line 86: Repeating my comment on this:

I think it's a pretty narrow name and implementation.
If we want the flexibility to support other seasonality configs in the future, I suggest:

1. Naming the column `bucket_seasonality`
I think it would also be much clearer to save there Sunday instead of 1, and null instead of 0.

2. Set a var in the top like:
`{%- set seasonality_expression = get_bucket_seasonality() %}`
And do the logic of returning 0/null or an expression there (easier to expand later and also looks clearer, I think).

macros/edr/tests/test_utils/validate_test_parameters.sql
```diff
@@ -0,0 +1,14 @@
...
+    {% if seasonality %}
+      {% if not timestamp_column %}
+        {% do exceptions.raise_compiler_error('Test with seasonality must have a timestamp_column, but none was provided') %}
+      {% endif %}
+      {% set supported_seasonality_values = ['day_of_week'] %}
+      {% if seasonality not in supported_seasonality_values %}
+        {% do exceptions.raise_compiler_error('Seasonality value should be one of' ~ supported_seasonality_values ~ ', got ' ~ seasonality ~ ' instead')%}
+      {% endif %}
+      {% if (time_bucket.count != 1) or (time_bucket.period != 'day') %}
+        {% do exceptions.raise_compiler_error('Seasonality is supported only with time_bucket 1 day, got ' ~ time_bucket ~ ' instead')%}
```
Maayan-s, line 11: time_bucket is a dict, this should be `time_bucket.period`

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
Maayan-s: Also now that you are not using modulo, for sure this is a bad column name

macros/edr/tests/test_utils/validate_test_parameters.sql
RoiTabach: I'm intentinoaly giving the whole time bucket in case it's a matter of `day` but not `1` (but I can print it prettier)

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
RoiTabach: > bucket_seasonality 
This name is so much better! 

> much clearer to save there Sunday instead of 1, and null instead of 0.
The 0 in case of no seasonality is just so we don't have to deal with differences in the DWH behaviors when partitioning by a None.
 
Do you think it can be a constant string without losing that better readability? "no_seasonality" or some form of that? 

About the days of the week vs the names of the days - I think day of week as number 0-6 is a more natural implementation than the long form of the name. This is a function that exists in all DWHs and versions, whereas getting the long name of the day of the week is something that's a bit less supported. For example in Spark dayofweek function exists, but to get the longform name this is the function call needed : `date_format(my_timestamp, 'EEEE')` (from this SO [thread](https://stackoverflow.com/a/33764798/20981468) )

Do you think this column would be more readable if it were to have the seasonality expression inside it, something like 
```
'{"seasonality":"day_of_week", "day_of_week":0}'
'{"seasonality":"day_of_week", "day_of_week":1}'
'{"seasonality":"day_of_week", "day_of_week":2}'
'{"seasonality":"day_of_week", "day_of_week":3}'
```
? 
And then the no seasonality case could look like
```
'{"seasonality":"none"}'
```

WDYT?

macros/utils/cross_db_utils/edr_day_of_week_expression.sql
RoiTabach: > Is the behavior consistent between all DWHs?
It's not - it's not even consistent in specific DWH version, since it's based on some account level definitions of the time the week starts etc. 
We only require for the day_of_week seasonality to have the same retval here for the same days of the week, which is the case as it is.

I think going for 100% consistency in the DWHs we support would be a lot of effort in dates handling internals and not sure what value it would bring.

RoiTabach: Cleaned up

macros/edr/tests/test_utils/validate_test_parameters.sql
RoiTabach: added

macros/utils/cross_db_utils/edr_day_of_week_expression.sql
RoiTabach: Using the name of the day will give a good 99.9% consistency

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
RoiTabach: using the name of the day for the bucket for cases where we have seasonality, and the string `no_seasonality` for cases where we don't have seasonality.

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -17,14 +17,20 @@
       {% set latest_full_refresh = none %}
     {% endif %}
 
+    {%- if metric_properties and metric_properties.seasonality and metric_properties.seasonality == 'day_of_week' %}
+        {% set bucket_seasonality_expr = elementary.edr_day_of_week_expression('bucket_end') %}
+    {% else %}
+        {% set bucket_seasonality_expr = elementary.const_as_string(elementary.edr_quote('no_seasonality')) %}
```
RoiTabach, line 23:
```suggestion
        {% set bucket_seasonality_expr = elementary.const_as_string('no_seasonality') %}
```

RoiTabach (commented):
.

macros/edr/tests/test_utils/validate_test_parameters.sql
```diff
@@ -0,0 +1,14 @@
+{% macro validate_seasonality_parameter(seasonality, time_bucket, timestamp_column) %}
```
Maayan-s, line 1: This is really nice, we should build these for more of our configs probably

macros/utils/cross_db_utils/day_of_week.sql
```diff
@@ -0,0 +1,34 @@
...
+{% macro bigquery__edr_day_of_week_expression(date_expr) %}
+    FORMAT_DATE('%A', {{ date_expr }})
+{% endmacro %}
+
+{% macro postgres__edr_day_of_week_expression(date_expr) %}
+    to_char({{ date_expr }}, 'Day')
+{% endmacro %}
+
+{% macro redshift__edr_day_of_week_expression(date_expr) %}
+{# RedShift returns the days padded with whitespaces to width of 9 #}
```
RoiTabach, line 19:
```suggestion
{# Redshift returns the days padded with whitespaces to width of 9 #}
```

RoiTabach (commented):
.

Maayan-s (approved):

PR #311, opened 2023-03-29
test-warehouse: uncomment integration tests

haritamar:


RoiTabach (approved):
can't believe we commented the tests out for a few weeks when this is the fix

PR #314, opened 2023-04-02
Add "direction" param to anomalies tests

noaKurman:
Added a flag called "anomaly_direction" implying if we want the anomaly test to only test upwards or downwards change,
Added the test to our e2e tests (all the relevant ones - volume, dimension, and column anomalies)

### DIFF
--- a/integration_tests/macros/e2e_tests/tests_validation.sql
+++ b/integration_tests/macros/e2e_tests/tests_validation.sql
@@ -119,8 +119,7 @@
             {% do dimensions_with_problems.append[dimensions] %}
         {% endif %}
     {% endfor %}
-
-    {% if results | length != 2 %}
+    {% if results | length != 3 %}
         {% do elementary.edr_log('FAILED: dimension anomalies tests failed because it has too many fail/error tests') %}
         {{ return(1) }}
     {% elif dimensions_with_problems %}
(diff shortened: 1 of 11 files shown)


integration_tests/macros/e2e_tests/tests_validation.sql
```diff
@@ -119,8 +119,7 @@
             {% do dimensions_with_problems.append[dimensions] %}
         {% endif %}
     {% endfor %}
-
-    {% if results | length != 2 %}
+    {% if results | length != 3 %}
```
noaKurman, line 122: I'm not sure why this is hardcoded in the first place, but I added another failed test so this is the new wanted value

integration_tests/run_e2e_tests.py
```diff
@@ -309,7 +329,7 @@ def e2e_tests(
 def print_failed_test_results(e2e_target: str, failed_test_results: List[TestResult]):
     print(f"Failed {e2e_target} tests:")
     for failed_test_result in failed_test_results:
-        print(f"{failed_test_result.type}: {failed_test_result.message}")
+        print(f"\033[1m\033[91m{failed_test_result.type}: {failed_test_result.message}\033[0m")
```
noaKurman, line 332: Wanted the end result to be bold and more noticeable, I can remove if it you think it's ugly

PR #315, opened 2023-04-02
Add "direction" param to anomalies tests

noaKurman:
Added a flag called "anomaly_direction" implying if we want the anomaly test to only test upwards or downwards change,
Added the test to our e2e tests (all the relevant ones - volume, dimension, and column anomalies)

### DIFF
--- a/integration_tests/macros/e2e_tests/tests_validation.sql
+++ b/integration_tests/macros/e2e_tests/tests_validation.sql
@@ -17,8 +17,8 @@
             {% do dimensions_with_problems.append[dimensions] %}
         {% endif %}
     {% endfor %}
-    {% if results | length != 2 %}
-        {% do elementary.edr_log('FAILED: dimension anomalies tests failed because it has too many fail/error tests') %}
+    {% if results | length != 3 %}
+        {% do elementary.edr_log('FAILED: dimension anomalies tests failed because it has too many fail/error tests: ' ~ results | length) %}
         {{ return(1) }}
     {% elif dimensions_with_problems %}
         {% do elementary.edr_log('FAILED: dimension anomalies tests failed on the dimensions - ' ~ dimensions_with_problems) %}
@@ -84,6 +84,42 @@
     {% endset %}
     {% set results = elementary.result_column_to_list(error_test_validation_query) %}
     {{ assert_lists_contain_same_items(results, ['error']) }}
+{% endmacro %}
+
+{% macro validate_spike_directional_anomalies() %}
+    {% set alerts_relation = ref('alerts_anomaly_detection') %}
+    {# Validating alert for correct direction anomalies #}
+
+    {% set row_count_validation_query %}
+        select distinct table_name
+        from {{ alerts_relation }}
+        where status in ('fail', 'warn') and tags like '%spike_directional_anomalies%';
+    {% endset %}
+    {% set results = elementary.result_column_to_list(row_count_validation_query) %}
+    -- The result list's purpose is a more readable error messages
+    {% set results_list = [] %}
+    {% for result in results %}
+        {% do results_list.append(result) %}
+    {% endfor %}
+    {{ assert_lists_contain_same_items(results_list,  ['any_type_column_anomalies']) }}
+{% endmacro %}
+
+{% macro validate_drop_directional_anomalies() %}
+    {% set alerts_relation = ref('alerts_anomaly_detection') %}
+    {# Validating alert for correct direction anomalies #}
+
+    {% set row_count_validation_query %}
+        select distinct table_name
+        from {{ alerts_relation }}
+        where status in ('fail', 'warn') and tags like '%drop_directional_anomalies%';
+    {% endset %}
+    {% set results = elementary.result_column_to_list(row_count_validation_query) %}
+    -- The result list's purpose is a more readable error messages
+    {% set results_list = [] %}
+    {% for result in results %}
+        {% do results_list.append(result) %}
+    {% endfor %}
+    {{ assert_lists_contain_same_items(results_list,  ['any_type_column_anomalies', 'dimension_anomalies', 'numeric_column_anomalies']) }}
 {% endmacro %}
 
 {% macro validate_error_model() %}
--- a/integration_tests/run_e2e_tests.py
+++ b/integration_tests/run_e2e_tests.py
@@ -152,6 +152,26 @@ def e2e_tests(
         ]
         test_results.extend(results)
 
+    if "directional_anomalies" in test_types:
+        dbt_runner.test(select="tag:drop_directional_anomalies")
+        results = [
+            TestResult(type="drop_directional_anomalies", message=msg)
+            for msg in dbt_runner.run_operation(
+                macro_name="validate_drop_directional_anomalies", should_log=False
+            )
+        ]
+        test_results.extend(results)
+
+    if "directional_anomalies" in test_types:
+        dbt_runner.test(select="tag:spike_directional_anomalies")
+        results = [
+            TestResult(type="spike_directional_anomalies", message=msg)
+            for msg in dbt_runner.run_operation(
+                macro_name="validate_spike_directional_anomalies", should_log=False
+            )
+        ]
+        test_results.extend(results)
+
     if "create_table" in test_types:
         # If there is a problem with create_or_replace macro, it will crash the test.
         dbt_runner.test(select="tag:table_anomalies")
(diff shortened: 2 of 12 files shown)


integration_tests/macros/e2e_tests/tests_validation.sql
```diff
@@ -119,8 +119,7 @@
             {% do dimensions_with_problems.append[dimensions] %}
         {% endif %}
     {% endfor %}
-
-    {% if results | length != 2 %}
+    {% if results | length != 3 %}
```
noaKurman, line 122: I'm not sure why this is hardcoded in the first place, but I added another failed test so this is the new wanted value

integration_tests/run_e2e_tests.py
```diff
@@ -309,7 +329,7 @@ def e2e_tests(
 def print_failed_test_results(e2e_target: str, failed_test_results: List[TestResult]):
     print(f"Failed {e2e_target} tests:")
     for failed_test_result in failed_test_results:
-        print(f"{failed_test_result.type}: {failed_test_result.message}")
+        print(f"\033[1m\033[91m{failed_test_result.type}: {failed_test_result.message}\033[0m")
```
noaKurman, line 332: Wanted the end result to be bold and more noticeable, I can remove if it you think it's ugly

integration_tests/macros/e2e_tests/tests_validation.sql
Maayan-s: Yes I hate that, I did this on the weekend to start moving in a better direction:
https://github.com/elementary-data/dbt-data-reliability/pull/313
(Validate that we only have errors in tests that have an `error_test` tag)

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -138,13 +137,19 @@
                 bucket_start,
                 bucket_end,
                 metric_value,
-                case 
+                case
                     when training_stddev is null then null
+                    {% if anomaly_direction is not none %}
```
Maayan-s, line 142: I think it's better to avoid these conditions in the query itself, as it is already hard to debug and understand it. 
Maybe set a default value of 'Both' to anomaly direction? 
It's also more explicit, is this is in fact the default behavior

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -49,6 +49,7 @@
     'elementary_full_refresh': false,
     'min_training_set_size': 14,
     'cache_artifacts': true,
+    'anomaly_direction': none,
```
Maayan-s, line 52: See my comment above about the default

macros/edr/tests/test_utils/get_anomaly_query.sql
```diff
@@ -51,12 +54,25 @@
...
+        {% else %}
+            abs(anomaly_score) > {{ sensitivity }}
+        {% endif %}
+    {% else %}
+        abs(anomaly_score) > {{ sensitivity }}
+    {% endif %}
+{% endmacro %}
 
+{%- macro is_score_anomalous_condition(sensitivity, anomaly_direction) -%}
+    {%- set spikes_only_metrics = ['freshness', 'event_freshness'] -%}
```
Maayan-s, line 72: Should this list be here or in the global vars? Is it possible that users would want to change it?

macros/edr/tests/test_utils/get_anomaly_query.sql
```diff
@@ -51,12 +54,25 @@
     {{- return(anomaly_query) -}}
 {% endmacro %}
 
-{%- macro is_score_anomalous_condition(sensitivity) -%}
-    {%- set spikes_only_metrics = ['freshness', 'event_freshness'] -%}
+{%- macro set_directional_anomaly(anomaly_direction, anomaly_score, sensitivity) -%}
```
Maayan-s, line 57: Just a thought as I see text comparison all over the code - 
Would this config work? Should it?
```
      - elementary.volume_anomalies:
          anomaly_direction: "Drop"
```

Maayan-s (changes requested):
Looks good!
Left some comments. 
Also we need to validate that the results look as intended in the UI.

macros/edr/tests/test_utils/get_anomaly_query.sql
noaKurman: Actually I didn't add this part (it's in the PR because I changed the function)
I asked about it as well, this was placed specifically here because it's a special case, when checking freshness there's no reason to see anything other than a "spike"
I've been explained that a drop would never be interesting, if you think otherwise I can edit that part too @Maayan-s

Maayan-s: Got it, makes sense

macros/edr/tests/test_utils/get_anomaly_query.sql
noaKurman: You're right, I changed it now to use "ilike" instead

Maayan-s: Maybe we should do something like what @RoiTabach did here:
https://github.com/elementary-data/dbt-data-reliability/pull/310/files/d304e37245216ecd1a781faf4346ff0e66871b6d#diff-3d90a1c0febaf4a23f00626f6686b1f1f2c17d1bd5f4f5f1e93b23f2ed0c91d6:~:text=macros/edr/tests/test_utils/validate_test_parameters.sql

What do you think?

integration_tests/run_e2e_tests.py
RoiTabach: It's really good imo 
Maybe adding a helper function with an indicative name would help make things readable?

macros/edr/tests/test_utils/get_anomaly_query.sql
noaKurman: Nice! 
Though I'm not sure having case-sensitivity checks is user friendly enough, for now I added a validation to only test if the parameters has the right string and I accept lower/higher cases, if you think it's redundant and we should just fail on upper case parameters let me know and I'll change it :)  
@Maayan-s

integration_tests/run_e2e_tests.py
noaKurman: Thanks, and I agree 
Though since there's a README and this task took longer than I expected anyway I think I'll postpone adding it to the next time I'll change the package 🙏

macros/edr/tests/test_utils/validate_test_parameters.sql
```diff
@@ -11,4 +11,15 @@
         {% do exceptions.raise_compiler_error('Seasonality is supported only with time_bucket 1 day, got period: ' ~ time_bucket.period ~ ' and count: ' ~ time_bucket.count ~ ' instead') %}
       {% endif %}
     {% endif %}
+{% endmacro %}
+
+{% macro validate_directional_parameter(anomaly_direction) %}
+    {% if anomaly_direction %}
+      {% set direction_case_insensitive = anomaly_direction | lower %}
+      {% if direction_case_insensitive != 'drop' and direction_case_insensitive != 'spike' and direction_case_insensitive != 'both' %}
```
Maayan-s, line 19: Not critical but for next time:
`direction_case_insensitive not in ['drop','spike',both']`

macros/edr/tests/test_utils/validate_test_parameters.sql
```diff
@@ -11,4 +11,15 @@
...
+        {% do exceptions.raise_compiler_error('Possible anomaly directions are: drop, spike') %}
```
Maayan-s, line 20: Errors need to be clear and actionable to the users, try to provide details.
Please change the error to:
```
{% do exceptions.raise_compiler_error("Supported anomaly directions are: both, drop, spike, recieved anomaly_direction: {}".format(anomaly_direction)) %}
```

macros/edr/tests/test_utils/validate_test_parameters.sql
```diff
@@ -11,4 +11,15 @@
...
+      {% endif %}
+    {% else %}
+      {% do exceptions.raise_compiler_error('anomaly_direction can\'t be empty') %}
```
Maayan-s, line 23:
```
{% do exceptions.raise_compiler_error('anomaly_direction can\'t be empty. Supported anomaly directions are: both, drop, spike') %}
```

macros/edr/tests/test_utils/get_anomaly_query.sql
```diff
@@ -36,11 +37,20 @@
       select
         metric_value as value,
         training_avg as average,
-        case when is_anomalous = TRUE then
+        -- when there's an anomaly we would want to use the last value of the metric (lag), otherwise visually the expactations would look out of bounds
```
Maayan-s, line 40: dbt uses `--` comments when parsing, so the BP here is to use the Jinja comments syntax `{# #}`

Maayan-s (approved):
LGTM

PR #318, opened 2023-04-06
Validate only test that should error return error

Maayan-s:


macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -242,7 +242,7 @@
     from bucket_freshness_ranked
     where row_number = 1
 {% else %}
-    {# Update freshness test not supported when timestamp column is not provided #}
+    {% do exceptions.raise_compiler_error("freshness_anomalies test is not supported whitout timestamp_column.") %}
```
Maayan-s, line 245: When I changed the e2e tests to fail if there is an unexpected error, I found that for `freshness_anomalies` without a timestamp column we pass on all platforms and error with a database error on Redshift. 
I think this is a better behavior.

PR #331, opened 2023-04-19
Add cached empty test results row instead of running the query a…

RoiTabach:
…ll the time


RoiTabach:
@roneilltm

RoiTabach:
Tested with Rory @roneilltm and seems this fix solves his issue (reducing a couple of thousand empty queries from running)

oravi (approved):

PR #335, opened 2023-04-23
Fixing anomaly tests on regular tables (non incremental) return all anomalies regardless of backfill days

Maayan-s:
https://elementary-community.slack.com/archives/C02CTC89LAX/p1682503992494309

### DIFF
--- a/integration_tests/macros/e2e_tests/tests_validation.sql
+++ b/integration_tests/macros/e2e_tests/tests_validation.sql
@@ -17,8 +17,11 @@
             {% do dimensions_with_problems.append[dimensions] %}
         {% endif %}
     {% endfor %}
-    {% if results | length != 3 %}
+    {% if results | length > 2 %}
         {% do elementary.edr_log('FAILED: dimension anomalies tests failed because it has too many fail/error tests: ' ~ results | length) %}
+        {{ return(1) }}
+    {% elif results | length < 2 %}
+        {% do elementary.edr_log('FAILED: dimension anomalies tests failed because it has not enough fail/error tests: ' ~ results | length) %}
         {{ return(1) }}
     {% elif dimensions_with_problems %}
         {% do elementary.edr_log('FAILED: dimension anomalies tests failed on the dimensions - ' ~ dimensions_with_problems) %}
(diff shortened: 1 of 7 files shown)


RoiTabach (approved):
good fix. Just to make sure: this syntax (`max () over (partition by)` ) is the same in all DWHs? (Equivalent question: Is this code flow covered in the integration tests?)

integration_tests/run_e2e_tests.py
```diff
@@ -387,7 +397,8 @@ def main(target, e2e_type, generate_data, clear_tests):
             "seasonal_volume",
             "table",
             "column",
-            "directional_anomalies"
```
Maayan-s, line 390: There was a missing comma here, so the tests for `directional_anomalies` (and `schema`) weren't running before.

integration_tests/macros/e2e_tests/tests_validation.sql
```diff
@@ -4,7 +4,7 @@
     {% set dimension_validation_query %}
         select *
         from {{ alerts_relation }}
-        where status in ('fail', 'warn') and sub_type = 'dimension'
+        where status in ('fail', 'warn') and tags like '%dimension_anomalies%'
```
Maayan-s, line 7: The tests for `directional_anomalies` (and `schema`) weren't running before.
When they did run the results were mixed with the dimension results, so I made changes to make sure each will be tested separately.

integration_tests/models/schema.yml
```diff
@@ -124,6 +124,25 @@ models:
...
               tags: ["error_test", "regular_tests"]
 
+  - name: backfill_days_column_anomalies
+    config:
+      elementary:
+        timestamp_column: updated_at
+    columns:
+      - name: "min_length"
+        tests:
+          - elementary.column_anomalies:
```
Maayan-s, line 134: Should pass because.the anomaly is 6 days ago and backfill_days is 2 by default.

integration_tests/models/schema.yml
```diff
@@ -124,6 +124,25 @@ models:
...
+              column_anomalies:
+                - min_length
+                - max_length
+              tags: ["backfill_days"]
+          - elementary.column_anomalies:
+              backfill_days: 7
```
Maayan-s, line 140: Should fail because the anomaly is 6 days ago and backfill_days here is 7

PR #338, opened 2023-04-29
[ELE-735] Add an exception when anomaly detection test is configured on ephemeral model

Maayan-s:
Happened twice on support this week, really hard to debug as the exception is not indicative.


macros/edr/system/system_utils/full_names.sql
```diff
@@ -62,10 +62,13 @@
 
 
 {% macro relation_to_full_name(relation) %}
-    {%- if relation.database %}
+    {%- if relation.identifier.startswith('__dbt__cte__') %}
```
haritamar, line 65: Maybe use `relation.is_cte`?

macros/utils/graph/is_ephemeral_model.sql
```diff
@@ -0,0 +1,5 @@
+{% macro is_ephemeral_model(model) %}
+    {% do return(
+      model.identifier.startswith('__dbt__cte__')
```
haritamar, line 3: Also here - I'm not sure, but is there a `model.type` we can use?

macros/edr/system/system_utils/full_names.sql
Maayan-s: @haritamar 
Nice one, didn't see it before! 
I did leave the `is_ephemeral` macro to be more explicit, but changed it to use the `is_cte` flag

PR #343, opened 2023-05-02
Ele 658 create view dbt columns

NoyaArie:
[ELE-658](https://linear.app/elementary/issue/ELE-658/create-a-view-table-with-users-columns-for-the-test-config-ui)

Add dbt_columns view.
The view includes data on all the columns of models and sources that are part of the user project.

Note:
For now, we don't support multi-DB for Redshift.
Postgres does not support cross-db queries.


macros/edr/metadata_collection/get_columns_by_configured_schemas.sql
```diff
@@ -0,0 +1,7 @@
+{% macro get_columns_by_configured_schemas(configured_schemas) %}
```
NoyaArie, line 1: extracted from `filtered_information_schema_columns.sql`

Maayan-s: Good call

macros/edr/metadata_collection/get_columns_from_information_schema.sql
```diff
@@ -15,13 +15,11 @@
         upper(column_name) as column_name,
         data_type
     from {{ schema_relation.information_schema('COLUMNS') }}
-    where upper(table_schema) = upper('{{ schema_name }}')
+    {%- if schema_name %} where upper(table_schema) = upper('{{ schema_name }}') {%- endif -%}
```
Maayan-s, line 18: This means that if you don't get a schema you select all schemas. Is that intended?

macros/edr/metadata_collection/get_columns_in_project.sql
```diff
@@ -0,0 +1,20 @@
+{% macro get_columns_in_project() %}
+    {% set configured_schemas = elementary.get_project_configured_schemas() %}
+    {{ elementary.get_columns_by_configured_schemas(configured_schemas) }}
+{% endmacro %}
+
+{% macro get_project_configured_schemas() %}
```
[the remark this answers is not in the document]
Maayan-s: Is it still used? Can it be replaced by your `macro`?

Maayan-s (commented):
Added small comments. 
I'm not sure why but I see here all of the commits of other branches that were merged after you created the branch. Could you consult with the team on how they rebase so it won't repeat? 
This time most of the commits were mine, so I managed :)

macros/edr/metadata_collection/get_columns_from_information_schema.sql
NoyaArie: yep
but I will remove the condition since I found out yesterday that Bigquery requires a schema name (or else, an error)

macros/edr/metadata_collection/get_columns_in_project.sql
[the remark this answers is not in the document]
NoyaArie: It is still in use, the schema changes tests use it. 
can't be replaced, has a different logic - tasks from the graph resource type `test`, where name start with `schema_changes` and package not `elementary`

models/edr/dbt_artifacts/dbt_columns.sql
```diff
@@ -0,0 +1,45 @@
...
+        database_name,
+        schema_name,
+        name as table_name
+    from {{ ref('dbt_sources') }}
+),
+
+tables_information as (
+    select * from dbt_models_data
+    union all
+    select * from dbt_sources_data
```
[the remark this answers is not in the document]
NoyaArie: good question, don't know. do you think we need to show also columns of snapshot tables?

macros/edr/metadata_collection/get_columns_in_project.sql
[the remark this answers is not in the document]
Maayan-s: Yes but the point there was to bring only the columns we must and not all project columns.
Now we bring all the columns anyway, so that logic is outdated.

models/edr/dbt_artifacts/dbt_columns.sql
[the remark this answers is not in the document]
Maayan-s: 100%, snapshots are models just with a different materialization. No idea why dbt threats thrm differently, probably just legacy.

macros/edr/system/configuration/get_configured_schemas_from_graph.sql
```diff
@@ -1,21 +1,14 @@
...
-                        {% if node_package_name != 'elementary' %}
-                            {% if adapter.check_schema_exists(node['database'], node['schema']) %}
-                                {% do configured_schemas.append((node['database'], node['schema'])) %}
-                            {% endif %}
-                        {% endif %}
-                    {% endfor %}
+        {% set root_project = context["project_name"] %}
+        {% set nodes = elementary.get_nodes_from_graph() %}
+        {% for node in nodes %}
+            {% if node.resource_type in ['model', 'source', 'snapshot'] and node.package_name == root_project %}
```
haritamartest, line 7: We should include `seed` as well in the list of resource types

models/edr/dbt_artifacts/dbt_columns.sql
```diff
@@ -0,0 +1,55 @@
...
+    from {{ ref('dbt_snapshots') }}
+),
+
+dbt_sources_data as (
+    select
+        database_name,
+        schema_name,
+        name as table_name
+    from {{ ref('dbt_sources') }}
+),
```
haritamartest, line 30: Also here, let's query from `dbt_seeds` too

macros/edr/metadata_collection/get_columns_by_configured_schemas.sql
haritamartest, line 1: Maybe it should be just `get_columns_by_schemas`?
(I think this function is not aware that they're "configured")

macros/edr/system/configuration/get_configured_schemas_from_graph.sql
NoyaArie: let's talk about it :) 
just because each time another one comes up, and want to make sure we won't miss something

haritamar (approved):

PR #346, opened 2023-05-10
Ele 700 add default values to time bucket 

Maayan-s:
<details>
<summary><a href="https://linear.app/elementary/issue/ELE-700/add-default-values-to-time-bucket-if-only-count-or-period-are-passed">ELE-700 Add default values to `time_bucket` if only count or period are passed</a></summary>
<p>

Currently, the default is `count=1, period=day`.
If the user only changes one of these, for example `period=hour` the test fails because `count` is missing.
</p>
</details>

<details>
<summary><a href="https://linear.app/elementary/issue/PRO-76/support-defining-anomaly-detection-time-buckets-at-the-model-level">ELE-162 Support defining anomaly detection time buckets at the model-level</a></summary>
<p>

Currently they're only passed as a test's argument.
This requires duplication in case multiple Elementary tests are defined.
</p>
</details>


integration_tests/dbt-macro-debugger
elongl, line 1: Why add this to here?

integration_tests/macros/e2e_tests/test_config_levels.sql
```diff
@@ -0,0 +1,20 @@
+{% test config_levels(model, expected_config, time_bucket) %}
+    {%- if execute and flags.WHICH in ['test', 'build'] %}
+        {%- set unexpected_config = [] %}
+        {%- set model_relation = dbt.load_relation(model) %}
+        {% set model_graph_node = elementary.get_model_graph_node(model_relation) %}
+
+        {%- set time_bucket = elementary.get_time_bucket(time_bucket, model_graph_node) %}
+        {%- if time_bucket != expected_config.get('time_bucket') %}
+            {%- set unexpected_message = ('got config: {0}, expected config: {1}').format(time_bucket, expected_config.get('time_bucket') ) %}
```
elongl, line 9: Why is `unexpected_config` instead of raising an error here?

integration_tests/macros/unit_tests/test_get_time_bucket.sql
elongl, line 1: Shall we delete this file?

integration_tests/models/config_levels_project.sql
elongl, line 1: Why not re-use the `one` model?

macros/edr/tests/test_configuration/get_time_bucket.sql
```diff
@@ -0,0 +1,67 @@
...
+{% endmacro %}
+
+{% macro get_time_bucket(time_bucket, model_graph_node) %}
+    {%- set configured_time_bucket = elementary.get_test_argument('time_bucket', time_bucket, model_graph_node) %}
+    {%- do elementary.validate_time_bucket(configured_time_bucket) %}
+    {%- set default_time_bucket = elementary.get_default_time_bucket() %}
+
+    {%- if not configured_time_bucket %}
+        {{ return(default_time_bucket) }}
+    {%- elif configured_time_bucket and not configured_time_bucket.period %}
```
elongl, line 16: There's no need to re-check if `configured_time_bucket` because you test that above.
Also, you can do this if you want:
```j2
{% set time_bucket = default_time_bucket.copy() %}
{% do time_bucket.update(configured_time_bucket) %}
```

macros/edr/tests/test_configuration/get_time_bucket.sql
```diff
@@ -0,0 +1,67 @@
...
+
+               time_bucket:
+                 count: int
+                 period: string
+            ") %}
+        {%- endif %}
+        {%- if time_bucket is mapping %}
+            {%- set invalid_keys = [] %}
+            {%- set valid_keys = ['period', 'count'] %}
+            {%- for key, value in time_bucket.items() %}
```
elongl, line 40: You should probably use `valid_keys` or preferable use the keys of the default time bucket.

macros/edr/tests/test_configuration/get_time_bucket.sql
```diff
@@ -0,0 +1,67 @@
...
+
+                   time_bucket:
+                     count: int
+                     period: string
+                ").format(invalid_keys, valid_keys)) %}
+            {%- endif %}
+        {%- endif %}
+
+        {% if time_bucket.count and time_bucket.count is not integer %}
+            {% do exceptions.raise_compiler_error("time_bucket.count expectes valid integer, got: {} (If it's an integer, try to remove quotes)".format(time_bucket.count)) %}
```
elongl, line 60: expects*

macros/edr/tests/test_configuration/get_time_bucket.sql
```diff
@@ -0,0 +1,67 @@
...
+        {% endif %}
+        {% set supported_periods = ['hour','day','week','month'] %}
```
elongl, line 62: Why are we making this validation? Worst case the warehouse would error.
For instance, `minute` might be supported and we're not allowing it here.

integration_tests/macros/e2e_tests/test_config_levels.sql
Maayan-s: Helps to debug, you can see clearly what was unexpected on the error message (if more than one config you will see all).

integration_tests/macros/unit_tests/test_get_time_bucket.sql
Maayan-s: No, we should run the unit tests in CI

integration_tests/models/config_levels_project.sql
Maayan-s: I wanted to create a dedicated section on the schema YML where I control all the config. Didn't want to mix with other tests. 
I think this should be our practice... Noa added the spike and drop tests to existing tests and it really messed up the e2e (I split that already).

PR #347, opened 2023-05-11
delete_and_insert macro deletes NULLs.

elongl:
Doing:
```sql
delete from <TABLE> where <COLUMN> in (null);
```
doesn't delete rows where the column's value is `NULL`.
Instead, you have to explicitly match the case for `NULL` like so:
```sql
delete from <TABLE> where <COLUMN> is null;
```

The problem that happened is that if a user was running dbt<1.4.0 in which the caching is not supported,
the `metadata_hash` field was `null` because there was no way to calculate the hash (`local_md5` introduced in dbt v1.4.0).

If the user later upgraded to dbt >=1.4.0, the previous artifacts that had `NULL`s in them were kept and weren't deleted.
This resulted in duplicated rows in the table and therefore the report.
This PR solves this issue.

As a workaround, doing `dbt run -s elementary` should resolve the issue as it ignores the caching and rebuilds the tables.

### DIFF
--- a/macros/utils/table_operations/delete_and_insert.sql
+++ b/macros/utils/table_operations/delete_and_insert.sql
@@ -32,7 +32,10 @@
     {% set query %}
         begin transaction;
         {% if delete_relation %}
-            delete from {{ relation }} where {{ delete_column_key }} in (select {{ delete_column_key }} from {{ delete_relation }});
+            delete from {{ relation }}
+            where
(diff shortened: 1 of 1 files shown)


IDoneShaveIt (approved):

Maayan-s:
https://elementary-community.slack.com/archives/C02CTC89LAX/p1683731828513099

PR #348, opened 2023-05-11
Not rely on adapter's make_temp_relation's implementation.

elongl:


macros/utils/table_operations/make_temp_relation.sql
```diff
@@ -0,0 +1,9 @@
+{% macro make_temp_relation(base_relation, suffix=none) %}
```
IDoneShaveIt, line 1: Is that fix relevant to all of the adapters?
Maybe we want to create a default one that uses the dbt one and one for databricks?

IDoneShaveIt (approved):
Looks good.
Was wandering if we prefer to use by default the dbt version and for databricks use a specific implementation 🤔

PR #349, opened 2023-05-11
Added insertion time to a table.

elongl:


macros/utils/table_operations/insert_rows.sql
```diff
@@ -52,8 +52,13 @@
     {% for row in rows %}
       {% set rendered_column_values = [] %}
       {% for column in columns %}
-        {% set column_value = elementary.insensitive_get_dict_value(row, column.name) %}
-        {% do rendered_column_values.append(elementary.render_value(column_value)) %}
+        {% if column.name == "created_at" %}
```
elongl, line 55: Maybe there should be a const, or given that it's dbt, a function (macro) that returns this const.

models/edr/data_monitoring/schema_changes/schema_columns_snapshot.sql
elongl, line 1: This model is populated with `MERGE` rather than `insert_rows`.

models/edr/data_monitoring/data_monitoring/data_monitoring_metrics.sql
elongl, line 1: This model is populated with `MERGE` rather than `insert_rows`.

macros/utils/table_operations/insert_rows.sql
elongl: Also could be that we need to provide this as value in the actual upload of the data rather than magically here.
The challenge is that we'd need to be able to tell it to be unquoted.

macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -53,53 +53,103 @@
...
+    {% endif %}
+
+    {%- set target_relation = adapter.get_relation(database=database_name, schema=schema_name, identifier='data_monitoring_metrics') -%}
+    {% if not target_relation %}
+      {% do exceptions.raise_compiler_error("Couldn't find Elementary's models. Please run `dbt run -s elementary`.") %}
+    {% endif %}
+
+    {%- set temp_relation = elementary.make_temp_view_relation(target_relation) -%}
+    {% set insert_query %}
+      INSERT INTO {{ target_relation }} (
```
elongl, line 72: This means that there could be multiple rows with the same `id`.
It probably means that we need to add an additional value to the [`id` construction](https://github.com/elementary-data/dbt-data-reliability/blob/539119bebb21ca9a035a9436f467f29b07c6f44b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql#L74-L75) such as the invocation ID.

haritamar (approved):

PR #350, opened 2023-05-13
Pro 134 allowing test configuration on different levels (project/model/test)

Maayan-s:
The goal of this PR is to allow robust configuration of anomaly detection tests.
Users could configure all of the test params in the project / model / test level, and Elementary would always prioritize the more granular config. 

Motivation:
- This will allow users to tailor the tests to the dataset and get higher level of accuracy. 
- Efficiency, users can configure at a higher level (folder of models) and save the need to configure by test. 
- It will make it easier for us to create new tests and reduce code duplication. 

Changes: 
- A macro named `get_anomalies_test_configuration` is called by each test. For each param, this macro searches for configuration at the test, then model then project level. 
- This macro also runs validations of the configs and raises indicative exceptions. 
- The macro returns `test_configuration` and `metric_propeties` dicts. It also writes the configuration to the cache, so we could access it from outside the test context (in the test materialization).


integration_tests/integration_tests/test_anomaly_scores_query.py
```diff
@@ -168,24 +168,31 @@ def test_anomaly_scores_query(
         "resource_type": "dummy_type_not_incremental",
         "alias": alias_table_name,
         "database": database,
-        "schema": schema,
+        "schema": schema
+    }
+
+    test_configuration = {
```
Maayan-s, line 174: Made integration test work with the `test_configuration` dict.

integration_tests/macros/e2e_tests/test_config_levels.sql
```diff
@@ -0,0 +1,53 @@
+{% test config_levels(model, expected_config, timestamp_column, time_bucket, where_expression, anomaly_sensitivity, anomaly_direction, days_back, backfill_days, seasonality, min_training_set_size) %}
```
Maayan-s, line 1: The test is executed 3 times:
- 2 tests on the same model, one with test config and the other with model config. The validation is that the model level configs override the project level, and the test level configs override the model level config. 
- 1 test on a model with no specific config, to see that the project defaults are picked up.

macros/edr/system/system_utils/get_test_argument.sql
```diff
@@ -1,7 +1,18 @@
-{% macro get_test_argument(argument_name, value=none) %}
+{% macro get_test_argument(argument_name, value, model_graph_node) %}
```
Maayan-s, line 1: Changed this to check at the model level as well, before checking vars.

integration_tests/models/schema.yml
```diff
@@ -397,34 +397,73 @@ models:
     tags: ["config_levels"]
```
Maayan-s, line 397: TODO: 
Add the same test for sources as well.

PR #355, opened 2023-05-16
add materialization to run results table

arilmav:
Adds the materialization field to the dbt_run_results table, addressing ELE-847.

Screenshot 2023-05-16 at 4 04 38 PM

### DIFF
--- a/macros/edr/dbt_artifacts/upload_run_results.sql
+++ b/macros/edr/dbt_artifacts/upload_run_results.sql
@@ -28,7 +28,8 @@
                                                                        ('compiled_code', 'long_string'),
                                                                        ('failures', 'bigint'),
                                                                        ('query_id', 'string'),
-                                                                       ('thread_id', 'string')
+                                                                       ('thread_id', 'string'),
+                                                                       ('materialization', 'string')
                                                                        ]) %}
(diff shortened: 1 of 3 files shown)


Maayan-s:
Thank you @arilmav! 
Code works as expected 👌🏻

I think that it would also be good to change the views `model_run_results` and `snapshot_run_results` to select the materialization from the results instead of the metadata model. 
This will make the view more accurate, as it will handle the case of a change in materialization (will show the relevant materialization for each run and not the current one). 
To maintain backward compatibility and avoid NULLs, I suggest `coalesce(run_results.materialization, models.materialization)`. 

Want to add that to the PR as well?

arilmav:
> Thank you @arilmav! Code works as expected 👌🏻
> 
> I think that it would also be good to change the views `model_run_results` and `snapshot_run_results` to select the materialization from the results instead of the metadata model. This will make the view more accurate, as it will handle the case of a change in materialization (will show the relevant materialization for each run and not the current one). To maintain backward compatibility and avoid NULLs, I suggest `coalesce(run_results.materialization, models.materialization)`.
> 
> Want to add that to the PR as well?

Done!

PR #363, opened 2023-05-28
Ele 826 orchestration integration

NoyaArie:
Include job URL for each invocation (calculate in case of GitHub actions)


Maayan-s:
@NoyaArie 
I think we should implement getting all the job and orchestrator related metadata from either var or env_var. 
Check for var, else check for env_var, else null. 
This way we could mitigate the Airflow + dbt cloud scenario, and also provide an easy interface for cases where env var is not that easy to use.

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -25,18 +25,19 @@
...
       'env_id': elementary.get_first_env_var(["DBT_ENV_ID"]),
       'project_id': elementary.get_first_env_var(["DBT_PROJECT_ID", "DBT_CLOUD_PROJECT_ID", "GITHUB_REPOSITORY"]),
       'cause_category': elementary.get_first_env_var(["DBT_CAUSE_CATEGORY", "DBT_CLOUD_RUN_REASON_CATEGORY", "GITHUB_EVENT_NAME"]),
       'cause': elementary.get_first_env_var(["DBT_CAUSE", "DBT_CLOUD_RUN_REASON"]),
       'pull_request_id': elementary.get_first_env_var(["DBT_PULL_REQUEST_ID", "DBT_CLOUD_PR_ID", "GITHUB_HEAD_REF"]),
       'git_sha': elementary.get_first_env_var(["DBT_GIT_SHA", "DBT_CLOUD_GIT_SHA", "GITHUB_SHA"]),
-      'orchestrator': elementary.get_orchestrator(),
+      'orchestrator': elementary.get_config_var("orchestrator") or elementary.get_orchestrator(),
       'dbt_user': elementary.get_first_env_var(["DBT_USER"]),
+      'job_url': elementary.get_job_url(),
```
Maayan-s, line 40: @NoyaArie
You think there isn't a use case to pass the job url as a var?

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -110,7 +112,19 @@
       {% do return(orchestrator) %}
     {% endif %}
   {% endfor %}
-  {% do return(elementary.get_first_env_var(["DBT_ORCHESTRATOR"])) %}
+  {% do return(elementary.get_first_env_var(["ORCHESTRATOR", ""DBT_ORCHESTRATOR"])) %}
+{% endmacro %}
+
+{% macro get_job_url() %}
+  {% if elementary.get_first_env_var(["GITHUB_ACTIONS"]) %}
+    {% set server_utl = elementary.get_first_env_var(["GITHUB_SERVER_URL"]) %}
```
haritamar, line 120: Fix typo (utl -> url)

macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -110,7 +112,19 @@
...
+    {% set repository = elementary.get_first_env_var(["GITHUB_REPOSITORY"]) %}
+    {% set run_id = elementary.get_first_env_var(["GITHUB_RUN_ID"]) %}
+
+    {% set github_job_url = server_utl ~ "/" ~ repository ~ "/actions/runs/" ~ run_id %}
+    {% do return(github_job_url) %}
+  {% endif %}
+  {% do return(elementary.get_var("job_url", ["JOB_URL", "DBT_JOB_URL"])) %}
```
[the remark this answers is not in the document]
haritamar: Also a thought (but not 100% sure about it) - maybe we want to save the just server URL, but calculate the full job URL elsewhere? (even directly in the SaaS / CLI)
May be a bit more robust to the case the exact format of the full URL will change over time in the orchestrator side.

haritamar (approved):

Maayan-s (approved):

PR #367, opened 2023-06-04
Ele 983 generate job url and job run url

NoyaArie:
Support job_url and job_run_url fields.
Generate values for:
- Airflow
  - job_url - [domain]/dags/[job_id]/grid
  - job_run_url - [domain]/dags/[job_id]/grid?dag_run_id=[job_run_id]
- dbt Cloud
  - job_url - https://cloud.getdbt.com/deploy/[account_id]/projects/[DBT_CLOUD_PROJECT_ID]/jobs/[DBT_CLOUD_JOB_ID]
  - job_run_url - https://cloud.getdbt.com/deploy/[account_id]/projects/[DBT_CLOUD_PROJECT_ID]/runs/[DBT_CLOUD_RUN_ID]
- Github Actions
  - same value for job_url and job_run_url
  - [GITHUB_SERVER_URL]/[GITHUB_REPOSITORY]/actions/runs/[GITHUB_RUN_ID]
  
  
### I run:
Screen Shot 2023-06-13 at 9 17 15

### The results in dbt_invocations table:
Screen Shot 2023-06-13 at 9 20 07

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_invocation.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_invocation.sql
@@ -6,6 +6,9 @@
 
   {% do elementary.file_log("Uploading dbt invocation.") %}
   {% set now_str = elementary.datetime_now_utc_as_string() %}
+  {% set orchestrator = elementary.get_orchestrator() %}
+  {% set job_id = elementary.get_var("job_id", ["JOB_ID", "DBT_JOB_ID", "DBT_CLOUD_JOB_ID"]) %}
+  {% set job_run_id = elementary.get_var("job_run_id", ["DBT_JOB_RUN_ID", "DBT_CLOUD_RUN_ID", "GITHUB_RUN_ID"]) %}
   {% set dbt_invocation = {
       'invocation_id': invocation_id,
       'run_started_at': elementary.run_started_at_as_string(),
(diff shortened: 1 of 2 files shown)


macros/edr/dbt_artifacts/upload_dbt_invocation.sql
```diff
@@ -120,12 +123,50 @@
...
 
-{% macro get_job_url() %}
+{% macro get_job_url(orchestrator, job_id) %}
   {% set var_value = elementary.get_var("job_url", ["JOB_URL", "DBT_JOB_URL"]) %}
   {% if var_value %}
     {% do return(var_value) %}
   {% endif %}
-  {% if elementary.get_first_env_var(["GITHUB_ACTIONS"]) %}
+  {% if orchestrator == 'airflow' %}
+    {% set server_url = elementary.get_var('orchestrator_domain', ["ORCHESTRATOR_DOMAIN"]) %}
```
elongl, line 132: I think this should be called `AIRFLOW_URL` rather than `ORCHESTRATOR_DOMAIN` because:
1. It's not only the domain, it's the URL, for instance, it needs to include `http://` or `https://`, port, and so on.
2. It only applies to Airflow so we'd want to couple it with the word Airflow.

elongl (commented):
Left one small comments, looks awesome.
Have we tested that with the various orchestrators?

elongl:
The `GITHUB_SERVER_URL` would be `https://github.com` and the `GITHUB_REPOSITORY` would be `elementary-data/elementary`, but that would work regardless.

elongl (approved):

PR #377, opened 2023-06-12
fallback for load relation

Maayan-s:


elongl:
Why won't we drop this?
I think the value is limited because any query on the tested table would show an error message saying it doesn't exist.

Maayan-s:
@elongl 
All the uses we do with `model` work. I tested. 
I assume we had edge cases with quoting that failed so we used the `load_relation`, but failing to use `load_raltion` doesn't mean the tested table can't be accessed. 
The user that reported saw this error `Unable to find table `HIVE_METASTORE.NZ_PROD__STAGE_30.GPEVENTS_30__DIM_ORDER_CREATED``, the table name here is from the relation that failed to load.

elongl (approved):

PR #379, opened 2023-06-14
Ele 966 add exposure data

NoyaArie:


macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -12,6 +12,7 @@
 {% macro get_dbt_exposures_empty_table_query() %}
     {% set dbt_exposures_empty_table_query = elementary.empty_table([('unique_id', 'string'),
                                                                      ('name', 'string'),
+                                                                     ('label', 'string'),
```
Maayan-s, line 15: I think that we had an issue with Redshift, and you need to add the column as last on the table for it not to fail when users upgrade from a version that doesn't have it to this one.

elongl (approved):

PR #380, opened 2023-06-18
default for macro `complete_buckets_cte`

Maayan-s:
image

https://github.com/elementary-data/elementary/issues/932


elongl:
I think a better solution would be to do:

```yaml
models:
  elementary:
    +enabled: "{{ adapter.type != '...' }}"
```

Maayan-s:
This macro is not used in models and hooks, it's used in the anomaly detection tests that can't be disabled per adapter. 
The issue here is that if there is no default it fails even in compile for unsupported adapters, not on execution.

elongl (approved):

PR #383, opened 2023-06-20
Test all adapter-specific macros have a default implementation.

elongl:


integration_tests/run_e2e_tests.py
```diff
@@ -360,13 +361,25 @@ def e2e_tests(
         ]
         test_results.extend(results)
 
+    if "adapter_macro_has_default" in test_types:
```
IDoneShaveIt, line 364: Maybe we should split the tests into 2 parts?
The first part will be to run tests like that so we could fail fast, and the second one will be to test the e2e flow.

when thinking about it out-loud this is a unit test so it does make sense to fail fast on it 🙂

IDoneShaveIt (approved):
LGTM

integration_tests/run_e2e_tests.py
IDoneShaveIt: So it is still part of the e2e test?

integration_tests/run_e2e_tests.py
```diff
@@ -419,6 +432,7 @@ def main(target, e2e_type, generate_data, clear_tests):
             "error_snapshot",
             "dimension",
             "create_table",
+            "adapter_macro_has_default",
```
IDoneShaveIt, line 435: We don't want it here anymore do we :) ?

IDoneShaveIt (approved):
LGTM

PR #396, opened 2023-06-29
ELE-1206 bug when upgrading elmentary package after metrics tables schema is changed

IDoneShaveIt:


macros/edr/tests/test_utils/create_elementary_test_table.sql
```diff
@@ -3,6 +3,15 @@
         {% set temp_table_name = elementary.table_name_with_suffix(test_name, "__" ~ table_type) %}
         {{ elementary.debug_log(table_type ~ ' table: ' ~ database_name ~ '.' ~ schema_name ~ '.' ~ temp_table_name) }}
 
+        {#
```
elongl, line 6: Why is this needed if we use `CREATE OR REPLACE` to create the test tables?

IDoneShaveIt:
There was no bug, It was an orchestration problem when upgrading elementary

PR #397, opened 2023-07-02
add dummy materialization macro

ellakz:


macros/materializations/dummy.sql
```diff
@@ -0,0 +1,10 @@
+{# Materialization that does not create any table at the end of its run #}
+{# An example for a usage case is when we want a module to appear on the Elementary lineage graph, but the table is created outside of dbt #}
```
elongl, line 2:
```suggestion
{# An example for a usage case is when we want a model to appear on the Elementary lineage graph, but the table is created outside of dbt #}
```

macros/materializations/dummy.sql
```diff
@@ -0,0 +1,10 @@
...
+{% materialization dummy, default -%}
+  {# The main statement executes the model, but does not create any table / view on the DWH #}
```
elongl, line 4: Why do we execute anything at all in the dummy materialization?

ellakz: @elongl  this part executes the sql from the model definition, which is supposed to be empty except for the config and the dependencies, but we do need those

integration_tests/models/dummy_model.sql
```diff
@@ -0,0 +1,4 @@
+{{ config(materialized='view') }}
+    begin;
+    commit;
```
haritamar, line 3: I think the syntax for transactions is slightly different between DBs + Databricks actually don't support them at all (interesting to see what happens in the CI once it passes)

In any case may be safer to do something like `SELECT 1` that will always work

macros/materializations/dummy.sql
```diff
@@ -0,0 +1,10 @@
+{# Materialization that does not create any table at the end of its run #}
+{# An example for a usage case is when we want a model to appear on the Elementary lineage graph, but the table is created outside of dbt #}
+{% materialization dummy, default -%}
```
haritamar, line 3: I'm not 100% sure about the name from a product perspective - since we're going to generate a bunch of these for users maybe it should say something like "external" or "non_dbt"?
Worth consulting with @oravi , the king of names :)

Though I agree that functionality-wise "dummy" makes sense.

integration_tests/macros/e2e_tests/tests_validation.sql
```diff
@@ -197,3 +197,7 @@
     {% set result = elementary.run_query(query) %}
     {% do assert_empty_table(result, "dbt_invocations") %}
 {% endmacro %}
+
+{% macro assert_no_dummy_model_table() %}
+    {{ load_relation(ref('dummy_model')) is none }}
```
haritamar, line 202: A couple of comments here:

1. I think it will be better to do `{% do return(load_relation(ref('dummy_model')) is none) %}` so the return value will be a boolean rather than a string containing "true".
2. This macro is a bit inconsistent with the other "assert" macros in the sense that it returns a value rather than print "SUCCESS" / "FAILURE" messages, so I think either:
- Remove the `assert_` prefix
- Change this method to print the messages directly

macros/materializations/dummy.sql
ellakz: @haritamar I copied this from the CLI repo and didn't change anything, maybe for other use cases we do need it? let's discuss tomorrow :)

haritamar (approved):

PR #401, opened 2023-07-06
[ELE-1195] Adding additional hourly seasonality option and adding seasonality as an option to freshness tests

mossyyy:
This pull request adds the following functionality:
1. Allows anomaly freshness tests to utilise seasonality for loads that are not uniform over time. A practical example of this would be a loading a table that has constant load throughout the day and additional bulk population through a batch process such that there's expected lag over 1 or 2 hours of the day loading. 
2. Adds two new seasonality methods: `hour_of_day` and `hour_of_week`. Hour of day doesn't enforce or automated a backfill period due to the fact that you may want quite a tight training window on it and each day you're still collecting 1 training point per bucket. `hour_of_week` follows the same enforcement as `day_of_week` as throughout a given week you only get one training point for each seasonality bucket.

Both these changes have been tested on snowflake by:
- Pulling this branch into my dbt dev environment and running these tests on large volume production like datasets successfully and with (expected) failures.

It's worth noting that utilising `hour_of_week` will potentially cause quite a volume of data to be generated and extended load times with the on-run-end hook as you're utilising similar training windows to `day_of_week` but generating 24x the number of data points.

This should close out [[ELE-1195] ](https://github.com/elementary-data/elementary/issues/962)


elongl:
Hi @mossyyy , thanks a lot for contributing this.
Can you pull the [following branch](https://github.com/elementary-data/dbt-data-reliability/tree/ele-1275-add-new-seasonality-contribution-use-cases-to-the-e2e) into yours?
Thanks.

macros/utils/cross_db_utils/hour_of_week.sql
```diff
@@ -0,0 +1,32 @@
+{% macro edr_hour_of_week_expression(date_expr) %}
+    {{ return(adapter.dispatch('edr_hour_of_week_expression','elementary')(elementary.edr_cast_as_date(date_expr))) }}
+{% endmacro %}
+
+{# Databricks, Spark: #}
+{% macro default__edr_hour_of_week_expression(date_expr) %}
+    concat(cast(DATE_FORMAT({{ date_expr }}, 'EEEE') as string), cast(HOUR({{ date_expr }}) as string))
```
elongl, line 7: Can we use the [`edr_cast_as_string`](https://github.com/elementary-data/dbt-data-reliability/blob/2e95cf900ea263341d057c552fb6fb0a7dfe41b4/macros/utils/data_types/cast_column.sql#L17) macro? The current implementation does not seem to work on all data platforms.

mossyyy: Heya! Thanks for the feedback because we're passing expressions with a jinja variable in it I think it's going to be cleaner to just use the logic that underpins `edr_cast_as_string()` if that's ok with you? 

Using the whole function I think is gonna end up being something like this unless I add another macro layer in. Could be a better way to do this but I'm not sure?
```
{% macro default__edr_hour_of_week_expression(date_expr) %}
    concat({{ elementary.edr_cast_as_string("DATE_FORMAT(" + date_expr + ", 'EEEE'))" }}, {{ elementary.edr_cast_as_string("HOUR(" + date_expr + ")" ) }})
{% endmacro %}
```

A cleaner approach seems to be just modifying my cast to use the `elementary.edr_type_string()` like this:
```{% macro default__edr_hour_of_week_expression(date_expr) %}
    concat(cast(DATE_FORMAT({{ date_expr }}, 'EEEE') as {{ elementary.edr_type_string() }}), cast(HOUR({{ date_expr }}) as {{ elementary.edr_type_string() }}))
{% endmacro %}``` Seems to keep things a bit cleaner as we're not passing a combination of string and variable components into another macro? 

I've put updated the PR to follow the second option of the above let me know your thoughts on this though!

Comment (elongl) [reply macros/utils/cross_db_utils/hour_of_week.sql]: Sounds totally reasonable 👌

**Comment by elongl** [comment]
Thanks a lot for addressing.
Testing your branch and will report shortly.

**Comment by elongl** [comment]
Hi, I've made very small changes to the tests.
Can you please re-pull [this branch](https://github.com/elementary-data/dbt-data-reliability/tree/ele-1275-add-new-seasonality-contribution-use-cases-to-the-e2e)?
Will merge afterwards. Thanks.

File: macros/edr/tests/test_configuration/get_days_back.sql [hunk 2]
```diff
@@ -1,6 +1,6 @@
 {% macro get_days_back(days_back, model_graph_node, seasonality=none) %}
     {% set days_back = elementary.get_test_argument('days_back', days_back, model_graph_node) %}
-    {% if seasonality and seasonality == 'day_of_week' %}
+    {% if seasonality and (seasonality == 'day_of_week' or seasonality == 'hour_of_week') %}
```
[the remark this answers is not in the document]
mossyyy: Yep, this looks better to me.

mossyyy:
@elongl merged your branch in, fingers crossed we get passing tests now! Thanks so much for your help with this by the way :)

elongl:
Hi @mossyyy, sorry for the headache, but it seems that the BigQuery implementation is not working properly [upon running the tests](https://github.com/elementary-data/dbt-data-reliability/actions/runs/5573273380/jobs/10180327421). It seems to be returning the following error:
> EXTRACT from DATE does not support the HOUR date part

The problem seems to be [here](https://github.com/elementary-data/dbt-data-reliability/pull/401/files#diff-ee97c5af72b3929011fe7454450d0620dcd52a6cf4594b5b0d599cd1ea73b399R10).
This cannot be done with a `DATE` object (casted above) and should be a timestamp one.

```sql
SELECT EXTRACT(HOUR FROM CURRENT_TIMESTAMP())
-- Works.
SELECT EXTRACT(HOUR FROM cast(CURRENT_TIMESTAMP() as date))
-- Doesn't work.
```
Care to address that please?
Rest of the tests passed and worked great!

mossyyy:
I've updated the PR but I'm keen to wrap my head around how I can do the integration tests hopefully locally or at the very least trigger the pipeline to run them so I can resolve issues. to prevent me wasting your time @elongl!

The above pipeline seems like it perma fails for me with non zero exits. When I was looking at the pipeline link you provided that got to the point of running tests that looks like there was a manual trigger. Is it expected that any pipelines triggered off me adding commits to a PR would fail due to a permission issue or something else? What's the easiest way for me to be able to run the multi warehouse pipeline and be able to self solve these issues so I can provide you guys working PRs moving forward?

mossyyy:
I just found the read.me in `integration_tests/` gonna see if I can't get that running locally!

elongl:
@mossyyy Totally understand and no worries at all!
This is a problem that we're working on.
The difficulty is that our tests work against the various data platforms and for security-purposes we cannot allow PRs from contributors to run against our instances because the contributed code might leak the secrets (credentials).
Once again, thanks a lot for your contribution. Merging!

elongl (approved):

PR #402, opened 2023-07-06
Ele 1246 calculate failures count

ofek1weiss:
Added failed count field to test results


macros/utils/common_test_configs.sql
```diff
@@ -0,0 +1,393 @@
+{% macro get_common_test_config_by_flattened_test(flattened_test) %}
+  {% set test_name = flattened_test['short_name'] %}
+  {% set test_namespace = flattened_test['test_namespace'] %}
+  {% do return(elementary.get_common_test_config_by_namespace_and_name(test_namespace, test_name)) %}
+{% endmacro %}
+
+{% macro get_common_test_config_by_namespace_and_name(test_namespace, test_name) %}
```
elongl, line 7: I think we can remove the `__by_namespace_and_name` from the macro's name.
Typically the parameters received by a macro shouldn't be repeated in its name.

macros/edr/materializations/tests/handle_failed_test_result_count.sql
```diff
@@ -0,0 +1,63 @@
+{% macro handle_failed_test_result_count(flattened_test) %}
```
elongl, line 1: I think that it'd be better to name this something like `cache_failed_test_result_count` in order to imply what it actually does.

macros/edr/materializations/tests/handle_failed_test_result_count.sql
```diff
@@ -0,0 +1,63 @@
...
+  {% set result_count_query = elementary.get_failed_test_result_count_query(flattened_test) %}
+  {% if result_count_query %}
+    {% set result_count = run_query(result_count_query).columns[0].values()[0] %}  
+  {% else %}
+    {% set result_count = none %}
```
elongl, line 6: Is there a point in caching the result count if it's `null`?
Maybe just not add it there?

macros/edr/materializations/tests/handle_failed_test_result_count.sql
```diff
@@ -0,0 +1,63 @@
...
+  {% endif %}
+  {% do elementary.get_cache('elementary_test_failed_count').update({model.unique_id: result_count}) %}
+{% endmacro %}
+
+{% macro get_failed_test_result_count_query(flattened_test) %}
+  {% set failed_count_config = elementary.get_common_test_failed_count_config(flattened_test) %}
+  {% set test_meta = flattened_test['meta'] %}
+  {% if 'elementary_failed_count' in test_meta %}
```
elongl, line 14: I'm not sure this name would be clear for the user.
What do you think about `fail_count_expression`?

macros/utils/common_test_configs.sql
ofek1weiss: the issue is i have both this and `get_common_test_config_by_flattened_test`, and if i remove it from both of them then i have two functions with the same name (which i dont believe is possible in jinja)

macros/edr/materializations/tests/handle_failed_test_result_count.sql
ofek1weiss: but it runs the query as well, so i dont think `cache` is the correct word here.
I agree that it isnt the best name, but I followed the convention of `handle_test_results`,

macros/edr/materializations/tests/handle_failed_test_result_count.sql
ofek1weiss: On it

macros/edr/materializations/tests/handle_failed_test_result_count.sql
```diff
@@ -0,0 +1,63 @@
+{% macro handle_failed_test_result_count(flattened_test) %}
+  {% set result_count_query = elementary.get_failed_test_result_count_query(flattened_test) %}
+  {% if result_count_query %}
+    {% set result_count = run_query(result_count_query).columns[0].values()[0] %}
```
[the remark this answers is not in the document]
ofek1weiss: did not know it exists, awesome

macros/edr/materializations/tests/handle_failed_test_result_count.sql
elongl: Got it, this makes sense to me.
What do you think about changing this macro to only return the fail count and rename to something very simplistic such as `get_test_fail_count` and then cache its value [here](https://github.com/elementary-data/dbt-data-reliability/blob/fc6e2caeff13e9d363ae6394217ab8cbbce4852a/macros/edr/materializations/tests/test.sql#L96)?

ofek1weiss: That seems right, will do

macros/edr/materializations/tests/handle_failed_test_result_count.sql
ofek1weiss: changed to `failed_count_expression` as it is really more easy, understandable and versitile, as to your point about `fail_calc`, it seems to not exactly match our use case here

macros/edr/materializations/tests/failed_test_result_count.sql
```diff
@@ -0,0 +1,34 @@
...
+    {% set result_count_query = elementary.get_failed_count_calc_query(failed_count_calc) %}
+    {% do return(result_count_query) %}
+  {% endif %}
+  {% do return(none) %}
+{% endmacro %}
+
+{% macro get_failed_count_calc(flattened_test) %}
+  {% set common_test_config = elementary.get_common_test_config(flattened_test) %}
+  {% if common_test_config %}
+    {% if "failed_count_calc" in common_test_config %}
```
elongl, line 22:
```py
      {% do return(common_test_config.get("failed_count_calc") %}
```
Instead of the `if`.

macros/edr/materializations/tests/test.sql
```diff
@@ -1,3 +1,16 @@
+{% macro create_test_result_temp_table() %}
```
elongl, line 1: Shall we maybe use [this macro](https://github.com/elementary-data/dbt-data-reliability/blob/6b8f5aa55b20743e99bf96e102b13dff2890e453/macros/utils/table_operations/create_temp_table.sql#L0-L1)?

Also, I'd clarify what `model["alias"]` means in the context of a test by exporting it to a variable.

macros/edr/materializations/tests/test.sql
```diff
@@ -1,3 +1,16 @@
...
+  {% set database, schema = elementary.get_package_database_and_schema() %}
+  {% set relation = make_temp_relation(api.Relation.create(identifier=model["alias"], database=database, schema=schema, type="table")) %}
+  {% call statement("temp_table") %}
+    {{ create_table_as(True, relation, sql) }}
+  {% endcall %}
+  {% set new_sql %}
+    select * from {{ relation }}
+  {% endset %}
+  {% do context.update({"sql": new_sql}) %}
```
elongl, line 10: This is a side-effect for this function, it's not clear that it overrides the `sql` context.
I think we should do that after the call to `create_test_result_temp_table` in the caller function.

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -53,6 +53,8 @@
     'anomaly_direction': 'both',
     'store_result_rows_in_own_table': true,
     'mute_dbt_upgrade_recommendation': false,
+    'calculate_failed_count': true,
+    'use_temp_tables': true
```
elongl, line 57: We should clarify that this is in the context of tests.
Think about how you can convey that in the naming.

macros/edr/materializations/tests/failed_test_result_count.sql
```diff
@@ -0,0 +1,34 @@
+{% macro get_failed_test_result_count(flattened_test) %}
```
elongl, line 1: The naming seems a bit inconsistent.
Do you think we should maybe call it `failed_count` everywhere?

macros/edr/system/system_utils/empty_table.sql
```diff
@@ -26,7 +26,8 @@
     ('failures', 'bigint'),
     ('test_short_name', 'string'),
     ('test_alias', 'string'),
-    ('result_rows', 'long_string')
+    ('result_rows', 'long_string'),
+    ('failed_count', 'bigint')
```
elongl, line 30: I think it should be `failed_row_count` or `row_count_failures`, otherwise it's really confusing compared to `failures`.

macros/edr/materializations/tests/failed_test_result_count.sql
ofek1weiss: i dont want to do that, because if the key does not exist, i want it to get to line 26

macros/edr/materializations/tests/test.sql
ofek1weiss: 👍

macros/edr/materializations/tests/test.sql
ofek1weiss: 👍

macros/edr/materializations/tests/failed_test_result_count.sql
ofek1weiss: changed it to `failed_row_count` everywhere (as per your next comment)

macros/edr/system/system_utils/empty_table.sql
ofek1weiss: on it

macros/utils/common_test_configs.sql
```diff
@@ -0,0 +1,393 @@
...
+    }
+    The config may contain:
+      description: A human readable description of the test.
+      failed_count: Configuration to get the number of failed rows from the test result, see handle_failed_test_result_count.sql for details.
+      quality_dimension: The quality dimension of the test, see https://www.precisely.com/blog/data-quality/data-quality-dimensions-measure for details.
+  #}
+  {% set common_tests_configs_mapping = {
+      "dbt": {
+        "not_null": {
+          "failed_count": {
```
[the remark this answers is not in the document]
ofek1weiss: changed to `failed_row_count_calc`

macros/edr/system/system_utils/get_config_var.sql
ofek1weiss: on it

elongl (approved):

PR #404, opened 2023-07-07
Fixes to dimension anomalies and added incremental to e2e tests

Maayan-s:
The changes to `dimension anomalies` are to handle new dimensions.
We used to create empty buckets for these new dimensions, so their appearance created an anomaly. 
On this PR: 
- New dimensions are excluded from empty buckets fill. 
- We don't fill buckets that are older than the dimension first appearance.

### DIFF
--- a/integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
+++ b/integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
@@ -23,5 +23,5 @@
         {%- endif %}
     {% endfor %}
 
-    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform', 'elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
+    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
 {% endmacro %}
--- a/integration_tests/models/copy_numeric_column_anomalies.sql
+++ b/integration_tests/models/copy_numeric_column_anomalies.sql
@@ -1,1 +1,13 @@
-select * from {{ ref("numeric_column_anomalies") }}
+select
+        {{ elementary.edr_cast_as_timestanp('updated_at') }},
+        occurred_at,
+        min,
+        max,
+        zero_count,
+        zero_percent,
+        average,
+        standard_deviation,
+        variance,
+        sum
+
+from {{ ref("numeric_column_anomalies") }}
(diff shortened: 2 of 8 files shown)


integration_tests/run_e2e_tests.py
```diff
@@ -143,6 +143,20 @@ def e2e_tests(
         ]
         test_results.extend(results)
 
+# Calculating metrics in training stage to test incremental models monitoring
```
Maayan-s, line 146: We didn't test the incremental use case, and we should.

integration_tests/models/schema.yml
```diff
@@ -121,7 +126,7 @@ models:
       - elementary.dimension_anomalies:
           dimensions:
             - platform
-          tags: ["dimension_anomalies", "should_fail"]
+          tags: ["dimension_anomalies"]
```
Maayan-s, line 129: It was a bug that this test was failing.

integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
```diff
@@ -23,5 +23,5 @@
         {%- endif %}
     {% endfor %}
 
-    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform', 'elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
+    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
```
Maayan-s, line 26: The test I removed shouldn't have failed, it was a bug.

integration_tests/run_e2e_tests.py
```diff
@@ -203,7 +217,10 @@ def e2e_tests(
         dbt_runner.test(select="tag:schema_changes")
         dbt_runner.test(select="tag:schema_changes_from_baseline")
 
-    dbt_runner.run(vars={"stage": "validation"})
+    dbt_runner.run(
+        select="tag:validation_stage",
```
Maayan-s, line 221: Makes the tests a bit faster.

integration_tests/models/copy_numeric_column_anomalies.sql
```diff
@@ -1 +1,13 @@
-select * from {{ ref("numeric_column_anomalies") }}
+select
+        {{ elementary.edr_cast_as_timestanp('updated_at') }},
```
Maayan-s, line 2: IDK why but this not being a timestamp made e2e fail

integration_tests/models/copy_numeric_column_anomalies.sql
```diff
@@ -1 +1,13 @@
-select * from {{ ref("numeric_column_anomalies") }}
+select
+        {{ elementary.edr_cast_as_timestamp('updated_at') }},
```
Maayan-s, line 2: IDK why by all the e2e failed on this not being a timestamp column

PR #405, opened 2023-07-08
Ele 1192 handling new dimension values

Maayan-s:
The changes to dimension anomalies query are to handle new dimensions.
We used to create empty buckets for these new dimensions, so their appearance created an anomaly.

On this PR:
- New dimensions are excluded from empty buckets fill.
- We don't fill buckets that are older than the dimension first appearance.

There is a product decision we should make - 
Is a new dimension an anomaly? Should this be a flag maybe?

### DIFF
--- a/integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
+++ b/integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
@@ -25,3 +25,17 @@
 
     {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
 {% endmacro %}
+
+{% macro create_new_dimension() %}
+    {% set dimension_validation_data = ref('dimension_anomalies_validation') %}
+    {%- set insert_dimension_query %}
+        INSERT INTO {{ dimension_validation_data }} values ('1969-12-28 00:00:00.000', 'windows', 1, 318);
+    {% endset %}
+{% endmacro %}
+
+{% macro delete_new_dimension() %}
+    {% set dimension_validation_data = ref('dimension_anomalies_validation') %}
+    {%- set delete_dimension_query %}
+        DELETE FROM{{ dimension_validation_data }} WHERE platform = 'windows';
+    {% endset %}
+{% endmacro %}
--- a/integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
+++ b/integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
@@ -23,5 +23,5 @@
         {%- endif %}
     {% endfor %}
 
-    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform', 'elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
+    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
 {% endmacro %}
--- a/integration_tests/models/schema.yml
+++ b/integration_tests/models/schema.yml
@@ -121,7 +121,7 @@ models:
       - elementary.dimension_anomalies:
           dimensions:
             - platform
-          tags: ["dimension_anomalies", "should_fail"]
+          tags: ["dimension_anomalies"]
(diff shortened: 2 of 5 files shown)


integration_tests/macros/e2e_tests/validate_dimensions_anomalies.sql
```diff
@@ -23,5 +23,5 @@
         {%- endif %}
     {% endfor %}
 
-    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform', 'elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
+    {{ assert_lists_contain_same_items(should_fail_names, ['elementary_dimension_anomalies_dimension_anomalies_platform__updated_at', 'elementary_dimension_anomalies_dimension_anomalies_platform__version__updated_at']) }}
```
Maayan-s, line 26: It was a bug that this test was failing

integration_tests/models/schema.yml
```diff
@@ -121,7 +121,7 @@ models:
       - elementary.dimension_anomalies:
           dimensions:
             - platform
-          tags: ["dimension_anomalies", "should_fail"]
+          tags: ["dimension_anomalies"]
           alias: "dimension_anomalies_no_timestamp"
```
Maayan-s, line 125: It was a bug that this test was failing

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -17,75 +25,93 @@
...
-                    dimension_value,
-                    1 as joiner
-                from dimension_values_without_outdated
-                union all
-                select distinct
-                    dimension_value,
-                    1 as joiner
-                from filtered_monitored_table
-            ) results
+        unique_previous_dimension_values as (
```
IDoneShaveIt, line 61: Could they be not unique?
On `training_set_dimensions` we already group by the dimension value and distinct it.
This table will be the same without the SUM and wit a joiner.
Maybe we could just add the joiner to the `training_set_dimensions` table?

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -17,75 +25,93 @@
...
 
-        {# Created buckets for each dimension value #}
+        {# Create buckets for each previous dimension value #}
         dimensions_buckets as (
             select edr_bucket_start, edr_bucket_end, dimension_value
-            from buckets left join dimension_values_union on buckets.joiner = dimension_values_union.joiner
+            from unique_previous_dimension_values left join buckets
+                on (buckets.joiner = unique_previous_dimension_values.joiner
+                {# This makes sure we don't create empty buckets for dimensions before their first apperance #}
+                and edr_bucket_end >= dimension_min_bucket_end)
```
IDoneShaveIt, line 75: Like!
Can we add a test case for it in our e2e?

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
```diff
@@ -17,75 +25,93 @@
...
+                on (dimensions_buckets.edr_bucket_start = start_bucket_in_data and dimensions_buckets.dimension_value = row_count_values.dimension_value)
+        ),
+
+        union_row_count_values as (
+            select distinct *
+            from
+            (
+            select * from row_count_values
+              union all
+            select * from fill_empty_buckets_row_count_values
```
IDoneShaveIt, line 113: `fill_empty_buckets_row_count_values` is built from `from dimensions_buckets left join row_count_values`.
Wouldn't union it with `row_count_values` is redundant and only causing  duplicate rows that we distinct them?
If not, please add a comment that explain this because it is hard to understand it frm reading the code 🙂

IDoneShaveIt: For reference, on the no timestamp part, the `fill_empty_buckets_row_count_values` is really only the filled values

IDoneShaveIt (commented):
Looking good!
Left a few comments, and lets add a test for it on the e2e 🙂

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
Maayan-s: We union so new buckets added in this run will be included (were filtered by the join we did on 'fill_empty_buckets_row_count_values'). I added a comment

macros/edr/data_monitoring/monitors_query/dimension_monitoring_query.sql
Maayan-s: Added, a bit hacky but @elongl is refactoring the e2e

IDoneShaveIt (approved):
LGTM!

PR #406, opened 2023-07-08
Added incremental to e2e

Maayan-s:


integration_tests/models/any_type_column_anomalies.sql
```diff
@@ -1,5 +1,12 @@
+{{ config(
+    materialized="incremental"
+) }}
+
 with training as (
     select * from {{ ref('any_type_column_anomalies_training') }}
+    {% if is_incremental() %}
+        where updated_at > (select max(updated_at) from {{ this }})
```
elongl, line 8: Why not `where 1 = 0` just to clarify?
Also can we add a comment to convey what we're trying to do?

PR #415, opened 2023-07-11
POC: dbt package tests.

elongl:


integration_tests_v2/tests/test_volume_anomalies.py
```diff
@@ -0,0 +1,70 @@
...
+    results = dbt_project.read_table(
+        "elementary_test_results",
+        where=f"table_name = '{test_id}'",
+        column_names=["status"],
+    )
+    return results
+
+
+def test_anomalyless_table_volume_anomalies(request, dbt_project: DbtProject):
+    test_id = request.node.name
```
haritamar, line 21: Consider making a fixture for the `test_id` instead of every test receiving a request object

haritamar (approved):

integration_tests_v2/tests/test_volume_anomalies.py
elongl: That's just so much better 🤦🏽

PR #423, opened 2023-07-17
elementary tests: add support for dbt "where" parameter to the tests

haritamar:


Maayan-s:
Looks good to me. 
I think you can just add `where: 1=1` config to one test of each type in the schema.yml and that will give us confidence. 

Also we should deprecate `where_expression` and just guide users to use `where`. No?

PR #425, opened 2023-07-18
add exception to upload source freshness 

Maayan-s:


macros/edr/dbt_artifacts/upload_source_freshness.sql
```diff
@@ -4,6 +4,10 @@
   {% endif %}
 
   {% set source_freshness_results_relation = ref('dbt_source_freshness_results') %}
+  {% if not source_freshness_results_relation %}
+    {% do exceptions.raise_compiler_error('Unable to upload source freshness as "dbt_source_freshness_results" table was not found in ' ~ target.database ~ '.' ~ target.schema ) %}
```
elongl, line 8: It should be in the Elementary schema, not the user's schema.
Think you should use:
```sql
        {% set database_name, schema_name = elementary.get_package_database_and_schema() %}
```

Maayan-s: @elongl 
This is executed by `edr`, so the target db and schema are the ones in the `elementary` profile. No?

elongl: This macro is not called anymore from anywhere.
I think you're referring to [this macro](https://github.com/elementary-data/elementary/blob/29eef4c29c2aea34cad0a642bdfb551a27908c83/elementary/monitor/dbt_project/macros/upload_source_freshness.sql#L0-L1).

`dbt run-operation elementary.upload_source_freshness` (Macro you're editing)--> `edr run-operation upload-source-freshness`.

PR #432, opened 2023-07-20
POC Metrics collection.

elongl:


integration_tests/integration_tests/tests/test_metrics.py
```diff
@@ -0,0 +1,39 @@
...
+        {"updated_at": date.strftime(DATE_FORMAT)}
+        for date in generate_dates(base_date=datetime.now())
+        for _ in range(random.randint(-5, 20))
+    ]
+    data2 = [
+        {"created_at": date.strftime(DATE_FORMAT)}
+        for date in generate_dates(base_date=datetime.now())
+        for _ in range(random.randint(0, 20))
+    ]
+    dbt_project.seed(data1, "first_metrics_table_seed")
```
elongl, line 19: Should we implement `seed_bulk()` for performance?
Alternatively we can just implement the various data seeders per adapter.

macros/edr/materializations/model/common.sql
```diff
@@ -0,0 +1,17 @@
+{% macro query_row_count_metric() %}
```
ofek1weiss, line 1: make this 'get_table_metrics`

macros/edr/materializations/model/common.sql
```diff
@@ -0,0 +1,17 @@
...
+  {% set query %}
+    select count(*) as row_count
+    from {{ this }}
+  {% endset %}
+  {% do return(elementary.result_value(query)) %}
+{% endmacro %}
+
+{% macro cache_row_count_metric(row_count) %}
```
ofek1weiss, line 9: then change this to be `cache_table_metrics`, and iterate over the fields and make a different record for each one (for now it will only be the row count, but it will make it much easier to add more metrics with less queries)

PR #433, opened 2023-07-22
Ele 1353 add where tests to new package tests infra

haritamar:


integration_tests/integration_tests/tests/conftest.py
```diff
@@ -8,6 +8,12 @@ def pytest_addoption(parser):
     parser.addoption("--target", action="store", default="postgres")
 
 
+def pytest_configure(config):
```
elongl, line 11: Isn't it better to [use `pytest.ini`](https://docs.pytest.org/en/latest/how-to/mark.html#registering-marks) for that?

integration_tests/integration_tests/tests/conftest.py
```diff
@@ -26,6 +32,14 @@ def init_tests_env(target, tmp_path_factory, worker_id: str):
             env_ready_indicator_path.touch()
 
 
+@pytest.fixture(autouse=True)
+def skip_by_targets(request, target):
+    if request.node.get_closest_marker("skip_targets"):
+        skipped_targets = request.node.get_closest_marker("skip_targets").args
```
elongl, line 38: From this code I understand that `skip_targets` takes a list, but it seems that you pass `"databricks"`.
What am I missing?

integration_tests/integration_tests/tests/dbt_project.py
```diff
@@ -98,15 +142,32 @@ def test(
...
+        else:
+            return self._read_single_test_result(test_id)
 
     def seed(self, data: List[dict], table_name: str):
         return DbtDataSeeder(self.dbt_runner).seed(data, table_name)
 
-    def _read_test_result(self, table_name: str) -> Dict[str, Any]:
-        return self.read_table(
+    def _read_test_results(self, table_name: str) -> List[Dict[str, Any]]:
+        test_execution_id = self.read_table(
```
elongl, line 155: 1. Can we maybe do this with a `JOIN` rather than two separate queries?
2. I think this is problematic as we add more tests because it'll be much more likely that tests will run in parallel and therefore this might return the results of a different test. I think we need to think about a way to extract the invocation id and then filter by it.

integration_tests/integration_tests/tests/dbt_project.py
```diff
@@ -98,15 +142,32 @@ def test(
...
+            column_names=["test_execution_id"],
             limit=1,
-        )[0]
+        )[0]["test_execution_id"]
+        return self.read_table(
+            "elementary_test_results",
+            where=f"test_execution_id = '{test_execution_id}'",
+        )
+
+    def _read_single_test_result(self, table_name: str) -> Dict[str, Any]:
```
elongl, line 167: Personally don't think the `single` is really necessary given that it's `result` and not `results`.

integration_tests/integration_tests/tests/test_all_columns_anomalies.py
```diff
@@ -0,0 +1,152 @@
...
+DBT_TEST_NAME = "elementary.all_columns_anomalies"
+DBT_TEST_ARGS = {
+    "timestamp_column": TIMESTAMP_COLUMN,
+    "column_anomalies": ["null_count"],
+}
+
+
+def test_anomalyless_all_columns_anomalies(test_id: str, dbt_project: DbtProject):
+    dates = generate_dates(base_date=date.today() - timedelta(1))
+    data: List[Dict[str, Any]] = sum(
```
elongl, line 17: Not sure I understand the requirement of using a `sum()` here (also to the rest).

integration_tests/integration_tests/tests/test_volume_anomalies.py
```diff
@@ -10,38 +11,82 @@
...
+        for cur_date in generate_dates(base_date=date.today())
+        for _ in range(1 if cur_date < cur_date.today() - timedelta(days=1) else 2)
     ]
     test_result = dbt_project.test(data, test_id, DBT_TEST_NAME, DBT_TEST_ARGS)
     assert test_result["status"] == "fail"
+
+
+@Parametrization.autodetect_parameters()
+@Parametrization.case(name="source", as_model=False)
+@Parametrization.case(name="model", as_model=True)
```
elongl, line 53: Does the `name` do something?

integration_tests/integration_tests/tests/conftest.py
haritamar: I made it accept *args but maybe it's confusing
I'll change to a list

integration_tests/integration_tests/tests/conftest.py
haritamar: Yeah, I'll change

integration_tests/integration_tests/tests/dbt_project.py
haritamar: I'll change to a JOIN (or a subquery)
Re. the parallel execution - I think the table name we filter on will be different, no?

integration_tests/integration_tests/tests/dbt_project.py
haritamar: It's not a must but I think it is a bit more explicit

integration_tests/integration_tests/tests/test_volume_anomalies.py
haritamar: it affects the test id
will be `test_volume_anomalies_with_where_parameter[model]` or `test_volume_anomalies_with_where_parameter[source]`

integration_tests/integration_tests/tests/test_all_columns_anomalies.py
elongl: Ohh, I see, personally find [this syntax](https://github.com/elementary-data/dbt-data-reliability/pull/433/files/781fee3d28ff3e9a99dff0198faffa8d3b22fef6#diff-c47f996a703879330b835a9dd776b9c32cbfa5a1184c13fbbed50a8a9e98b46dR33) clearer but not critical.

elongl (approved):

PR #434, opened 2023-07-23
Ele 1165 create a macro to generate the cloud profile

ofek1weiss:


macros/commands/create_elementary_user.sql
```diff
@@ -0,0 +1,34 @@
+{% macro create_elementary_user(user="elementary", password=none, role="ELEMENTARY_ROLE") %}
```
elongl, line 1: 1. Why is the role in caps?
2. Any reason not to call the role just `elementary`?

macros/utils/cross_db_utils/generate_elementary_profile_parameters.sql
```diff
@@ -0,0 +1,104 @@
+{% macro generate_elementary_profile_parameters(method=none, overwrite_values=none) %}
+  {#
+    Returns a list of parameters for elementary profile.
+    Each parameter consists of a list consisting of:
```
elongl, line 4: I tend to think that this should just be a list of dicts instead.

macros/utils/cross_db_utils/generate_elementary_profile_parameters.sql
```diff
@@ -0,0 +1,104 @@
+{% macro generate_elementary_profile_parameters(method=none, overwrite_values=none) %}
```
elongl, line 1: Maybe rename to `generate_elementary_profile_args`?

macros/commands/create_elementary_user.sql
```diff
@@ -0,0 +1,34 @@
...
+    {% endif %}
+  {% endfor %}
+{% endmacro %}
+
+
+{% macro generate_password() %}
+  {# Heavily inspired by https://gist.github.com/benwalio/8598cf9f642271ffdaf3daa82c1802cb #}
+  {% set ns = namespace(placeholder="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", password="") %}
+  {% for x in ns.placeholder %}
+    {% set ns.password = [ns.password, (x | replace('x', [0,1,2,3,4,5,6,7,8,9,'a','b','c','d','e','f','A','B','C','D','E','F'] | random ))] | join %}
```
elongl, line 31: Any reason this is only hex characters?
Maybe do something more like [this](https://stackoverflow.com/a/39596292)?
Guess not that critical.

macros/commands/create_elementary_user.sql
ofek1weiss: from what i gathered when researching, role names are almost always all caps, and usually (thought not always, and this may just be a guild thing) end with the `_ROLE` suffix

macros/utils/cross_db_utils/generate_elementary_profile_parameters.sql
ofek1weiss: sounds reasonable

macros/utils/cross_db_utils/generate_elementary_profile_parameters.sql
ofek1weiss: ok

macros/commands/create_elementary_user.sql
ofek1weiss: could not find anything like the `strings` module in jinja, and was a bit lazy..., but i cant do it.

macros/commands/create_elementary_user.sql
elongl: Cool. We'll learn if that's not the case.

macros/utils/cross_db_utils/generate_elementary_profile_args.sql
```diff
@@ -0,0 +1,107 @@
...
+  {% do return(parameters) %}
+{% endmacro %}
+
+{% macro _parameter(name, value, comment=none) %}
+  {% do return({"name": name, "value": value, "comment": comment}) %}
+{% endmacro %}
+
+{% macro snowflake__generate_elementary_profile_args(method, elementary_database, elementary_schema) %}
+  {% do return([
+    _parameter("type", target.type),
```
elongl, line 27: Technically this could be overwritten by a user with the same macro because we don't prefix with `elementary.`, but it's extremely unlikely so I suppose that's negligible.

elongl (approved):

PR #435, opened 2023-07-23
Renamed a macro.

elongl:


ofek1weiss (commented):
I think the renaming in python is unnecessary

It's very intuitive that `DbtProject.run_query` renders the query as a jinja template, like everything else in dbt, and is cleaner and more understandable than `render_run_query`

ofek1weiss (approved):

PR #443, opened 2023-07-28
Ele 1254 table name missing for sources relation tests

Nanoro-it:
https://github.com/elementary-data/elementary/issues/853


haritamar (approved):

haritamar:
Hi @Nanoro-it !
Thanks for your contribution.
I changed it a bit so it will consider both `ref` and `source` instead of one after the other.
I think it's good to be merged

PR #447, opened 2023-08-01
Ele 1453 implement metrics collection

elongl:


macros/edr/tests/on_run_end/insert_metrics.sql
```diff
@@ -0,0 +1,12 @@
+{% macro insert_metrics() %}
+  {% set metrics = elementary.get_cache("tables").get("metrics") %}
+  {% set database_name, schema_name = elementary.get_package_database_and_schema() %}
+  {%- set target_relation = adapter.get_relation(database=database_name, schema=schema_name, identifier='data_monitoring_metrics') -%}
+  {% if not target_relation %}
+    {% do exceptions.raise_compiler_error("Couldn't find Elementary's models. Please run `dbt run -s elementary`.") %}
```
elongl, line 6: Should this raise or return?

macros/edr/materializations/model/common.sql
```diff
@@ -0,0 +1,28 @@
+{% macro query_table_metrics() %}
+  {% set query %}
+    select count(*) as row_count
```
elongl, line 3: The freshness metrics (model run timestamp) can be calculated from this metric.
Should we add an empty `freshness` or `build` metric for clarity?

macros/edr/materializations/model/metrics.sql
```diff
@@ -0,0 +1,29 @@
+{% macro query_table_metrics() %}
+  {% set query %}
+    select count(*) as row_count
```
elongl, line 3: The freshness metrics (model run timestamp) can be calculated from this metric.
Should we add an empty freshness or build metric for clarity?

macros/edr/materializations/model/metrics.sql
```diff
@@ -0,0 +1,29 @@
+{% macro query_table_metrics() %}
```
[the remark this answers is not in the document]
elongl: I prefer the duplication for now, `table_monitoring_query` is extremely complex in my opinion.
This is very simple and straight-forward, I'd rather keep it that way in spite of duplication.

Regarding the first point, maybe I don't fully understand what you meant, but I don't find:
```sql
{% macro get_row_count_metric_expr() %}
count(*) as row_count
{% endmacro %}


{% macro query_table_metrics() %}
  {% set query %}
    select
    {{ elementary.get_row_count_metric_expr() }}
    from {{ this }}
  {% endset %}
...
```

More readable than:
```sql
{% macro query_table_metrics() %}
  {% set query %}
    select
      count(*) as row_count
    from {{ this }}
  {% endset %}
...
```

I like how all the metrics are centralized in one clear query.
Though, if the expressions would be more complicated than `count(*)` than I can understand.

macros/edr/tests/on_run_end/insert_metrics.sql
```diff
@@ -0,0 +1,12 @@
+{% macro insert_metrics() %}
+  {% set metrics = elementary.get_cache("tables").get("metrics") %}
+  {% set database_name, schema_name = elementary.get_package_database_and_schema() %}
+  {%- set target_relation = adapter.get_relation(database=database_name, schema=schema_name, identifier='data_monitoring_metrics') -%}
```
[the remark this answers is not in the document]
elongl: Those metrics [aren't being used in the tests](https://github.com/elementary-data/dbt-data-reliability/blob/3c4ce9e5d646ae9ac48fbf247a439d192c645038/macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql#L55) for now because they don't have metric properties.

macros/edr/materializations/model/metrics.sql
```diff
@@ -0,0 +1,35 @@
...
+      "column_name": none,
+      "metric_name": metric_name,
+      "metric_value": metric_value,
+      "updated_at": elementary.datetime_now_utc_as_string()
+    }) %}
+  {% endfor %}
+  {% do return(metrics) %}
+{% endmacro %}
+
+{% macro query_metrics() %}
```
elongl, line 25: Can we potentially store too many metrics in memory?
TODO: Think about the test metrics collection to understand that.

macros/edr/materializations/model/incremental.sql
```diff
@@ -0,0 +1,34 @@
+{% materialization incremental, default %}
+  {% set relations = dbt.materialization_incremental_default() %}
+  {% set metrics = elementary.query_metrics() %}
```
ofek1weiss, line 3: is `if execute` relevant here?

macros/edr/materializations/model/metrics.sql
```diff
@@ -0,0 +1,35 @@
+{% macro query_table_metrics() %}
+  {% set query %}
+    select
+      null as build,
```
ofek1weiss, line 4: what is `null as build`?

macros/edr/materializations/model/incremental.sql
elongl: Nope

macros/edr/materializations/model/metrics.sql
elongl: I wanted to create a metric to record the time of the model build.
This will be used for calculating the freshness.
Do you have a clearer metric name?
Maybe `build_time`?

ofek1weiss (approved):

PR #449, opened 2023-08-01
Resolve open issue: #1043

theromsh:
Resolves the following: 
https://github.com/elementary-data/elementary/issues/1043

Thread:
https://elementary-community.slack.com/archives/C02CTC89LAX/p1690388674267909

### DIFF
--- a/macros/utils/table_operations/delete_and_insert.sql
+++ b/macros/utils/table_operations/delete_and_insert.sql
@@ -57,7 +57,16 @@
 {% macro spark__get_delete_and_insert_queries(relation, insert_relation, delete_relation, delete_column_key) %}
     {% set queries = [] %}
 
-    {% if delete_relation %}
+    {% if delete_relation and relation.is_delta %}
+        {% set delete_query %}
+            merge into {{ relation }} as source
+            using {{ delete_relation }} as target
+            on (source.{{ delete_column_key }} = target.{{ delete_column_key }}) or source.{{ delete_column_key }} is null
+            when matched then delete;
+        {% endset %}
+        {% do queries.append(delete_query) %}
+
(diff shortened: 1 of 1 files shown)


macros/utils/table_operations/delete_and_insert.sql
```diff
@@ -57,7 +57,16 @@
 {% macro spark__get_delete_and_insert_queries(relation, insert_relation, delete_relation, delete_column_key) %}
     {% set queries = [] %}
 
-    {% if delete_relation %}
+    {% if delete_relation and relation.is_delta %}
```
haritamar, line 60: Cool @theromsh !
A couple of notes:

- I think it will be a bit cleaner to do:
```
{% if delete_relation %}
   {% if relation.is_delta %}
      ...
   {% else %}
      ...
   {% endif %}
{% endif %}

- Please add a comment clarifying that DELETEs are not support in the `is_delta` mode

Thanks!
```

theromsh: I don't think nesting condition should be a good practice, because it's less readable in my opinion (you always have to memorize the previous conditions).
Here's an article that explains why nested ifs should be avoided 
https://lawyerdev.medium.com/i-never-write-nested-ifs-e4e91a5440ee

haritamar: It's not very critical, but I admit I don't fully agree with that article :)
Especially since reading `{% elif delete_relation %}` forces me to remember the previous condition, and that it actually means both `delete_relation` and `not relation.is_delta`.

But I'm fine with keeping it as-is, as it's overall pretty clear.
I can merge tomorrow probably.

ellakz (approved):

PR #450, opened 2023-08-02
Change dbt columns materialization in bigquery

ofek1weiss:


macros/edr/dbt_artifacts/dbt_columns/get_dbt_columns_query.sql
```diff
@@ -0,0 +1,66 @@
+{% macro get_dbt_columns_query(is_on_run_end=false) %}
+  {% if is_on_run_end %}
+    {% set get_relation = elementary.get_elementary_relation %}
```
elongl, line 3: Not sure if we care, but just a thought, if a user installs Elementary and doesn't run our models, this would return `None` and then the query would be `from None`.

macros/edr/dbt_artifacts/dbt_columns/get_dbt_columns_query.sql
```diff
@@ -0,0 +1,66 @@
+{% macro get_dbt_columns_query(is_on_run_end=false) %}
```
elongl, line 1: The true distinction that you want to make is whether you're in a model build context or not so I'd rename this to `in_model_build_context` or something like that. For instance, in the `on-run-start`, materializations, or other places, it's still meaningful.

macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
```diff
@@ -26,6 +26,10 @@
         {% do elementary.file_log('[{}] Artifacts already ran.'.format(artifacts_model)) %}
       {% endif %}
     {% endfor %}
+    {% if elementary.get_dbt_columns_materialization() != "view" %}
+      {% do adapter.commit() %}
```
elongl, line 30: Is this needed? There's `should_commit=true` for the uploads.

macros/edr/dbt_artifacts/dbt_columns/upload_dbt_columns.sql
```diff
@@ -0,0 +1,11 @@
+{% macro upload_dbt_columns() %}
+  {% set relation = elementary.get_elementary_relation('dbt_columns') %}
+  {% if execute and relation %}
+    {% set query %}
```
elongl, line 4: Use [`create_or_replace`](https://github.com/elementary-data/dbt-data-reliability/blob/3c4ce9e5d646ae9ac48fbf247a439d192c645038/macros/utils/table_operations/create_or_replace.sql#L0-L1).

macros/edr/dbt_artifacts/dbt_columns/get_dbt_columns_materialization.sql
```diff
@@ -0,0 +1,16 @@
...
+{% endmacro %}
+
+
+{% macro default__get_dbt_columns_materialization() %}
+  {% do return("view") %}
+{% endmacro %}
+
+
+{% macro bigquery__get_dbt_columns_materialization() %}
+  {% do return("incremental") %}
```
elongl, line 15: Why `incremental` and not `table`?
The `incremental` for our other artifacts is meant so that the `table` materialization won't truncate them and disrupt the caching, but here it's not relevant.

macros/edr/dbt_artifacts/dbt_columns/get_dbt_columns_query.sql
ofek1weiss: if he does not run our models, this wont run as well

macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
ofek1weiss: but we need to commit before running the `dbt_columns` query, as it queries the other artifacts

macros/edr/dbt_artifacts/dbt_columns/upload_dbt_columns.sql
elongl: What do you mean? There's a link in my comment, lol.

macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
elongl: Yes, but I'm saying that the there are no pending transactions at this point because of the [`should_commit=True`](https://github.com/elementary-data/dbt-data-reliability/blob/b618ea2ef9f689d31ae55c1c41135dbd0d2be505/macros/edr/dbt_artifacts/upload_dbt_artifacts.sql#L20) and each model commits when it finishes executing.

elongl (approved):

macros/edr/dbt_artifacts/dbt_columns/get_dbt_columns_materialization.sql
darrylng: Concur on this point. `dbt_columns` don't need to be materialized as incremental as we get the tables info and columns info every time this model is run.

PR #451, opened 2023-08-02
integration tests - fix test runs on models

haritamar:


integration_tests/integration_tests/tests/dbt_project.py
```diff
@@ -128,11 +129,13 @@ def test(
                 {"name": test_column, "tests": [{dbt_test_name: test_args}]}
             ]
 
+        temp_model_context: Any
```
elongl, line 132: Can we just rename this to `temp_table_ctx` because it doesn't make sense to be called a model when `as_model=False`.

elongl (approved):

PR #452, opened 2023-08-03
Resolve open issue: #1048

theromsh:
Resolves https://github.com/elementary-data/elementary/issues/1048

Thread:
https://elementary-community.slack.com/archives/C02CTC89LAX/p1690360303394909

### DIFF
--- a/macros/utils/table_operations/make_temp_relation.sql
+++ b/macros/utils/table_operations/make_temp_relation.sql
@@ -16,6 +16,10 @@
     {% do return(tmp_relation) %}
 {% endmacro %}
 
+{% macro databricks__edr_make_temp_relation(base_relation, suffix) %}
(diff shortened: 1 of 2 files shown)


macros/utils/table_operations/make_temp_relation.sql
```diff
@@ -1,11 +1,32 @@
+{% macro _make_temp_relation(base_relation, suffix=none) %}
```
theromsh, line 1: I added _ prefix because I didn't want to override the original adapters' macros (otherwise I won't be able to call `dbt.make_temp_relation` line 6)

haritamar: Makes total sense
Can you perhaps call it `edr_make_temp_relation`?
Just because we used a similar convention elsewhere

ellakz (approved):

PR #455, opened 2023-08-03
ELE-1479 fix dump table

IDoneShaveIt:


macros/commands/dump_table.sql
```diff
@@ -45,9 +45,14 @@
             {{ elementary.escape_select(column_names) }}, 
             row_number() over (partition by {{ dedup_by_column }} order by {{ order_by_dedup_column }} desc) as row_index
         from {{ relation }}
+    ),
+
+    deduped_relation as (
+        select {{ elementary.escape_select(column_names) }}
```
elongl, line 51: Maybe do `SELECT *` here and below?

elongl (approved):

PR #486, opened 2023-08-13
changed`filtered_information_schema_columns` materialization

harukitake:
- https://github.com/elementary-data/elementary/issues/945
- https://github.com/elementary-data/elementary/issues/541

I'm in trouble with this issue. 
I'm using Snowflake, and when I looked at the query profile, I found that the test was taking a long time because the query compile time was long. By changing the materialization from view to incremental, I expect to reduce the compile time and improve the performance of baseline schema change tests.

### DIFF
--- a/models/edr/metadata_store/filtered_information_schema_columns.sql
+++ b/models/edr/metadata_store/filtered_information_schema_columns.sql
@@ -2,6 +2,7 @@
   config(
     materialized="incremental",
     incremental_strategy="delete+insert",
+    unique_key="full_table_name"
   )
 }}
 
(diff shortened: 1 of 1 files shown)


harukitake:
@elongl Can you review?

Since I changed the model incremental, I think it's necessary to build the model before running the baseline schema tests.
I want to discuss how to fix for it.

elongl:
@harukitake As you correctly said, this table needs to be a view so that if new tables are added, it'll still work without requiring the user to re-run this model prior to using schema changes tests. We need to come up with an alternative solution in order to cut down the time. Perhaps we should query the table directly rather than the information schema.

harukitake:
Team has improved it. Thanks.

PR #489, opened 2023-08-15
Don't query metrics if require_partition_filter=True.

elongl:


macros/edr/materializations/model/metrics.sql
```diff
@@ -22,8 +22,20 @@
...
 {% endmacro %}
 
-{% macro query_metrics() %}
+{% macro can_query_metrics() %}
   {% if not elementary.get_config_var('collect_metrics') %}
+    {% do return(false) %}
+  {% endif %}
+
+  {% if model.config.require_partition_filter %}
+    {% do return(false) %}
```
elongl, line 31: Wonder if we should send only the `build_timestamp` metric or is it negligible 🤔

noaKurman (approved):

PR #490, opened 2023-08-16
Added supported languages to materializations.

elongl:


macros/edr/materializations/model/incremental.sql
```diff
@@ -20,8 +29,8 @@
   {% do return(relations) %}
 {% endmaterialization %}
 
-{% materialization incremental, adapter="bigquery" %}
-  {% set relations = dbt.materialization_incremental_bigquery() %}
+{% materialization incremental, adapter="bigquery", supported_languages=["sql", "python"] %}
```
elongl, line 32: Currently testing if this would break `dbt<1.3.0` users because `supported_languages` was unknown.

IDoneShaveIt (approved):
LGTM

macros/edr/materializations/model/incremental.sql
elongl: It does break it.

PR #493, opened 2023-08-17
Run CI on both lowest and highest versions of dbt.

elongl:


.github/workflows/test-all-warehouses.yml
```diff
@@ -25,6 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        dbt-version: ["1.3.0", null]
```
elongl, line 28: Schema names will conflict.

PR #504, opened 2023-08-27
Slack notifications for scheduled workflow

noaKurman:


.github/workflows/notify_slack.yml
```diff
@@ -0,0 +1,51 @@
+name: Notify Slack
```
elongl, line 1: Why re-create this?
We can call it from the other repo.

.github/workflows/test-all-warehouses.yml
```diff
@@ -43,3 +43,19 @@ jobs:
...
       dbt-data-reliability-ref: ${{ inputs.dbt-data-reliability-ref }}
     secrets: inherit
+
+  notify_failures:
+    name: Notify Slack
+    secrets: inherit
+    needs: [test]
+    if: |
+      always() &&
+      ! cancelled() &&
```
elongl, line 53: Does it still work when there's a space here?

elongl (approved):

.github/workflows/test-all-warehouses.yml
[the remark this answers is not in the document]
noaKurman: Yes it does

PR #515, opened 2023-08-30
added query history implementation for bigquery

ofek1weiss:


macros/utils/cross_db_utils/get_query_history.sql
```diff
@@ -0,0 +1,31 @@
...
+  referenced_table.project_id as source_db,
+  referenced_table.dataset_id as source_schema,
+  referenced_table.table_id as source_table,
+  destination_table.project_id as dest_db,
+  destination_table.dataset_id as dest_schema,
+  destination_table.table_id as dest_table
+from `{{ target.project }}`.`region-{{ target.location }}`.INFORMATION_SCHEMA.JOBS, 
+  UNNEST(referenced_tables) AS referenced_table
+where
+  {% if from_time %}
```
elongl, line 22: Should we maybe raise an error if this macro is called without `from_time` in order to avoid huge queries?

macros/utils/cross_db_utils/get_query_history.sql
```diff
@@ -0,0 +1,31 @@
+{% macro get_query_history(from_time) %}
```
elongl, line 1: Rename this to `get_query_history_query`.

models/edr/run_results/query_history.sql
```diff
@@ -0,0 +1,19 @@
+{{
```
elongl, line 1: How do we ensure that this model doesn't error for users who don't have permissions to `JOBS`?
Do everyone have the option to query it and at least get partial results (only for their user, etc.)?

macros/utils/cross_db_utils/get_query_history.sql
```diff
@@ -0,0 +1,31 @@
...
+  start_time,
+  end_time,
+  state,
+  referenced_table.project_id as source_db,
+  referenced_table.dataset_id as source_schema,
+  referenced_table.table_id as source_table,
+  destination_table.project_id as dest_db,
+  destination_table.dataset_id as dest_schema,
+  destination_table.table_id as dest_table
+from `{{ target.project }}`.`region-{{ target.location }}`.INFORMATION_SCHEMA.JOBS,
```
elongl, line 19: Why don't you use `.information_schema()` like [here](https://github.com/elementary-data/dbt-data-reliability/blob/618d592ad91ff5ca61dc942036edb8cffa95b0f6/macros/edr/metadata_collection/get_columns_from_information_schema.sql#L22)?

macros/utils/cross_db_utils/get_query_history.sql
ofek1weiss: sure thing

models/edr/run_results/query_history.sql
ofek1weiss: the permission to list jobs is pretty basic in bigquery so i dont think we will encounter it, but i can (and will now) add a `can_query_relation` check

macros/utils/cross_db_utils/get_query_history.sql
ofek1weiss: honestly did not think about it, tho it might not work due to the region as part of the identifier, will check

macros/utils/cross_db_utils/get_query_history.sql
```diff
@@ -0,0 +1,31 @@
...
+
+{% macro bigquery__get_query_history(from_time) %}
+select 
+  job_id || referenced_table.project_id || referenced_table.dataset_id || referenced_table.table_id as id,
+  job_id,
+  statement_type,
+  start_time,
+  end_time,
+  state,
+  referenced_table.project_id as source_db,
```
elongl, line 13: If you named the other `dest`, name this `src_`.

macros/utils/cross_db_utils/get_query_history_query.sql
```diff
@@ -0,0 +1,38 @@
+{% macro get_query_history_query(from_time) %}
+  {% do return(adapter.dispatch("get_query_history_query", "elementary")(from_time)) %}
+{% endmacro %}
+
+
+{% macro bigquery__get_query_history_query(from_time) %}
+{% if not from_time %}
```
elongl, line 7: This `if` should be to the high-level macro.

PR #521, opened 2023-09-05
Update upload_dbt_models.sql

manojkumarfl:
append =True,

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_models.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_models.sql
@@ -2,7 +2,7 @@
     {% set relation = elementary.get_elementary_relation('dbt_models') %}
     {% if execute and relation %}
         {% set models = graph.nodes.values() | selectattr('resource_type', '==', 'model') %}
-        {% do elementary.upload_artifacts_to_table(relation, models, elementary.flatten_model, should_commit=should_commit, metadata_hashes=metadata_hashes) %}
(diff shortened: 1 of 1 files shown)


manojkumarfl:
append=True

ofek1weiss:
Hi @manojkumarfl 
I believe we do not want to use append=True for dbt_models, as it should represent the current status of the models in the project.
Why do you want to change that? is there an issue with that?

manojkumarfl:
Hello.   
In this requirement, we have multiple projects, where  this package "elementary" is used to build audit across all projects.
But unfortunately when we call this package ,  The Audit tables, specifically "DBT_MONITOR"."DBT_MODELS" table is getting Overwritten.   Can we do something about it?

manojkumarfl:
I tried, dispach config, to override upload_dbt_models.sql macro with a custom macro,  but since we are not calling this macro implicitly, and calling the entire package,   we are unable to overwrite.
So ultimately for every elementary package trigger from different projects, the table "DBT_MODELS" audit is getting truncated and loaded

ofek1weiss:
I believe having multiple dbt projects for the same database is not the best practice, and doing something like having separate folders inside a single dbt project and using a `--select` to run each seperatly would be a better approach for you (regardless of elementary, tho it will solve this issue).

If you still have to do it for some reason, then you should try having a separate elementary schema for each project, that way they would not collide.

ellakz:
Hi @manojkumarfl ,
At this point we do not support multiple environments with this setup. In the future we might want to but this will require extensive research on the effects of this on all of our features so we cannot approve this right now. 
I recommend using a different Elementary schema per project.

PR #522, opened 2023-09-05
Ele 1694 bugfixes + tests for metric backfill logic

haritamar:


macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -48,7 +48,7 @@
     'upload_artifacts_method': 'diff',
     'project_name': none,
     'elementary_full_refresh': false,
-    'min_training_set_size': 14,
+    'min_training_set_size': 7,
```
haritamar, line 51: I did this because I noticed that even though `backfill_days` is 2, we only report anomalies for 1 days because the training set size for the day before is 13.
But thought in general that this number should be smaller than `days_back` - so the test can detect anomalies sooner, but will use more data once it is available.

integration_tests/tests/test_anomalies_backfill_logic.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+    assert test_result["status"] == "fail"
+    assert get_row_count_metrics(dbt_project, test_id) == {
+        cur_date: 5 if cur_date < utc_today - timedelta(BACKFILL_DAYS) else 1
+        for cur_date in data_dates
+        if cur_date >= utc_today - timedelta(DAYS_BACK)
+    }
+
+
+def test_longer_backfill_in_case_of_a_gap(dbt_project: DbtProject, test_id: str):
```
haritamar, line 116: From this test and below, I ignored the test results and focused on the metrics which are the main goal of this file

.github/workflows/test-warehouse.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         run: sudo apt-get install python-dev libsasl2-dev gcc
 
       - name: Install dbt
-        run: pip install --pre
+        run: pip install
```
haritamar, line 85: Removing the --pre for now because some of the adapter pre-releases are currently broken, and I want to test on the latest stable version

integration_tests/tests/test_all_columns_anomalies.py
```diff
@@ -95,13 +95,13 @@ def test_all_columns_anomalies_with_where_expression(
 
     params = dict(DBT_TEST_ARGS, where="universe = 'Marvel'")
     test_results = dbt_project.test(
-        test_id, DBT_TEST_NAME, params, multiple_results=True
+        f"{test_id}_marvel", DBT_TEST_NAME, params, multiple_results=True
```
haritamar, line 98: Needed to change the ID, because after the fixes to the backfill it only updates the last 2 dates

integration_tests/tests/dbt_project.py
```diff
@@ -36,7 +44,22 @@ def get_dbt_runner(target: str) -> DbtRunner:
...
 class DbtProject:
     def __init__(self, target: str):
-        self.dbt_runner = get_dbt_runner(target)
+        self.dbt_project_copy_dir = Path(mkdtemp(prefix="integration_tests_project_"))
+        shutil.copytree(
+            DBT_PROJECT_PATH,
+            self.dbt_project_copy_dir,
+            dirs_exist_ok=True,
+            symlinks=True,
+        )
```
haritamar, line 53: Doing this to avoid some races that are happening in the CI
(I believe / suspect the random failures we're seeing are because multiple processes of dbt are running on the same folder and updating the manifest file - so now each test will have its own project folder)

.github/workflows/test-warehouse.yml
```diff
@@ -82,7 +90,8 @@ jobs:
         run: sudo apt-get install python-dev libsasl2-dev gcc
 
       - name: Install dbt
-        run: pip install --pre
+        run:
+          pip install${{ (inputs.include-dbt-pre-releases == 'true' && ' --pre') || '' }}
```
haritamar, line 94: Not running on --pre by default because dbt's pre-releases are currently broken.
Added the option to enable it as a parameter, maybe we should also have a scheduled workflow that runs it (in a lower frequency)

macros/edr/data_monitoring/data_monitors_configuration/get_buckets_configuration.sql
```diff
@@ -64,40 +64,40 @@
...
                 then {{ elementary.edr_cast_as_timestamp(elementary.edr_timeadd(metric_properties.time_bucket.period, 'periods_until_max', 'days_back_start')) }} {# Add full buckets to days_back_start #}
-                else {{ elementary.edr_cast_as_timestamp(elementary.edr_timeadd(metric_properties.time_bucket.period, 'periods_until_max', 'last_max_bucket_end')) }} {# Add full buckets to last_max_bucket_end #}
+                else {{ elementary.edr_cast_as_timestamp(elementary.edr_timeadd(metric_properties.time_bucket.period, 'periods_until_max', 'max_existing_bucket_end')) }} {# Add full buckets to max_existing_bucket_end #}
             end as max_bucket_end
         from full_buckets_calc
     {%- endset %}
 
     {# We assume we should also cosider sources as incremental #}
-    {% if not (elementary.is_incremental_model(elementary.get_model_graph_node(model), source_included=true) or unit_test) %}
+    {% if not (elementary.is_incremental_model(elementary.get_model_graph_node(model_relation), source_included=true) or unit_test) %}
```
haritamar, line 100: This is something that broke a while ago (I think due to my changes), that actually caused us to always return False from `is_incremental_model`

macros/edr/data_monitoring/data_monitors_configuration/get_buckets_configuration.sql
```diff
@@ -47,8 +47,8 @@
 
     {%- set incremental_bucket_times_query %}
         with bucket_times as (
-            select min(last_bucket_end) as last_max_bucket_end,
-                   min(first_bucket_end) as last_min_bucket_end,
+            select max(last_bucket_end) as max_existing_bucket_end,
+                   {{ elementary.edr_cast_as_timestamp(elementary.edr_timeadd(metric_properties.time_bucket.period, -1 * metric_properties.time_bucket.count, 'min(first_bucket_end)')) }} as min_existing_bucket_start,
```
haritamar, line 51: Fixed here and below some inaccurate comparisons (e.g. comparing bucket starts to bucket ends)

macros/edr/system/system_utils/get_config_var.sql
ofek1weiss: just make sure to update the docs, as i think the number 14 is mentioned quite a lot

integration_tests/tests/dbt_project.py
```diff
@@ -36,7 +44,22 @@ def get_dbt_runner(target: str) -> DbtRunner:
 
 class DbtProject:
     def __init__(self, target: str):
-        self.dbt_runner = get_dbt_runner(target)
+        self.dbt_project_copy_dir = Path(mkdtemp(prefix="integration_tests_project_"))
```
ofek1weiss, line 47: im worried it might increase the runtime of the tests substantially, would you mind checking?
in case it does, maybe a copy per process/thread?

.github/workflows/test-warehouse.yml
ofek1weiss: but from what elon told me once, we added this because we want to make sure **we** dont break with the newer versions of dbt

ofek1weiss (approved):

PR #524, opened 2023-09-06
added conditional statement to quote temp_table_name if adapter quoting config is set to False 

trevdoz:
added conditional statement to check if dbt adapter equals 'snowflake' and if so then add quotes around temp_table_name for elementary test table to handle hyphen names in model for snowflake adapter (since default quoting config is set to False for Snowflake according to dbt docs)

### DIFF
--- a/macros/edr/tests/test_utils/create_elementary_test_table.sql
+++ b/macros/edr/tests/test_utils/create_elementary_test_table.sql
@@ -2,7 +2,7 @@
     {% if execute %}
         {% set temp_table_name = elementary.table_name_with_suffix(test_name, "__" ~ table_type ~ elementary.get_timestamped_table_suffix()).replace("*", "") %}
         
-        default_identifier_quoting = api.Relation.get_default_quote_policy().get_part("identifier")        
+        {% set default_identifier_quoting = api.Relation.get_default_quote_policy().get_part("identifier") %}        
         {% if adapter.config.quoting.get("identifier",default_identifier_quoting) %}
             {% set temp_table_name = adapter.quote(temp_table_name) %}
         {% endif %}
--- a/macros/edr/tests/test_utils/create_elementary_test_table.sql
+++ b/macros/edr/tests/test_utils/create_elementary_test_table.sql
@@ -1,6 +1,10 @@
 {% macro create_elementary_test_table(database_name, schema_name, test_name, table_type, sql_query) %}
     {% if execute %}
         {% set temp_table_name = elementary.table_name_with_suffix(test_name, "__" ~ table_type ~ elementary.get_timestamped_table_suffix()).replace("*", "") %}
+        {% if adapter.type() == 'snowflake'}
+            {% set temp_table_name = adapter.quote(temp_table_name) %}
+        {% endif %}
+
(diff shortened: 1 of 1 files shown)


trevdoz:
this is related to issue [1128 ](https://github.com/elementary-data/elementary/issues/1128)

ofek1weiss:
hi @trevdoz 
any reason you made it snowflake specific? i believe it would be beneficial to all adapters

trevdoz:
hey @ofek1weiss,

I was discussing this with @Maayan-s on Slack and since create_elementary_test_table macro uses dbt.get_or_create_relation to create the temp_table_relation, it will use the adapter's default quoting configuration (unless overwritten in dbt_project.yml) when resolving the relation. 

Since Snowflake is the only adapter that dbt by default sets the quoting configuration to false (and highly recommends due to case sensitivity when quoting), I figured Snowflake is the only adapter where using a hypen in the model name becomes an issue.(https://docs.getdbt.com/reference/project-configs/quoting)

Happy to make the change to apply quoting as default regardless of adapter if that is preferred

ofek1weiss:
What scares me a bit is - what if someone changes the config in their project to have it quote the identifier, this will cause this to break right?
i think a safer solution might be to use `api.Relation.get_default_quote_policy().identifier` to see if identifier quoting in enabled by default.
im imagining it like this:
```
{% if not api.Relation.get_default_quote_policy().identifier %}
    {% set temp_table_name = adapter.quote(temp_table_name) %}
{% endif %}
```
what do you think?

trevdoz:
That is definitely more robust. Totally on board with that, I will make the changes

trevdoz:
@ofek1weiss okay implemented adapter-agnostic version

macros/edr/tests/test_utils/create_elementary_test_table.sql
```diff
@@ -1,6 +1,12 @@
 {% macro create_elementary_test_table(database_name, schema_name, test_name, table_type, sql_query) %}
     {% if execute %}
         {% set temp_table_name = elementary.table_name_with_suffix(test_name, "__" ~ table_type ~ elementary.get_timestamped_table_suffix()).replace("*", "") %}
+        
+        {% set default_identifier_quoting = api.Relation.get_default_quote_policy().get_part("identifier") %}        
+        {% if not adapter.config.quoting.get("identifier",default_identifier_quoting) %}
```
ofek1weiss, line 6: @trevdoz  little nitpick:
```suggestion
        {% if not adapter.config.quoting.get("identifier", default_identifier_quoting) %}
```
other then that look great, ill test it and then it can be merged 👍

trevdoz:
@ofek1weiss team work makes the dream work! thanks for your help on this...i notice the tests are failing due to DWH connection. is that due to my end not having a profile.yml created in my branch?

ofek1weiss:
@trevdoz dont worry, the tests currently dont work for contributions and thats ok, just wanted to see that the pre-commits run fine (which they do).
ill now run the tests internally and one they succeed (🤞 ) ill merge this

ofek1weiss (approved):
👍

trevdoz:
You rock! @ofek1weiss

PR #530, opened 2023-09-10
Add generic test for exposure schema validation

erikzaadi:
Added a generic test that will validate model columns and data types pending that the `meta` attribute is populated with a dict in the following schema:

```json
{ 
    "referenced_columns": [
        {
            "column_name": "column_A",
            "data_type": "string"
        },
        {
            "column_name": "column_B",
            "data_type": "numeric",
            "node": "ref('table')"
        } 
    ]
}
```

Where the `data_type` is optional.

`source` is optional if you only have 1 `depends_on`  node.

E.g in a `exposures.yml`:

```yml
exposures:

  - name: customers
    label: CustomersFTW
    type: dashboard
    maturity: high
    url: https://bi.tool/dashboards/1
    description: >
      Did someone say "exponential growth"?

    depends_on:
      - ref('customers')

    owner:
      name: Callum McData
      email: data@jaffleshop.com
    meta:
      referenced_columns:
        - column_name: "customer_id"
          data_type: "numeric"
        - column_name: 'ZOMG'
          data_type: "string"
        - name: 'WithNoDataType'
          node: ref('customers')
```

### DIFF
--- a/.github/workflows/test-warehouse.yml
+++ b/.github/workflows/test-warehouse.yml
@@ -78,6 +78,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.8.17"
+          cache: "pip"
 
       - name: Install Spark requirements
         if: inputs.warehouse-type == 'spark'
--- a/integration_tests/dbt_project/models/exposures.yml
+++ b/integration_tests/dbt_project/models/exposures.yml
@@ -1,0 +1,42 @@
+version: 2
+
+exposures:
+  - name: customers
+    label: CustomersFTW
+    type: dashboard
+    maturity: high
+    url: https://bi.tool/dashboards/1
+    description: >
+      Did someone say "exponential growth"?
+
+    depends_on:
+      - ref('customers')
+
+    owner:
+      name: Callum McData
+      email: data@jaffleshop.com
+    meta:
+      referenced_columns:
+        - column_name: id
+          data_type: numeric
+          node: ref('customers')
+
+  - name: orders
+    label: Returned Orders
+    type: dashboard
+    maturity: high
+    url: https://bi.tool/dashboards/2
+    description: >
+      Did someone say "exponential growth"?
+
+    depends_on:
+      - ref('orders')
+
+    owner:
+      name: Callum McData
+      email: data@jaffleshop.com
+    meta:
+      referenced_columns:
+        - column_name: "order_id"
+          data_type: "string"
+        - column_name: "ZOMG"
--- a/integration_tests/tests/test_exposure_schema_validity.py
+++ b/integration_tests/tests/test_exposure_schema_validity.py
@@ -1,0 +1,192 @@
+from typing import List
+
+import pytest
+from dbt_project import DbtProject
+
+DBT_TEST_NAME = "elementary.exposure_schema_validity"
+
+
+def seed(dbt_project: DbtProject):
+    seed_result = dbt_project.dbt_runner.seed(full_refresh=True)
+    assert seed_result is True
+
+
+def validate_failing_query_output(
+    dbt_dir: str,
+    test_output: str,
+    table_name: str,
+    expected_content_strings: List[str],
+):
+    generated_query = f"target/compiled/elementary_tests/models/schema.yml/elementary_exposure_schema_validity_{table_name}_.sql"
+    assert generated_query in test_output
+    output_file = f"{dbt_dir}/{generated_query}"
+    with open(output_file) as f:
+        generated_query_content = f.read()
+        for content_string_to_validate in expected_content_strings:
+            assert content_string_to_validate in generated_query_content
+
+
+def test_exposure_schema_validity_existing_exposure_yml_invalid(
+    test_id: str, dbt_project: DbtProject
+):
+    seed(dbt_project)
+    run_result = dbt_project.dbt_runner.run(
+        models="orders", full_refresh=True, quiet=True
+    )
+    assert run_result is True
+    test_result, test_output = dbt_project.dbt_runner._run_command(
+        command_args=["test", "-s", "tag:exposure_orders"],
+        log_format="text",
+        capture_output=True,
+        quiet=True,
+        log_output=False,
+    )
+    assert test_result is False
+    validate_failing_query_output(
+        dbt_project.project_dir_path,
+        test_output,
+        "orders",
+        [
+            "different data type for the column order_id string vs",
+            "ZOMG column missing in the model",
+        ],
+    )
+
+
+def test_exposure_schema_validity_existing_exposure_yml_valid(
+    test_id: str, dbt_project: DbtProject
+):
+    seed(dbt_project)
+    run_result = dbt_project.dbt_runner.run(
+        models="customers", full_refresh=True, quiet=True
+    )
+    assert run_result is True
+    test_result, test_output = dbt_project.dbt_runner._run_command(
+        command_args=["test", "-s", "tag:exposure_customers"],
+        capture_output=True,
+        quiet=True,
+        log_output=False,
+    )
+    assert test_result is True
+
+
+@pytest.mark.skip_targets(["spark"])
+def test_exposure_schema_validity_no_exposures(test_id: str, dbt_project: DbtProject):
+    test_result = dbt_project.test(test_id, DBT_TEST_NAME)
+    assert test_result["status"] == "pass"
+
+
+@pytest.mark.skip_targets(["spark"])
+def test_exposure_schema_validity_correct_columns_and_types(
+    test_id: str, dbt_project: DbtProject
+):
+    explicit_target_for_bigquery = (
+        "other"
+        if dbt_project.dbt_runner.target in ["bigquery", "snowflake", ""]
+        else "string"
+    )
+    DBT_TEST_ARGS = {
+        "node": "models.exposures_test",
+        "columns": [{"name": "order_id", "dtype": "string"}],
+        "exposures": {
+            "ZOMG": {
+                "meta": {
+                    "referenced_columns": [
+                        {
+                            "column_name": "order_id",
+                            "data_type": explicit_target_for_bigquery,
+                        }
+                    ]
+                },
+                "url": "http://bla.com",
+                "name": "ZOMG",
+                "depends_on": {"nodes": ["models.exposures_test"]},
+            }
+        },
+    }
+    test_result = dbt_project.test(
+        test_id, DBT_TEST_NAME, DBT_TEST_ARGS, columns=[dict(name="bla")], as_model=True
+    )
+    assert test_result["status"] == "pass"
+
+
+@pytest.mark.skip_targets(["spark"])
+def test_exposure_schema_validity_correct_columns_and_invalid_type(
+    test_id: str, dbt_project: DbtProject
+):
+    DBT_TEST_ARGS = {
+        "node": "models.exposures_test",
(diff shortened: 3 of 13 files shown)


ellakz:
@erikzaadi for clarity's sake I'd use "data_type" and not dtype :)
also can you please add tests and documentation?

macros/edr/tests/test_exposure_schema_validity.sql
```diff
@@ -0,0 +1,58 @@
+{% test exposure_schema_validity(model, exposures, node, columns) %}
```
haritamar, line 1: Maybe add a comment that `exposures`, `nodes` and `columns` exist as parameters only for dependency injection from the integration tests (right?)

integration_tests/tests/test_exposure_schema_validity.py
```diff
@@ -0,0 +1,161 @@
...
+                "dtype": "numeric"
+            }
+        ],
+        "exposures": {
+            "ZOMG": {
+                "meta": {
+                    "columns": [{
+                        "name": "order_id",
+                        "data_type": "string"
+                    }]
```
haritamar, line 140: I'd discuss also with @ellakz 
but I think each column specified in the exposure should say which model it is from - otherwise I find it confusing if the exposure depends on multiple models.

macros/edr/tests/test_exposure_schema_validity.sql
```diff
@@ -0,0 +1,58 @@
...
+                {%- endfor -%}
+            {%- endif -%}
+        {%- endfor -%}
+        {%- for invalid_exposure in invalid_exposures %}
+            {{ 'UNION ALL ' if not loop.first }}SELECT '{{ invalid_exposure['exposure'] }}' as exposure, '{{ invalid_exposure['url'] }}' as url, '{{ invalid_exposure['error'] }}' as error
+        {%- endfor -%}
+        {{ elementary.test_log('end', full_table_name, 'exposure validation') }}
+        {% if invalid_exposures | length == 0 %}
+            {{ elementary.no_results_query() }}
+        {% endif %}
```
haritamar, line 54: I think it may be more readable to write:
```
{% if invalid_exposures | length > 0 %}
    {%- for invalid_exposure in invalid_exposures %}
        {{ 'UNION ALL ' if not loop.first }}SELECT '{{ invalid_exposure['exposure'] }}' as exposure, '{{invalid_exposure['url'] }}' as url, '{{ invalid_exposure['error'] }}' as error
    {%- endfor -%}
{% else %}
    {{ elementary.no_results_query() }}
{% endif %}
```

macros/edr/tests/test_exposure_schema_validity.sql
haritamar, line 1: Can you add the new test to `common_test_configs.sql`?
(and maybe also to `get_test_type.sql` though I'm actually not sure it's needed there)

integration_tests/tests/test_exposure_schema_validity.py
```diff
@@ -0,0 +1,161 @@
+from datetime import date, timedelta
+from typing import Any, Dict, List
+
+from data_generator import DATE_FORMAT, generate_dates
+from dbt_project import DbtProject
+
+DBT_TEST_NAME = "elementary.exposure_schema_validity"
+
```
haritamar, line 8: Can we maybe add one test that relies on "real" exposures (rather than dependency injection)?
My main concern is that the format of exposures in the graph will change between dbt versions and the test will miss it.

I think you can just commit models / exposures that you need to the test dbt project.

macros/edr/tests/test_exposure_schema_validity.sql
erikzaadi: needed it in the `get_test_type.sql` (https://github.com/elementary-data/dbt-data-reliability/pull/530/files#diff-7767fb1ef194c47d2d634293a374325976c7d9383d80c7610d8e82637617bf00R33), will add to `common_test_configs.sql` as well

haritamar: ah sorry missed it, cool

elongl:
Personally think `columns` needs to be renamed to `referenced_columns`.

integration_tests/tests/test_exposure_schema_validity.py
```diff
@@ -0,0 +1,108 @@
...
+            }
+        },
+    }
+    test_result = dbt_project.test(
+        test_id, DBT_TEST_NAME, DBT_TEST_ARGS, columns=[dict(name="bla")], as_model=True
+    )
+    assert test_result["status"] == "pass"
+
+
+def test_expose_schema_validity_with_correct_columns_and_invalid_type(
```
elongl, line 34:
```suggestion
def test_exposure_schema_validity_with_correct_columns_and_invalid_type(
```

macros/edr/tests/test_exposure_schema_validity.sql
```diff
@@ -0,0 +1,64 @@
+{% test exposure_schema_validity(model, exposures, node, columns) %}
+    {%- if not execute -%}
+        {%- do return(none) -%}
+    {%- endif -%}
+
+    {# Parameters used only for dependency injection in integration tests #}
+    {%- set node = node | default(context['model']['attached_node']) -%}
```
elongl, line 7: Can you replace the `default()` to `or` operators?
Also, when accessing a dict value, you can use `.get(key, default)`.

macros/edr/tests/test_exposure_schema_validity.sql
```diff
@@ -0,0 +1,64 @@
...
+
+    {%- set model_relation = elementary.get_model_relation_for_test(model, context["model"]) -%}
+    {%- set full_table_name = elementary.relation_to_full_name(model_relation) -%}
+    {{- elementary.test_log('start', full_table_name, 'exposure validation') -}}
+
+    {%- set matching_exposures = [] -%}
+
+    {%- for exposure in exposures -%}
+        {%- if node in exposure.depends_on.nodes and exposure['meta'] | default(none) is not none -%}
+            {%- do matching_exposures.append(exposure) -%}
```
elongl, line 19: I'd expect the `matching_exposures` to only be exposures that are referencing the tested model in `depends_on`. This would also save a lot of unnecessary rendering when there are a lot of exposures.

macros/edr/tests/test_exposure_schema_validity.sql
```diff
@@ -0,0 +1,64 @@
...
+                                    })
+                            -%}
+                        {%- endif -%}
+                    {%- endif -%}
+                {%- endfor -%}
+            {%- endif -%}
+        {%- endfor -%}
+        {%- if invalid_exposures | length > 0 -%}
+            {%- for invalid_exposure in invalid_exposures %}
+                {{ 'UNION ALL ' if not loop.first }}SELECT '{{ invalid_exposure['exposure'] }}' as exposure, '{{ invalid_exposure['url'] }}' as url, '{{ invalid_exposure['error'] }}' as error
```
elongl, line 55: I think it's better to use [`exceptions.raise_compiler_error`](https://docs.getdbt.com/reference/dbt-jinja-functions/exceptions#raise_compiler_error) because query results are not presented while exceptions are.

macros/edr/tests/test_exposure_schema_validity.sql
```diff
@@ -0,0 +1,64 @@
...
+    {%- if matching_exposures | length > 0 -%}
+        {%- set columns_dict = {} -%}
+        {%- for column in columns -%}
+            {%- do columns_dict.update({ column['name'] | upper : elementary.normalize_data_type(column['dtype']) }) -%}
+        {%- endfor -%}
+        {%- set invalid_exposures = [] -%}
+        {%- for exposure in matching_exposures -%}
+            {# Depend on meta since column level info is not available on exposures #}
+            {%- set meta = exposure['meta'] | default(none) -%}
+            {%- if meta != none and meta['columns'] | default(none) is iterable -%}
```
elongl, line 31: I think `referenced_columns` is clearer for the user.

integration_tests/tests/test_exposure_schema_validity.py
elongl: This also exists on the other tests.

macros/edr/tests/test_exposure_schema_validity.sql
erikzaadi: As the property name in the `meta` property?

macros/edr/tests/test_exposure_schema_validity.sql
erikzaadi: replaced to `or`

macros/edr/tests/test_exposure_schema_validity.sql
erikzaadi: Done

macros/edr/tests/test_exposure_schema_validity.sql
erikzaadi: That's what this loop does

.github/workflows/test-warehouse.yml
```diff
@@ -78,6 +78,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.8.17"
+          cache: "pip"
```
haritamar, line 81: Cool
Did you see performance improvements in the job?
I tried adding it once to another workflow and it didn't save as much as I hoped, but still I guess it's a good configuration

integration_tests/dbt_project/models/exposures.yml
```diff
@@ -0,0 +1,42 @@
...
+      - ref('customers')
+
+    owner:
+      name: Callum McData
+      email: data@jaffleshop.com
+    meta:
+      referenced_columns:
+        - name: id
+          data_type: numeric
+          source: ref('customers')
```
haritamar, line 22: I think that maybe instead of "source" we should call it "node"?
Since it can contain either models or sources.

macros/edr/tests/test_utils/get_test_type.sql
```diff
@@ -37,6 +40,8 @@
       {% do return("schema_change") %}
     {% elif flattened_test.short_name | lower in python_tests %}
         {% do return("python_test") %}
+    {% elif flattened_test.short_name | lower in dbt_tests %}
+        {% do return("dbt_test") %}
```
IDoneShaveIt, line 44: I am not sure this is the right thing to do.
Maybe the definition of test type is a bit off, but I do not think that we need to categorize ore none anomaly / schema / python tests as `dbt test`.

Maybe we need to put it in a category like 'integrity` and start having a better types because dbt test doesn't says much and can be really confusing.

@haritamar WDYT?

macros/edr/tests/test_exposure_schema_validity.sql
IDoneShaveIt, line 1: Can we add warnings foe the user in case the exposures he set don't contain the right meta format (or meta at all)?
I think that saying "Your test passed - your exposures are valid" while we didn't checked them because they have missing properties or something.

IDoneShaveIt: Not sure if that is me being picky, but I think that we should had comments that explains the test behaviour as well.
Although it is a nice test, it is still hard to understand how it works just from reading it (for example, I didn't understand by myself that we need to make sure the exposure contains the right meta fields)

IDoneShaveIt (commented):
Overall really nice job!
Left some comments - nothing to serious 🙂

IDoneShaveIt (approved):
LGTM

PR #539, opened 2023-09-19
Added 'partition_by_field' to 'dbt_models'.

elongl:


macros/edr/dbt_artifacts/upload_dbt_models.sql
```diff
@@ -71,8 +77,9 @@
         'package_name': node_dict.get('package_name'),
         'original_path': node_dict.get('original_file_path'),
         'path': node_dict.get('path'),
-        'generated_at': elementary.datetime_now_utc_as_string()
-    }%}
+        'generated_at': elementary.datetime_now_utc_as_string(),
+        'bigquery_partition_by_field': config_dict.get("partition_by", {}).get("field"),
```
haritamar, line 81: Maybe get the entire "partition_by" json?
I think the type and granularity may also be useful

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -62,7 +62,8 @@
     'ignore_small_changes': {
       'spike_failure_percent_threshold': none,
       'drop_failure_percent_threshold': none
-    }
+    },
+    'include_other_warehouse_specific_columns': false,
```
ofek1weiss, line 66: why not just `include_warehouse_specific_columns`? why the `other`?

macros/edr/dbt_artifacts/upload_dbt_models.sql
```diff
@@ -71,8 +77,9 @@
         'package_name': node_dict.get('package_name'),
         'original_path': node_dict.get('original_file_path'),
         'path': node_dict.get('path'),
-        'generated_at': elementary.datetime_now_utc_as_string()
-    }%}
+        'generated_at': elementary.datetime_now_utc_as_string(),
+        'bigquery_partition_by': config_dict.get("partition_by"),
```
ofek1weiss, line 81: just a thought, maybe we can stick with the `dbt` way of doing double underscore between adapter name and name, (`bigquery__partition_by`)

elongl: Because for BigQuery users, I'll add BigQuery fields and this `var` _does not_ affect it.
It does affect if I'm a BigQuery user and whether it'll add Redshift fields.

macros/edr/system/system_utils/get_config_var.sql
elongl: Because for BigQuery users, I'll add BigQuery fields and this `var` does not affect it.
It does affect if I'm a BigQuery user and whether it'll add Redshift fields.

haritamar (approved):

PR #547, opened 2023-09-25
Ele 1773 package tests fixes

haritamar:


integration_tests/tests/test_anomalies_backfill_logic.py
```diff
@@ -189,6 +191,7 @@ def test_full_backfill_if_metric_not_updated_for_a_long_time(
             ).isoformat()
         },
     )
+    assert test_result["status"] != "error"
```
haritamar, line 194: These are tests where I don't care about success / failure but about the metric values - but still wanted to add a check it didn't result in an error.

erikzaadi (approved):

PR #549, opened 2023-09-26
Performace optimization

ofek1weiss:


macros/edr/materializations/test/test.sql
```diff
@@ -12,6 +12,10 @@
   {% if sample_limit == 0 %} {# performance: no need to run a sql query that we know returns an empty list #}
     {% do return([]) %}
   {% endif %}
+  {% if elementary.did_test_pass() %}
```
elongl, line 15: For some of Elementary's tests such as `volume_anomalies` and `freshness_anomalies`, we sample the test even on success.

elongl (approved):

PR #560, opened 2023-10-01
changed so test will run in the correct time

ofek1weiss:


macros/edr/materializations/test/test.sql
elongl, line 1: Can you please rename `original_materialization` to `materialization_macro` or `materialize_macro`?

elongl (approved):

PR #582, opened 2023-10-07
anomalies: add support for anomaly_exclude_filter

haritamar:
This PR adds a capability to exclude metrics from the anomaly detection by an SQL expression. 
This is similar to a "where" expression - but will exclude completely certain buckets from the computation (rather than, for example, resulting in a zero metric for volume).

The "where" expression supports the following fields:
1. `metric_date` - the date of the relevant bucket (even if the bucket is not daily).
2. `metric_time_bucket` - the exact time bucket
3. `metric_value` - the value of the metric

Configuration examples
```yaml
# Exclude specific date
anomaly_exclude_metrics: metric_date = '2023-10-01'

# Exclude dates below a threshold
anomaly_exclude_metrics: metric_date <= '2023-10-01'

# Exclude exact time buckets
anomaly_exclude_metrics: metric_time_bucket >= '2023-10-01 06:00:00' and metric_time_bucket <= '2023-10-01 07:00:00'

# Exclude metrics by metric value
anomaly_exclude_metrics: metric_value < 10
```


macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -141,6 +148,7 @@
                 last_value(bucket_end) over (partition by metric_name, full_table_name, column_name, dimension, dimension_value, bucket_seasonality order by bucket_end asc rows between unbounded preceding and current row) training_end,
                 first_value(bucket_end) over (partition by metric_name, full_table_name, column_name, dimension, dimension_value, bucket_seasonality order by bucket_end asc rows between unbounded preceding and current row) as training_start
             from grouped_metrics
+            where not is_excluded
```
haritamar, line 151: Right now we simply exclude the "is_excluded" rows
but ideally later the goal is to return them, and grey them out in the UI or something like that (that's why I made it a field)

macros/utils/dict_utils/undefined_dict_keys_to_none.sql
```diff
@@ -1,6 +1,6 @@
-{% macro empty_dict_keys_to_none(dict) %}
+{% macro undefined_dict_keys_to_none(dict) %}
```
haritamar, line 1: We only want to turn undefined -> none
but not other empty values (empty strings, empty lists, etc)

macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -31,6 +31,8 @@
     {%- set detection_end = elementary.get_detection_end(test_configuration.detection_delay) %}
     {%- set min_bucket_start_expr = elementary.get_trunc_min_bucket_start_expr(detection_end, metric_properties, test_configuration.days_back) %}
 
+    {%- set metric_time_bucket_expr = 'case when bucket_start is not null then bucket_start else bucket_end end' %}
```
haritamar, line 34: I added this because I think that for timestamped tests it's better to use `bucket_start`, but for non-timestamped tests we currently only fill `bucket_end`
(This is specifically for the interface of `anomaly_exclude_metrics` as explained below)

erikzaadi (approved):

.github/workflows/test-warehouse.yml
```diff
@@ -115,7 +115,7 @@ jobs:
 
       - name: Test
         working-directory: "${{ env.TESTS_DIR }}/tests"
-        run: py.test -n8 -vvv --target "${{ inputs.warehouse-type }}" --junit-xml=test-results.xml --html=detailed_report.html --self-contained-html
+        run: py.test -n8 -vvv --target "${{ inputs.warehouse-type }}" --junit-xml=test-results.xml --html=detailed_report_${{ inputs.warehouse-type }}_dbt_${{ inputs.dbt-version }}.html --self-contained-html
```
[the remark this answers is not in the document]
haritamar: yeah minor optimization - the file was always called `detailed_report.html` for all targets and it was confusing when I downloaded it :)
so gave it the same name as the artifact
e.g. `detailed_report_snowflake_dbt_1.6.0`

macros/utils/data_types/data_type.sql
```diff
@@ -123,9 +123,5 @@
 {% endmacro %}
 
 {% macro default__edr_type_date() %}
```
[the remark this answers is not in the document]
haritamar: actually yes - I've noticed we're not handling dates correctly
I think we just didn't care about it, and Maayan needed to handle an edge case for BQ.
But I checked across all the warehouses we support and the correct date type for all of them is `date`.

I changed it because otherwise it caused filters like `metric_date in ('2023-11-01', '2023-11-03')` (which I added in my tests) not to work because the code cast them to a timestamp

PR #586, opened 2023-10-08
ELE-1873: implement stubs in the package for getting test results

dapollak:


macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -63,7 +63,8 @@
       'drop_failure_percent_threshold': none
     },
     'include_other_warehouse_specific_columns': false,
-    'fail_on_zero': false
+    'fail_on_zero': false,
+    'dismiss_on_run_end': false
```
haritamar, line 67: Maybe call this `disable_on_run_end` just to be consistent with other flags

macros/utils/graph/get_elementary_relation.sql
```diff
@@ -1,7 +1,15 @@
 {% macro get_elementary_relation(identifier) %}
+  {% do return(adapter.dispatch("get_elementary_relation", "elementary")(identifier)) %}
```
haritamar, line 2: Why did you need to add `adapter.dispatch`?
Do you think this method will be different for different adapters?

macros/edr/system/system_utils/get_config_var.sql
haritamar: Also - maybe call it `disable_elementary_on_run_end` (a bit clearer when referenced from a different project)

macros/utils/graph/get_elementary_relation.sql
dapollak: As I understand, in order to be able to use [this](https://docs.getdbt.com/reference/project-configs/dispatch-config) to override a macro, it must be called through the adapter object.

PR #587, opened 2023-10-09
Changed the columns models.

elongl:
`dbt_columns` now only contains columns that are documented in the dbt project.
`information_schema_columns` contains all the columns from all the schemas that your dbt project addresses.


macros/edr/dbt_artifacts/upload_dbt_columns.sql
```diff
@@ -0,0 +1,67 @@
...
+        ('generated_at', 'string'),
+        ('metadata_hash', 'string'),
+    ] %}
+    {% set dbt_columns_empty_table_query = elementary.empty_table(columns) %}
+    {{ return(dbt_columns_empty_table_query) }}
+{% endmacro %}
+
+{% macro flatten_table_columns(table_node) %}
+    {% set column_nodes = table_node.get("columns") %}
+    {% if not column_nodes %}
```
elongl, line 31: Should we not include columns that don't have descriptions for optimization's-sake?

models/edr/dbt_artifacts/dbt_columns.sql
elongl, line 1: There should probably be a combined view model of `dbt_columns` and `information_schema_columns`.

macros/edr/dbt_artifacts/upload_dbt_columns.sql
```diff
@@ -0,0 +1,67 @@
...
+    {% set config_dict = elementary.safe_get_with_default(column_node, 'config', {}) %}
+    {% set config_meta_dict = elementary.safe_get_with_default(config_dict, 'meta', {}) %}
+    {% set meta_dict = elementary.safe_get_with_default(column_node, 'meta', {}) %}
+    {% do meta_dict.update(config_meta_dict) %}
+    {% set config_tags = elementary.safe_get_with_default(config_dict, 'tags', []) %}
+    {% set global_tags = elementary.safe_get_with_default(column_node, 'tags', []) %}
+    {% set meta_tags = elementary.safe_get_with_default(meta_dict, 'tags', []) %}
+    {% set tags = elementary.union_lists(config_tags, global_tags) %}
+    {% set tags = elementary.union_lists(tags, meta_tags) %}
+    {% set flatten_column_metadata_dict = {
```
elongl, line 52: Should we add `resource_type`?

macros/edr/dbt_artifacts/upload_dbt_columns.sql
```diff
@@ -0,0 +1,67 @@
...
+        'parent_unique_id': table_node.get('unique_id'),
+        'unique_id': "column.{}.{}".format(table_node.get('unique_id'), column_node.get('name')),
+        'name': column_node.get('name'),
+        'data_type': column_node.get('data_type'),
+        'tags': elementary.filter_none_and_sort(tags),
+        'meta': meta_dict,
+        'description': column_node.get('description') or none,
+        'database_name': table_node.get('database'),
```
elongl, line 60: Should we remove the relation's properties given that we have them in other artifacts?
This could supposably speed up the artifacts caching, but would also require joining the other artifacts to find the relation of a column.

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -55,7 +55,7 @@
     'calculate_failed_count': true,
     'tests_use_temp_tables': false,
     'collect_metrics': true,
-    'upload_dbt_columns': false,
+    'upload_information_schema_columns': false,
```
elongl, line 58: Breaking change.

models/edr/dbt_artifacts/information_schema_columns.sql
elongl, line 1: Is this a good name for the model?

models/edr/dbt_artifacts/dbt_columns.sql
elongl, line 1: Renaming the model is also a breaking change at itself.

haritamar: I think that will be good, yeah
see my comment on the test recommendations in the CLI

haritamar: It may be good to bump another minor here, since this is a bit more sensitive.

models/edr/dbt_artifacts/information_schema_columns.sql
haritamar: hmm, the only thing that bothers me with it is that it is still going to contain data only for the dbt tables, and not the entire information schema.
But maybe it's fine.

macros/edr/system/system_utils/get_config_var.sql
haritamar: Yeah I think it's fine, given that it's off by default + only relevant for BQ

macros/edr/dbt_artifacts/upload_dbt_columns.sql
haritamar: I'm in favor of including all of them unless we see it causes a performance issue

macros/edr/dbt_artifacts/upload_dbt_columns.sql
haritamar: Can help with joins, no?
Sounds like a good idea

macros/edr/dbt_artifacts/upload_dbt_columns.sql
haritamar: How significant is it in terms of size? It's relatively small fields no?
I think it can be nice to keep them if possible

models/edr/dbt_artifacts/dbt_columns.sql
elongl: Yeah, I agree, will do.

elongl: Created [here](https://github.com/elementary-data/dbt-data-reliability/pull/587/commits/cd34e431351d1a453d7fffa719d4c6ce3432b3b9). What do you think?

models/edr/dbt_artifacts/information_schema_columns.sql
haritamar: Hmm I agree though I think actually that the name `dbt_schemas_columns` is even more confusing :)
So maybe it can stay `information_schema_columns`, since the difference from the other model is clearer.

models/edr/dbt_artifacts/columns.sql
```diff
@@ -0,0 +1,38 @@
...
+
+select
+    database_name,
+    schema_name,
+    table_name,
+    name,
+    data_type,
+    description
+from information_schema_columns
+left join dbt_columns using (database_name, schema_name, table_name, name)
```
haritamar, line 38: I think that maybe it will be good to do `full join`, so if there are columns that only appear in `dbt_columns` they will also be included.
WDYT?

(I actually didn't know until 5 minutes ago that full join is a thing hh, wonder if it works as I think it does)

models/edr/dbt_artifacts/dbt_columns.sql
haritamar: Added one comment there

models/edr/dbt_artifacts/columns.sql
ellakz: dicussed f2f and decided we shouldn't do it (we don't want to show partial columns for tables that are not fully documented)

haritamar (approved):

PR #588, opened 2023-10-10
Change the severity if length of INSERT exceeds `query_max_size`

yu-iskw:
## Overview

We can change the severity, if the length of INSERT statements generated by `get_insert_rows_queries` exceeds `query_max_size`. We can also refactor the macro to handle the limitation of the length an INSERT statement.

First, the current implementation might have a pitfall if the length of `new_insert_query` after adding `row_sql` exceeds `query_max_size` due to the large `row_sql`. We may be able to avoid the issue by increasing the value of `query_max_size`. However, there are some warehouses which have the limitations on the length of an INSERT statement as BigQuery.  Those can be the length as string or the number of elements in the `VALUES` clause. 

Second, if the length of an INSERT statement is large, a warehouse might raise an issue. So, it might be ok for elementary to show warning messages which can be clues to inspect what happens. I think elementary just focus on how to chunk values to an INSERT statement.

https://github.com/elementary-data/dbt-data-reliability/blob/ea690c42601f340ea8d927afff0e2d82d688e6b1/macros/utils/table_operations/insert_rows.sql#L55-L87

### DIFF
--- a/macros/utils/table_operations/insert_rows.sql
+++ b/macros/utils/table_operations/insert_rows.sql
@@ -70,7 +70,7 @@
 
           {% if new_insert_query | length > query_max_size %}
             {% do elementary.file_log("Oversized row for insert_rows: {}".format(query_with_row)) %}
-            {% do exceptions.raise_compiler_error("Row to be inserted exceeds var('query_max_size'). Consider increasing its value.") %}
+            {% do exceptions.warn("Row to be inserted exceeds var('query_max_size'). Consider increasing its value.") %}
           {% endif %}
 
           {% if current_query.data != base_insert_query %}


yu-iskw:
@elongl I'm sorry for bothering you, but I would like to get your thought on it. The issue frequently blocks us to insert dbt results to the elementary's table.

elongl:
Hi @yu-iskw, thanks a lot for raising this issue.
May I ask, why don't you raise the `query_max_size` if that's the case?
Also, do you know which table is being inserted into when this error raises?
I'm curious as to where there's so much data on a single row.

Regardless, this does make sense to me so I'll discuss this with the team and update shortly.

ellakz (approved):

yu-iskw:
@elongl Good question. As we set the smaller value to `query_max_size` due to the quota of BigQuery, it can be failed to insert a large number of rows with a larger `query_max_size`. 

https://github.com/elementary-data/dbt-data-reliability/blob/ddb5b41332ec151d56e9f554e9b00fdf3b583198/macros/edr/system/system_utils/get_config_var.sql#L75-L79

PR #592, opened 2023-10-18
ELE-1906 - Fix schema changes upgrade false positive

IDoneShaveIt:
On BigQuery use column relation data_type instead of dtype


erikzaadi (approved):
You might need to fix the exposure schema validation test (IIRC it checks the dtypes there as well)

Besides that:

PR #600, opened 2023-11-01
Ele 1946 dbt columns flags

ellakz:


macros/edr/dbt_artifacts/upload_dbt_artifacts.sql
```diff
@@ -8,7 +8,6 @@
       "dbt_metrics": elementary.upload_dbt_metrics,
       "dbt_exposures": elementary.upload_dbt_exposures,
       "dbt_seeds": elementary.upload_dbt_seeds,
-      "dbt_columns": elementary.upload_dbt_columns,
```
haritamar, line 11: Hmm I think it's problematic to remove it from here - it means `dbt_columns` won't be cached.
Can you add the `disable_dbt_columns_autoupload` logic here instead?

macros/edr/dbt_artifacts/upload_dbt_columns.sql
```diff
@@ -35,7 +35,10 @@
 
     {% set flattened_columns = [] %}
     {% for column_node in column_nodes.values() %}
-        {% do flattened_columns.append(elementary.flatten_column(table_node, column_node)) %}
+        {% set flat_column = elementary.flatten_column(table_node, column_node) %}
+        {% if (not elementary.get_config_var('upload_only_columns_with_descriptions')) or (flat_column['description'] != none and flat_column['description'] != '') %}
```
haritamar, line 39: Maybe on the right it can just be `or flat_column['description']`?

ellakz: yeah wasn't sure if it works like in python

haritamar (approved):

PR #605, opened 2023-11-06
get_profile_creation_query: add query history permissions to redshift

haritamar:


macros/utils/cross_db_utils/get_profile_creation_query.sql
```diff
@@ -17,8 +17,26 @@ GRANT SELECT ON FUTURE VIEWS IN SCHEMA {{ parameters["database"] }}.{{ parameter
...
+-- any user)
+CREATE USER {{ parameters["user"] }} WITH PASSWORD '{{ parameters["password"] }}' SYSLOG ACCESS UNRESTRICTED;
+
+-- Grant read access to the Elementary schema
+GRANT USAGE ON SCHEMA {{ parameters["schema"] }} TO {{ parameters["user"] }};
+GRANT SELECT ON ALL TABLES IN SCHEMA {{ parameters["schema"] }} TO {{ parameters["user"] }};
+ALTER DEFAULT PRIVILEGES IN SCHEMA {{ parameters["schema"] }} GRANT SELECT ON TABLES TO {{ parameters["user"] }};
+
+-- Grant metadata access to tables in the warehouse
+GRANT SELECT ON svv_table_info to {{ parameters["user"] }};
```
haritamar, line 31: This table is actually one of the few system tables in Redshift where you have to explicitly give access (for others you can always select and you can only see what you get access to).

It looks like it has interesting information so I added it (even though it's not related to query history)

PR #609, opened 2023-11-08
get_test_buckets_min_and_max fixes

haritamar:


ellakz (approved):
looks good! this logic is much more readable :)

integration_tests/tests/test_anomalies_backfill_logic.py
```diff
@@ -266,3 +266,68 @@ def test_backfill_when_metric_doesnt_exist_back_enough(
...
+        DBT_TEST_NAME,
+        DBT_TEST_ARGS,
+        data=data,
+        as_model=True,
+        materialization="incremental",
+        test_vars={"days_back": (utc_today - date_gap_end).days - 1},
+    )
+    assert test_result["status"] != "error"
+
+    assert get_row_count_metrics(dbt_project, test_id) == {
```
ellakz, line 309: so in this test the data is not filled because the days_back doesn't reach the gap, so the gap stays, and on the other test it is filled because the days_back is bigger? I think it can be useful to add a comment here :)

PR #611, opened 2023-11-11
Ele 1979 add athena to package tests in ci

haritamar:


macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -70,10 +70,11 @@
...
                     -- a valid one (this check is important for buckets that are not aligned with a day).
-                    -- Note - using concat because BQ can't handle IN with multiple values.
-                    and {{ elementary.edr_concat('bucket_start', 'bucket_end') }} in (
-                        select {{ elementary.edr_concat('edr_bucket_start', 'edr_bucket_end') }} from buckets
-                    )
+                    and {{ elementary.edr_multi_value_in(
+                        [elementary.edr_cast_as_timestamp('bucket_start'), elementary.edr_cast_as_timestamp('bucket_end')],
+                        ['edr_bucket_start', 'edr_bucket_end'],
+                        'buckets'
+                    ) }}
```
haritamar, line 77: The original check didn't work for Athena so I did a fix here.
Decided to also separate the "multi-value in" logic to its own macro, and only do the CONCAT in BQ since it's only needed there.

ofek1weiss (approved):

PR #613, opened 2023-11-20
Optimize 'table_monitoring_query' Macro by Selecting Specific Columns

pei0804:
- https://github.com/elementary-data/elementary/issues/1286

### DIFF
--- a/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
@@ -4,7 +4,7 @@
     {%- set timestamp_column = metric_properties.timestamp_column %}
 
     with monitored_table as (
-        select * from {{ monitored_table }}
+        select timestamp_column from {{ monitored_table }}
         {% if metric_properties.where_expression %} where {{ metric_properties.where_expression }} {% endif %}
     ),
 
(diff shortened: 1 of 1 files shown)


pei0804:
I run a local project by branch code.

```
12:04:21  1 of 1 START test elementary_freshness_anomalies_hoge_model_2__14__3__hour__1___load_at  [RUN]
12:04:22  Elementary: Started running data monitors on:  HOGE_MDOEL
12:04:31  Elementary: Finished running data monitors on: HOGE_MDOEL
12:04:33  1 of 1 PASS elementary_freshness_anomalies_hoge_model_2__14__3__hour__1___load_at  [PASS in 12.91s]
```

pei0804:
Maybe I was mistaken.

pei0804:
A little more verification.

pei0804:
We have reproduced a case where the query was slow.
This problem is reproduced when the size of each record is large and the number of records is large.

For small data, the problem is not reproduced.

pei0804:
Snowflake Profile before tuning.
This query was expensive, so I stopped in the middle of the process.
Query_-_01b0749b-0001-265f-0000-87ed18aa11ce

Snowflake Profile after tuning.
Query_-_01b07498-0001-265f-0000-87ed18a9db52

it is wiser to squeeze in useless scans.

pei0804:
I haven't check any impact on other models.

pei0804:
If this model needs to require an arbitrary column, close PR.
I'm finding another test choice.

pei0804:
I give up on this change because it affects other tests.

PR #619, opened 2023-11-27
Use google cloud function for inserts

jordanwillifordcruise:
Added support for handle_tests_results and insert_rows to use a new insert_rows_method method called gcp-cloud-function which calls a UDF to push results into BigQuery via Pub/Sub rather than direct insert queries.  This significantly increases Elementary's capacity to insert records/test results when using BigQuery.

### DIFF
--- a/macros/edr/system/hooks/on_run_end.sql
+++ b/macros/edr/system/hooks/on_run_end.sql
@@ -5,12 +5,10 @@
         {% do return("") %}
       {% endif %}
 
-      {# comment
       {% if flags.WHICH in ['run', 'build'] %}
         {% do elementary.insert_metrics() %}
       {% endif %}
-      #}
-
+      
       {# Update metadata for all dbt artifacts. #}
       {% if not elementary.get_config_var('disable_dbt_artifacts_autoupload') %}
         {% do elementary.upload_dbt_artifacts() %}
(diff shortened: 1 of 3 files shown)


haritamar:
Hi @jordanwillifordcruise,
This looks like a great contribution!
Do you mind sharing an example of how it's configured? In particular curious what are the expected values of `insert_rows_udf` and `insert_rows_topics`.

jordanwillifordcruise:
> Hi @jordanwillifordcruise, This looks like a great contribution! Do you mind sharing an example of how it's configured? In particular curious what are the expected values of the `insert_rows_udf` and `insert_rows_topics` vars.

Here is an example configuration.  You identify a UDF that publishes to a topic (first argument of SELECT), and then define 3 pubsub topics to send data to.  Those topics are defined to pass records straight to BigQuery, which is a simple option.  I did also define schemas for those tables (all of this is created in Terraform) which is not ideal, it means this is coupled to schema/code changes for those tables.

```
  # This is to prevent BQ from exceeding the query size limit, specifically used to prevent errors writing Elementary metadata.
  insert_rows_method: gcp-cloud-function
  insert_rows_udf: "_project.dataset.publish_to_pubsub_function_"
  insert_rows_topics: {
    "data_monitoring_metrics": "projects/_project_/topics/elementary-monitoring-metrics-topic",
    "test_result_rows": "projects/_project_/topics/elementary-test-result-rows-topic",
    "elementary_test_results": "projects/_project_/topics/elementary-elementary-test-results-topic"
  }
  query_max_size: 100000
```

PR #627, opened 2023-12-07
ELE-2132: Freshness first metric is null

dapollak:


macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -71,17 +71,21 @@
...
 
 {% macro get_timestamp_table_query(monitored_table, metric_properties, timestamp_column, table_monitors, min_bucket_start, max_bucket_end, full_table_name_str) %}
-    with monitored_table as (
+    with not_time_filtered_monitored_table as (
         select
             {{ elementary.edr_cast_as_timestamp(timestamp_column) }} as {{ timestamp_column }}
             {%- if metric_properties.timestamp_column and metric_properties.event_timestamp_column %}
             , {{ elementary.edr_cast_as_timestamp(metric_properties.event_timestamp_column) }} as {{ metric_properties.event_timestamp_column }}
             {%- endif %}
         from {{ monitored_table }}
```
haritamar, line 80: Maybe to be 100% safe regarding performance, we can add still add here a time filter but do something like:
```
timestamp_column >= min_bucket_start - <buffer>
```
?

haritamar (approved):

PR #628, opened 2023-12-08
:bug: Fixed exposure schema validity test err messages

pushrbx:
Problem:
When the exposure schema validity test fails for data type mismatches, it doesn't show what was the received type.
Example:
```
different data type for the column ASOF_DATE DATE vs 
```
This PR fixes that, and adds a new test for it.

Note:
- The added test can be also merged together with `test_exposure_schema_validity_correct_columns_and_invalid_type` test, but I wasn't sure.


erikzaadi (approved):
image

haritamar (approved):

PR #629, opened 2023-12-10
Ensure that exposure validity test verifies multiple exposure scenario

erikzaadi:
Before this fix, we would never check the exposure validity if we had more than one exposure for that model

### DIFF
--- a/macros/utils/cross_db_utils/get_profile_creation_query.sql
+++ b/macros/utils/cross_db_utils/get_profile_creation_query.sql
@@ -55,14 +55,13 @@ CREATE OR REPLACE PROCEDURE ELEMENTARY_GRANT_INFO_SCHEMA_ACCESS(database_name ST
 {%- set databases = elementary.get_configured_databases_from_graph()%}
 {% for database in databases %}
 {#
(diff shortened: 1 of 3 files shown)


NoyaArie (approved):
image

macros/utils/cross_db_utils/get_profile_creation_query.sql
```diff
@@ -55,14 +55,13 @@ CREATE OR REPLACE PROCEDURE ELEMENTARY_GRANT_INFO_SCHEMA_ACCESS(database_name ST
...
 {#
-  'snowflake' database is excluded because it does not support granting individual privileges (we ask for relevant access to it 
+  'snowflake' database is excluded because it does not support granting individual privileges (we ask for relevant access to it
   via the database roles below)
   see: https://docs.snowflake.com/en/sql-reference/account-usage#enabling-the-snowflake-database-usage-for-other-roles
 #}
-  {%- if database | lower == 'snowflake' -%}
-    {%- continue -%}
-  {%- endif -%}
+  {%- if database | lower != 'snowflake' -%}
```
erikzaadi, line 62: @pei0804 FYI, this caused the e2e to fail

PR #634, opened 2023-12-14
Support dbt retry

haritamar:


macros/utils/command_type_utils.sql
```diff
@@ -0,0 +1,11 @@
+{% macro is_test_command(command) %}
```
elongl, line 1: Maybe the macros shouldn't receive anything and read `flags.WHICH` themselves given that it's a global?

haritamar: done

PR #648, opened 2024-01-14
handle passing dimension test

ofek1weiss:
currently, when a dimension anomalies test passes, two weird things happen in the report:

its "Test Type" is set to "generic" (instead of "dimension")

when expanding it, it shows "Not enough data to calculate anomaly score", which is incorrect

fixes:

change the test type to "dimension"


haritamar:
Closing after discussing with @ofek1weiss , we'll reopen if relevant

PR #651, opened 2024-01-16
remove explicit dependencies of anomaly tests in elementary tables to…

ellakz:
… fix build bug

### DIFF
--- a/macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
+++ b/macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
@@ -6,9 +6,8 @@
     {%- if not data_monitoring_metrics_table %}
         {#  data_monitoring_metrics_table is none except for integration-tests that test the get_anomaly_scores_query macro,
           and in which case it holds mock history metrics #}
-          {%- set data_monitoring_metrics_table = ref('data_monitoring_metrics') %}
+        {%- set data_monitoring_metrics_table = elementary.get_elementary_relation('data_monitoring_metrics') %}
     {%- endif %}
-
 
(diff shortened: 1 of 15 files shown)


macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -6,10 +6,10 @@
     {%- if not data_monitoring_metrics_table %}
         {#  data_monitoring_metrics_table is none except for integration-tests that test the get_anomaly_scores_query macro,
           and in which case it holds mock history metrics #}
-          {%- set data_monitoring_metrics_table = ref('data_monitoring_metrics') %}
+        {%- set elementary_database, elementary_schema = elementary.get_package_database_and_schema() %}
```
elongl, line 9: Can you use `elementary.get_elementary_relation` instead of `ref`?
Apply that to all.

ellakz: get_elementary_relation uses get_node, won't that create a dependency between the models? because that's what we were trying to avoid here

elongl: The [`get_node`](https://github.com/elementary-data/dbt-data-reliability/blob/f1a343e41581394a25a00d8f69b933c25850ba65/macros/utils/graph/get_node.sql#L0-L1) macro only searches the graph. For dependency resolution, dbt only uses `ref`.

ellakz: changed. looks like it's still working

elongl (approved):

PR #657, opened 2024-01-24
Add `unique_combination_of_columns` to `common_tests_configs_mapping`

syou6162:
We use `unique_combination_of_columns` from dbt_utils very often. We would like to be able to include the value of `quality_dimension` when using `unique_combination_of_columns`. Therefore, we have added `unique_combination_of_columns` to `common_tests_configs_mapping`.

### DIFF
--- a/macros/utils/common_test_configs.sql
+++ b/macros/utils/common_test_configs.sql
@@ -312,7 +312,7 @@
         },
         "unique_combination_of_columns": {
           "quality_dimension": "uniqueness",
-          "failed_row_count_calc": "sum(num_rows)"
+          "failed_row_count_calc": "count(*)"
         }
       },
       "elementary": {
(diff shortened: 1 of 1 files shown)


syou6162:
@ofek1weiss @ellakz @dapollak Coud you review this pull request?

ofek1weiss:
Hey @syou6162 , i believe this wont work, as the results of `unique_combination_of_columns` does not contain a column named `num_rows` (as you can see [here](https://github.com/dbt-labs/dbt-utils/blob/23da1f424398696e25656dc6f89066b9a5d13794/macros/generic_tests/unique_combination_of_columns.sql)), you can see for example that in the test `cardinality_equality` [here](https://github.com/dbt-labs/dbt-utils/blob/23da1f424398696e25656dc6f89066b9a5d13794/macros/generic_tests/cardinality_equality.sql#L21), there is a column named `num_rows`

syou6162:
@ofek1weiss Thanks for the review! 

I understand the details of `failed_row_count_calc`. For `unique_combination_of_columns`, the column name `columns_csv` changes dynamically, so it certainly doesn't work with `sum(num_rows)`.

I took this point and changed it to sum with `count(*)` as well as `accepted_range`. Please review again.

PR #660, opened 2024-02-08
Support all anomaly vars on all configuration levels

ofek1weiss:


macros/edr/system/system_utils/get_test_argument.sql
```diff
@@ -1,13 +1,12 @@
 {% macro get_test_argument(argument_name, value, model_graph_node) %}
```
ofek1weiss, line 1: differentiate between `undefined` and `None`

macros/edr/system/system_utils/get_config_var.sql
```diff
@@ -19,8 +19,6 @@
 {# We use this macro to define and call vars, as the global vars defined in dbt_project.yml
    of the package are not accessible at on-run-start and on-run-end #}
   {% set default_config = {
-    'detection_period': elementary.get_period_default_var(period='day', count=2),
-    'training_period': elementary.get_period_default_var(period='day', count=14),
```
ofek1weiss, line 23: as these overwrite `days_back` and `backfill_days`, these default var values overwrite model/test values

elongl (approved):

PR #661, opened 2024-02-08
ELE-2470 temp tables are not being deleted

IDoneShaveIt:


macros/edr/system/hooks/on_run_end.sql
```diff
@@ -5,6 +5,14 @@
         {% do return("") %}
       {% endif %}
 
+      {# 
+        Elementary temp tables are not realy temp and should be cleaned on the end of the run.
+        We want to make sure we clean the temp tables even if elementary on run end hooks are disabled.
+      #}
+      {% if elementary.get_config_var("clean_elementary_temp_tables") %}
```
IDoneShaveIt, line 12: Not sure if we still want to keep this var.
@haritamar WDYT?

dapollak (approved):

PR #674, opened 2024-03-05
improvement: bigquery specific for query_table_metrics

dongchris:
Using information schema to get row count is much more performant than doing a full table scan

https://cloud.google.com/bigquery/docs/information-schema-table-storage#schema

### DIFF
--- a/macros/edr/materializations/model/metrics.sql
+++ b/macros/edr/materializations/model/metrics.sql
@@ -25,8 +25,8 @@
 {% macro bigquery__query_table_metrics() %}
     select
       {{ modules.datetime.datetime.utcnow().timestamp() }} as build_timestamp,
-      row_count
-    from {{ this.schema }}.__TABLES__
+      total_rows AS row_count
+    from {{ this.database }}.{{ this.schema }}.INFORMATION_SCHEMA.TABLE_STORAGE
     where table_id = '{{ this.name }}'
   {% endset %}
 
--- a/macros/edr/materializations/model/metrics.sql
+++ b/macros/edr/materializations/model/metrics.sql
@@ -23,6 +23,7 @@
 {% endmacro %}
 
 {% macro bigquery__query_table_metrics() %}
+  {% set query %}
     select
       {{ modules.datetime.datetime.utcnow().timestamp() }} as build_timestamp,
       total_rows AS row_count
--- a/macros/edr/materializations/model/metrics.sql
+++ b/macros/edr/materializations/model/metrics.sql
@@ -33,7 +33,11 @@
   {% set metrics = [] %}
   {% for metric_column in elementary.run_query(query).columns %}
     {% set metric_name = metric_column.name %}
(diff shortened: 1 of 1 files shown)


macros/edr/materializations/model/metrics.sql
```diff
@@ -22,6 +22,30 @@
...
+      {{ modules.datetime.datetime.utcnow().timestamp() }} as build_timestamp,
+      row_count
+    from {{ this.schema }}.__TABLES__
+    where table_id = '{{ this.name }}'
+  {% endset %}
+
+  {% set metrics = [] %}
+  {% for metric_column in elementary.run_query(query).columns %}
+    {% set metric_name = metric_column.name %}
+    {% set metric_value = metric_column[0] %}
```
haritamar, line 36: Can you add a safety that the query actually returns any results? (it should, but mainly to avoid edge cases / never fail here)

haritamar:
Thanks @dongchris for the awesome contribution!
Actually realized as well just recently that `count(*)` does a full scan in BQ, we were sure from some reason it's optimized. 

I've added a couple of comments, will be great if you can take a look.
Also worth verifying the the [integration tests](https://github.com/elementary-data/dbt-data-reliability/blob/master/CONTRIBUTING.md#running-integration-tests) pass, but will also check that.

dongchris:
> Thanks @dongchris for the awesome contribution! Actually realized as well just recently that `count(*)` does a full scan in BQ and didn't get to fix this yet, so I'm glad about this.
> 
> I've added a couple of comments, will be great if you can take a look. Also worth verifying the the [integration tests](https://github.com/elementary-data/dbt-data-reliability/blob/master/CONTRIBUTING.md#running-integration-tests) pass, but will also check that.

For sure! This personally cuts down our Bigquery consumption by ~7% and is huge in cost savings. This matters a lot when the data volume is huge (TB, PBs)

macros/edr/materializations/model/metrics.sql
dongchris: can you clarify what is the expected behavior? if `elementary.run_query(query).columns` returns 0 columns, then it just wouldn't update `metrics` dict right?

dongchris:
> Thanks @dongchris for the awesome contribution! Actually realized as well just recently that `count(*)` does a full scan in BQ and didn't get to fix this yet, so I'm glad about this.
> 
> I've added a couple of comments, will be great if you can take a look. Also worth verifying the the [integration tests](https://github.com/elementary-data/dbt-data-reliability/blob/master/CONTRIBUTING.md#running-integration-tests) pass, but will also check that.

addressed your comments

was having trouble running the integration tests (followed the instructions on the elementary website) so if you could help check that, it would be great!

macros/edr/materializations/model/metrics.sql
dongchris: actually, added! @haritamar

haritamar (approved):

Issue #675, opened 2024-03-06
Error `The query is too large. The maximum standard SQL query length is 1024.00K characters`  when using insert_rows_method `chunk`


aibazhang:
Hi guys!

I've added arg chunk_size when calling insert_rows to fix the error `Maximum number of resources referenced per query are 1,000 resources` serval days ago.  https://github.com/elementary-data/dbt-data-reliability/pull/669

However, a new error has occurred `The query is too large. The maximum standard SQL query length is 1024.00K characters, including comments and white space characters.` 

It appears that the compiled_code was too long when running following query, even though we only have 400 records to insert.

```sql
insert into `my-project`.`elementary`.`dbt_run_results`
         (model_execution_id,unique_id,invocation_id,generated_at,created_at,name,message,status,resource_type,execution_time,execute_started_at,execute_completed_at,compile_started_at,compile_completed_at,rows_affected,full_refresh,compiled_code,failures,query_id,thread_id,materialization,adapter_response) values ...
``` 

Is there a way to make `max_query_size` work even when using insert_rows_method `chunk`?
If you have any ideas, please let me know, I can submit a PR.

Thank you

Environment
```markdown
- dbt=1.7.8
- adapter: bigquery=1.7.4
- elementary-data/elementary=0.14.1
```

Variables
```yaml
elementary:
    "insert_rows_method": "chunk"
    "dbt_artifacts_chunk_size": 500
```

elongl:
Hi @aibazhang, thanks a lot for your contribution!
Would you please care to elaborate on why you're using the `chunk` method instead of `max_query_size`?
Asking because I think it'd make more sense to solve a query size issue that's imposed by BigQuery.

tsmo4:
> Hi @aibazhang, thanks for opening this issue. Would you please care to elaborate on why you're using the `chunk` method instead of `max_query_size`? Asking because I think it'd make more sense to solve a query size issue that's imposed by BigQuery.

Speaking to our use case, we have a sizable DBT project, and it appears that we need _both_ options in place on a run of the post hook in order for it to execute. When executing the upload of the dbt test meta data we hit errors:
- without the `chunk` method in place Big Query throws a resource error, due to the size and complexity of the generated query (not it's code length) `Maximum number of resources`
- with the `chunk` method in place, which invalidates the  `max_query_size` restriction, the query becomes too long and BQ throws an error. `The query is too large`

elongl:
[...]

PR #679, opened 2024-03-12
support query_max_size for insert_rows_method `chunk`

aibazhang:
Hey guys,

To fix the issue raised in
https://github.com/elementary-data/dbt-data-reliability/issues/675
I added support for query_max_size to the insert_rows_method `chunk`.

This addition is important to prevent errors like `The query is too large. The maximum standard SQL query length is 1024.00K characters` (in BigQuery) when use the insert_rows_method `chunk`

# Test

I intended to write a unit test for this change, but I encountered errors running the current test cases even after following the procedure outlined in
https://github.com/elementary-data/dbt-data-reliability/blob/master/integration_tests/README.md

I faced conflicting dependencies errors when attempting to install the required packages.

```
ERROR: Cannot install elementary-data and urllib3==2.0.4 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested urllib3==2.0.4
    dbt-core 1.7.9 depends on urllib3~=1.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

```

After I added `elementary-data` and `dbt-postgres` to the `requirements.txt` , another error occurred.

requirements.txt
```
pytest
pytest-xdist
pytest-parametrization
pytest-html
filelock
# urllib3==2.0.4
elementary-data==0.14.1
dbt-postgres==1.7.9
```

error

```
ERROR    elementary.clients.dbt.dbt_runner:dbt_runner.py:175 Encountered an error:
Compilation Error
  dbt found 2 package(s) specified in packages.yml, but only 0 package(s) installed in dbt_packages. Run "dbt deps" to install package dependencies.
```

Therefore, instead of writing a unit test, I tested this PR by running `dbt run` in our own environment, and it worked well.

### DIFF
--- a/macros/utils/list_utils/split_list_to_chunks.sql
+++ b/macros/utils/list_utils/split_list_to_chunks.sql
@@ -1,13 +1,16 @@
 {% macro split_list_to_chunks(item_list, chunk_size=50) %}
     {% set chunks = [] %}
     {% set current_chunk = [] %}
+    {% set current_length = 0 %}
     {% for item in item_list %}
         {% set reminder = loop.index0 % chunk_size %}
-        {% if reminder == 0 and current_chunk %}
+        {% if (reminder == 0 and current_chunk) or (current_length + item | length) > elementary.get_config_var('query_max_size') %}
             {% do chunks.append(current_chunk.copy()) %}
             {% do current_chunk.clear() %}
+            {% set current_length = 0 %}
         {% endif %}
         {% do current_chunk.append(item) %}
+        {% do current_length + item | length %}
     {% endfor %}
     {% if current_chunk %}
         {% do chunks.append(current_chunk) %}


elongl:
Hi @aibazhang, thanks a lot for your contribution!
Would you please care to elaborate on why you're using the `chunk` method instead of `max_query_size`?
Asking because I think it'd make more sense to solve a query size issue that's imposed by BigQuery.

haritamar:
Looks good, merging!
(re. BQ - we have an issue in our test env, not a problem here)

haritamar (approved):

PR #686, opened 2024-04-01
Add get_requires_permissions and validate_required_permissions macros

IDoneShaveIt:


erikzaadi (approved):

macros/utils/graph/get_relevant_databases.sql
```diff
@@ -0,0 +1,14 @@
...
+    {% set sources = graph.sources.values() | selectattr('resource_type', '==', 'source') | list %}
+    {% set nodes = models + sources %}
+    {% for node in nodes %}
+        {% set database_name = node.get('database') %}
+        {% if database_name %}
+            {% do database_names.append(database_name) %}
+        {% endif %}
+    {% endfor %}
+    {% set unique_database_names = database_names | unique | list %}
+    {% do return(unique_database_names) %}
```
ofek1weiss, line 13: You need to add target.database, as as there can be a config where all the models are built in it (they will appear in the query history in it) but none of the tables are saved there

IDoneShaveIt: I am not sure I understand how could the graph's nodes have a different database than the one they are written too 🤔 ?
Care to explain to me the case? I feel like I am missing something 🙂

PR #687, opened 2024-04-05
Ele 2853 trino tests

haritamar:


macros/utils/table_operations/get_relation_max_length.sql
```diff
@@ -28,5 +28,5 @@
 {% endmacro %}
 
 {% macro trino__get_relation_max_name_length(temporary, relation, sql_query) %}
-    {{ return(255) }}
+    {{ return(128) }}
```
haritamar, line 31: When running with hive it seems that this limit needs to be lower, so lowered it
(Preferred adding that rather than Glue so it will be completely local)

PR #694, opened 2024-04-18
ELE-2932: Column anomalies group_by

dapollak:


ofek1weiss (approved):

macros/edr/materializations/test/test.sql
```diff
@@ -170,6 +171,8 @@
   {% set test_results_description %}
       {% if has_anomaly_score %}
           {{ elementary.insensitive_get_dict_value(latest_row, 'anomaly_description') }}
+      {% elif rows_with_score %}
```
haritamar, line 174: Maybe it makes sense to remove `has_anomaly_score` and keep only `rows_with_score`? (and also apply to the log above)

haritamar (approved):

PR #698, opened 2024-04-24
added execution_time to elementary_test_results

ofek1weiss:


ofek1weiss:
this is unneeded, as the execution time is already saved in `dbt_run_results`

PR #707, opened 2024-05-08
Ele 3034 additional package bugfixes

haritamar:


integration_tests/tests/test_dbt_artifacts/test_artifacts.py
```diff
@@ -19,6 +19,11 @@ def test_artifacts_caching(dbt_project: DbtProject):
     second_row = read_model_artifact_row(dbt_project)
     assert first_row == second_row, "Artifacts are not cached at the on-run-end."
 
+
+def test_artifacts_updating(dbt_project: DbtProject):
```
haritamar, line 23: Separated the original test to two, with the second test also running for dbt 1.3.0

macros/edr/data_monitoring/monitors/column_boolean_monitors.sql
```diff
@@ -1,7 +1,7 @@
 {% macro count_true(column_name) -%}
-    coalesce(sum(case when cast({{ column_name }} as {{ elementary.edr_type_bool() }}) is true then 1 else 0 end), 0)
+    coalesce(sum(case when cast({{ column_name }} as {{ elementary.edr_type_bool() }}) = true then 1 else 0 end), 0)
```
haritamar, line 2: The "is true" syntax seems to work only for BQ, where = seems to always work and I think is good enough

macros/edr/materializations/model/incremental.sql
```diff
@@ -62,7 +62,7 @@
   {% do return(relations) %}
 {% endmaterialization %}
 
-{% materialization incremental, adapter="athena", supported_languages=["sql"] %}
+{% materialization incremental, adapter="athena", supported_languages=["sql", "python"] %}
```
haritamar, line 65: Athena added support for python models and because of us they don't work, so added here to the supported languages.
Added to trino as well because there's actually no harm in doing so (we did the same for redshift even though it doesn't yet support Python models)

IDoneShaveIt (approved):
LGTM

PR #708, opened 2024-05-08
Ele 3041 remove model and incremental materializations from package

haritamar:


macros/edr/materializations/test/test.sql
```diff
@@ -1,58 +1,17 @@
-{% macro create_test_result_temp_table() %}
```
haritamar, line 1: I just re-ordered the macros in this file and removed one unused one - no logic change.

elongl (approved):

PR #714, opened 2024-05-19
Fixed Athena temp tables cleaning

bebbo203:
Solves [#1514](https://github.com/elementary-data/elementary/issues/1514).

Specifically for dbt-athena-community, this fix aims to clean the temporary tables in a better way, including dropping files from S3 in case of external tables.

### DIFF
--- a/macros/edr/tests/on_run_end/handle_tests_results.sql
+++ b/macros/edr/tests/on_run_end/handle_tests_results.sql
@@ -111,8 +111,7 @@
     {% do elementary.run_query(insert_query) %}
 
     {% if not elementary.has_temp_table_support() %}
-        {% do adapter.drop_relation(temp_relation) %}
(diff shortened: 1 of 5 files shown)


macros/edr/tests/on_run_end/handle_tests_results.sql
```diff
@@ -160,6 +160,7 @@
 
     {% if not elementary.has_temp_table_support() %}
         {% do adapter.drop_relation(temp_relation) %}
+        {% do adapter.clean_up_table(temp_relation) %}
```
haritamar, line 163: @bebbo203 - this fails in Databricks since it doesn't have this column.
I'm guessing you should use `fully_drop_relation` here as well?

haritamar:
I think this looks good!
There is an unrelated failing test (due to another PR that was recently merged) but here we're good to go.

haritamar (approved):

PR #723, opened 2024-05-29
ELE-3139: Limit calculation bug when `spike_failure_percent_threshold` is specified

dapollak:


macros/edr/data_monitoring/anomaly_detection/get_anomaly_scores_query.sql
```diff
@@ -258,5 +258,5 @@
       {%- endset -%}
     {% endif %}
 
-    {{ return({"min_metric_value": min_val, "max_metric_value": max_val}) }}
+{{ return({"min_metric_value": min_val, "max_metric_value": max_val}) }}
```
NoyaArie, line 261: the indentation should be the same as the `endif`

NoyaArie (approved):
👍 💯

PR #735, opened 2024-07-08
Fix trino temp tables cleanup

ndrluis:


ofek1weiss:
@ndrluis putting it here as well
It seems that this code does not work: https://github.com/elementary-data/dbt-data-reliability/actions/runs/10268957700/job/28413257114?pr=744#step:14:1468
'dbt.context.providers.RuntimeDatabaseWrapper object' has no attribute 'clean_up_table

ndrluis:
Sorry @ofek1weiss, I misunderstood the problem. I have now made the correct fix, but we need to wait for this fix from [dbt-trino](https://github.com/starburstdata/dbt-trino/pull/427)

ndrluis:
I was informed that the correct macro exists in the [dbt-utils-trino](https://github.com/starburstdata/dbt-trino-utils), so to make this work properly, we need to have this package. @ofek1weiss, how can we use this library, or would it be better for us to copy the macro from the library and include it here to avoid requiring all users to install a specific trino package?

mattiamatrix:
Hello, any updates on this? Is it safe to delete the `__tmp_` tables manually?

NoyaOffer:
dbt-trino made the changes - this should work as expected now

mattiamatrix:
@NoyaOffer can you please reference their fix?

PR #736, opened 2024-07-16
Added a required 'name' argument to 'collect_metrics'.

elongl:


macros/edr/tests/test_collect_metrics.sql
```diff
@@ -35,12 +35,16 @@
     {% set table_metrics = [] %}
     {% set col_to_metrics = {} %}
     {% for metric in metrics %}
+        {% if not metric.name %}
```
elongl, line 38: Should we ensure the metric's name is unique across the entire project, or allow the same metric name to be calculated from different tests?

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -66,11 +60,19 @@
...
     )
 {% endmacro %}
 
-{% macro get_timestamp_table_query(monitored_table, metric_properties, timestamp_column, table_monitors, min_bucket_start, max_bucket_end, full_table_name_str) %}
+{% macro get_no_timestamp_volume_query(metric, metric_properties) %}
+    select
+        {{ elementary.const_as_string(metric.name) }} as metric_name,
+        {{ elementary.row_count() }} as metric_value
+    from monitored_table
+    group by 1
```
elongl, line 72: What's this `group by` for?

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -152,39 +151,46 @@
...
+            "row_count": elementary.get_no_timestamp_volume_query,
+            "event_freshness": elementary.get_no_timestamp_event_freshness_query
+        } %}
+    {% endif %}
 
-    {%- set metric_macro = metrics_macro_mapping.get(metric_name) %}
+    {%- set metric_macro = metrics_macro_mapping.get(metric.type) %}
     {%- if not metric_macro %}
-        {%- do return(none) %}
+        {%- do exceptions.raise_compiler_error("Unable to get macro query for metric '{}'".format(metric.type)) %}
```
elongl, line 182: _Note_: Changed this to raise an exception.

macros/edr/tests/test_collect_metrics.sql
haritamar: Maybe we should check that it's unique in the **model**?
Since in any way in the cloud the test is defined on a specific model (or source) - so I don't really care if the same metric name appears across models.
But I do want to avoid relying on the dbt test ID for filtering the metric (since it's sensitive to the test parameters)

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
haritamar: Seems redundant

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
```diff
@@ -152,39 +151,46 @@
...
+    {% if metric_properties.timestamp_column %}
+        {% set metrics_macro_mapping = {
+            "row_count": elementary.row_count_metric_query,
+            "freshness": elementary.freshness_metric_query,
+            "event_freshness": elementary.event_freshness_metric_query
+        } %}
+    {% else %}
+        {% set metrics_macro_mapping = {
+            "row_count": elementary.get_no_timestamp_volume_query,
+            "event_freshness": elementary.get_no_timestamp_event_freshness_query
```
haritamar, line 176: Maybe let's ensure that we have at least a sanity test for each of the "no timestamps" (I'm not 100% sure if we do or not)

macros/edr/tests/test_utils/collect_table_metrics.sql
```diff
@@ -18,42 +18,39 @@
...
         {% do metric_names.append(metric.name) %}
     {% endfor %}
-    {% set table_monitors = elementary.get_final_table_monitors(monitors=metric_names) %}
-    {% if not table_monitors %}
-        {% do return(none) %}
-    {% endif %}
-
-    {% if dimensions and table_monitors != ["row_count"] %}
-        {% do exceptions.raise_compiler_error("collect_metrics test does not support non row_count dimensional table metrics.") %}
-    {% endif %}
```
haritamar, line 28: Why did you remove this? Seems still to be a relevant check (that we support dimensions only for row count)

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
```diff
@@ -1,11 +1,18 @@
...
     {% for dimension_column in dimensions %}
       {% do prefixed_dimensions.append("dimension_" ~ dimension_column) %}
     {% endfor %}
 
+    {% set metric_types = [] %}
+    {% set metric_name_to_type = {} %}
+    {% for metric in column_metrics %}
+        {% do metric_types.append(metric.type) %}
+        {% do metric_name_to_type.update({metric.name: metric.type}) %}
+    {% endfor %}
```
haritamar, line 14: Not sure this is the right place, but can we also save `metric_type` in `data_monitoring_metrics`?
(not sure what's the right place for this comment)

macros/edr/tests/test_utils/collect_table_metrics.sql
elongl: Was just moved below.

macros/edr/data_monitoring/monitors_query/table_monitoring_query.sql
elongl: We do for [volume](https://github.com/elementary-data/dbt-data-reliability/blob/5f2ad803e76dcadf6e5d483c3d24705a015966f3/integration_tests/tests/test_volume_anomalies.py#L195), I'll add one for `event_freshness`.

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
haritamar: Hmm the only case where I think it might matter is when the metric definition is **changed** - e.g. the name remains the same but the metric type is changed.
But that makes me think that it should actually be added to `metric_properties`. WDYT?

This achieves two things:
1. The metric will be backfilled if the type changes (same as any other configuration change like `timestamp_column`)
2. In the cloud I implemented logic that only pulls metrics with the most recent metric properties (to ensure we only rely on metrics collected with the most recent configuration - a bit like what happens in `get_anomaly_scores_query`).

Maybe we want to add it only to metrics collected via `collect_metrics` to avoid invalidating all the metrics in existing OSS tests for users (though not sure if it's critical).

macros/edr/tests/test_utils/collect_table_metrics.sql
haritamar: I see below that you just check if there is one metric, but not that its type is row_count

macros/edr/tests/test_collect_metrics.sql
elongl: Added [here](https://github.com/elementary-data/dbt-data-reliability/pull/736/commits/ce97140a6f6487dc7a4dd39e3d522d124b282c78).

I originally implemented it using the `cache` and therefore the changes,
but it made sense for me that they'd live in the same module so I kept it.

macros/edr/tests/test_utils/collect_table_metrics.sql
elongl: https://github.com/elementary-data/dbt-data-reliability/pull/736/commits/12a7cc97e3210a9e0d2feaf6d5ff2af7e1643741

haritamar (approved):

PR #738, opened 2024-07-19
ele-1641 add training_period parameter to the event_freshness_anomalies

annav00:
Fix for https://github.com/elementary-data/elementary/issues/1641

Added `training_period` parameter to the event_freshness_anomalies test


haritamar:
Seems like the fix was merged in a separate PR, so closing (please lmk if I'm wrong)

PR #745, opened 2024-08-07
Add seed_run_results model

kokorin:


models/edr/run_results/seed_run_results.sql
```diff
@@ -0,0 +1,44 @@
...
+    run_results.execute_started_at,
+    run_results.execute_completed_at,
+    run_results.compile_started_at,
+    run_results.compile_completed_at,
+    run_results.compiled_code,
+    run_results.adapter_response,
+    run_results.thread_id,
+    seeds.database_name,
+    seeds.schema_name,
+    coalesce(run_results.materialization, seeds.materialization) as materialization,
```
ofek1weiss, line 36: looks like the `dbt_seeds` model does not have a `materialization` column, you can just do 
```suggestion
    run_results.materialization,
```

kokorin: Thank you, all test passed locally, I will check why

kokorin: Added tests to validate elementary view models including `seed_run_results`.

ofek1weiss (approved):
Very nice, approved 👍 🚀

PR #748, opened 2024-08-28
Ele 3577 add index data monitoring metrics 

NoyaArie:


models/edr/data_monitoring/data_monitoring/data_monitoring_metrics.sql
```diff
@@ -3,6 +3,7 @@
     materialized='incremental',
     unique_key='id',
     on_schema_change='append_new_columns',
+    indexes=[{'columns': ['full_table_name', "column_name", "metric_name"]}] if target.type == "postgres" else [],
```
ofek1weiss, line 6: i believe `indexes` is a configuration that only applies to postgres, so the `if target.type == "postgres"` check is not necessary, but it can be nice for clarity. so idk

ofek1weiss (approved):

PR #758, opened 2024-10-02
feat: collect job run id from env vars

AlexBoutou:
This PR adds collection of JOB_RUN_ID from env_vars.

[The doc specifies those](https://github.com/elementary-data/elementary/blob/master/docs/_snippets/guides/collect-job-data.mdx) env vars should be collected : ORCHESTRATOR, JOB_NAME, JOB_ID, JOB_URL, JOB_RUN_ID, JOB_RUN_URL

But JOB_RUN_ID is currently missing.


haritamar (approved):
Thanks for the contribution, approved!

PR #763, opened 2024-11-02
Update split_list_to_chunks.sql

Niteeshkanungo:
The old method mentioned is incorrectly calculating and updating the iteration for BQ load. 
The correction should fix the issue for dbt artifact loads.

### DIFF
--- a/macros/utils/list_utils/split_list_to_chunks.sql
+++ b/macros/utils/list_utils/split_list_to_chunks.sql
@@ -12,7 +12,7 @@
             {% set current_length = 0 %}
         {% endif %}
         {% do current_chunk.append(item) %}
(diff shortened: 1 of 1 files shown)


Smitspeed29 (approved):

haritamar:
Hi @Niteeshkanungo , thanks a lot for contributing this PR!
After some testing and a closer look at the code, I realized this is not the only thing broken with the `chunk` insert method, and it's still not going to work...

So I created a separate PR that adds chunk size support to the `max_query_size` method, and made it the only method.

haritamar:
Closing since [this PR](https://github.com/elementary-data/dbt-data-reliability/pull/765) was merged instead.

PR #765, opened 2024-11-05
Ele 3771 have a single insert rows method that supports both limits

haritamar:


.github/workflows/test-all-warehouses.yml
```diff
@@ -37,8 +37,8 @@ jobs:
       matrix:
         dbt-version:
           ${{ inputs.dbt-version && fromJSON(format('["{0}"]', inputs.dbt-version)) ||
-          ! contains(github.event_name, 'pull_request') && fromJSON('["1.3.0", "latest_pre"]') ||
-          fromJSON('["latest_pre"]') }}
+          ! contains(github.event_name, 'pull_request') && fromJSON('["1.3.0", "latest_official"]') ||
+          fromJSON('["latest_official"]') }}
```
haritamar, line 41: Some pre versions are broken again, and I want tests to work.
I added a `latest_pre` run only on postgres so if there are cross-db things that break we'll notice, wdyt about this middle ground?

Niteeshkanungo: Isn't query_max_size different for different warehousing solutions?

haritamar: Hi @Niteeshkanungo - my comment above is specifically about testing on dbt pre-releases.
The change above will still run on all warehouses with the latest official dbt version, but test the latest dbt pre-release (right now `1.9.0b3`) only on postgres.

elongl (approved):

PR #771, opened 2025-01-03
Remove transaction for bigquery in delete_and_insert macro

rloredo:
Solves https://github.com/elementary-data/elementary/issues/1766

### DIFF
--- a/macros/utils/table_operations/delete_and_insert.sql
+++ b/macros/utils/table_operations/delete_and_insert.sql
@@ -132,3 +132,18 @@
 
     {% do return(queries) %}
 {% endmacro %}
+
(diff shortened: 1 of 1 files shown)


macros/utils/table_operations/delete_and_insert.sql
```diff
@@ -132,3 +132,26 @@
 
     {% do return(queries) %}
 {% endmacro %}
+
+{% macro bigquery__get_delete_and_insert_queries(relation, insert_relation, delete_relation, delete_column_key) %}
```
elazarlachkar, line 136: While your implementation is valid, your end goal of running 2 dbt runs in parallel won't be resolved by this.
Commented on the [issue](https://github.com/elementary-data/elementary/issues/1766).

PR #793, opened 2025-03-14
fix: filter dbt models to upload only current project

cedric-orange:
When you are using some tools like [dbt-loom](https://github.com/nicholasyager/dbt-loom), `graph.nodes.values()` returns nodes from multiples project.

dbt-loom works by fetching public model definitions from your dbt artifacts, and injecting those models into your dbt project.

The goal of this PR is to only retrieve nodes from the current project using [project_name](https://docs.getdbt.com/reference/dbt-jinja-functions/project_name) selection to avoid uploading of unnecessary information. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artifact filtering to ensure all dbt artifacts (models, seeds, snapshots, sources, tests, columns, and exposures) are properly scoped to the current project during upload operations, preventing cross-project artifact mixing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_columns.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_columns.sql
@@ -1,7 +1,7 @@
 {%- macro upload_dbt_columns(should_commit=false, metadata_hashes=none) -%}
     {% set relation = elementary.get_elementary_relation('dbt_columns') %}
     {% if execute and relation %}
-        {% set tables = graph.nodes.values() | selectattr("package_name", "equalto", project_name) | list + graph.sources.values() | list %}
+        {% set tables = graph.nodes.values() | selectattr('package_name', '==', project_name) | list + graph.sources.values() | selectattr('package_name', '==', project_name) | list %}
         {% do elementary.upload_artifacts_to_table(relation, tables, elementary.flatten_table_columns, should_commit=should_commit, metadata_hashes=metadata_hashes) %}
     {%- endif -%}
     {{- return('') -}}
--- a/macros/edr/dbt_artifacts/upload_dbt_exposures.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_exposures.sql
@@ -1,7 +1,7 @@
 {%- macro upload_dbt_exposures(should_commit=false, metadata_hashes=none) -%}
     {% set relation = elementary.get_elementary_relation('dbt_exposures') %}
     {% if execute and relation %}
-        {% set exposures = graph.exposures.values() | selectattr('resource_type', '==', 'exposure') | selectattr("package_name", "equalto", project_name) %}
+        {% set exposures = graph.exposures.values() | selectattr('resource_type', '==', 'exposure') | selectattr('package_name', '==', project_name) %}
         {% do elementary.upload_artifacts_to_table(relation, exposures, elementary.flatten_exposure, should_commit=should_commit, metadata_hashes=metadata_hashes) %}
     {%- endif -%}
     {{- return('') -}}
--- a/macros/edr/dbt_artifacts/upload_dbt_models.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_models.sql
@@ -1,7 +1,7 @@
 {%- macro upload_dbt_models(should_commit=false, metadata_hashes=none) -%}
     {% set relation = elementary.get_elementary_relation('dbt_models') %}
     {% if execute and relation %}
-        {% set models = graph.nodes.values() | selectattr('resource_type', '==', 'model' ) | selectattr("package_name", "equalto", project_name) %}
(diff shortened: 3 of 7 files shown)


michael-myaskovsky:
Thanks for your contribution!
Please add a flag for this behavior, after that I'll be able to merge.

michael-myaskovsky:
Since it changes the behavior, we want it to be an option that the user actively chooses. Either by configuration, or  by an argument that the program receives. With the default being set to the original behavior.

cedric-orange:
This does not change current behavior.
But this protects against duplicates that are populated in the database when dbt-loom is used.

macros/edr/dbt_artifacts/upload_dbt_columns.sql
```diff
@@ -1,7 +1,7 @@
 {%- macro upload_dbt_columns(should_commit=false, metadata_hashes=none) -%}
     {% set relation = elementary.get_elementary_relation('dbt_columns') %}
     {% if execute and relation %}
-        {% set tables = graph.nodes.values() | list + graph.sources.values() | list %}
+        {% set tables = graph.nodes.values() | selectattr("package_name", "equalto", project_name) | list + graph.sources.values() | list %}
```
elazarlachkar: I'm not familiar with `project_name`, but noticed in the dbt docs that it is accessed using the context (`context[project_name]`). Does this work without it?

elazarlachkar (commented):
Hi @cedric-orange!

I'm sorry for the delayed response.
Your PR looks good to me and I would like to merge it, I just commented with one question.
Could you look at it, and also update your branch to the latest `elementary-data:master` for the CI to pass?

Also, is that relevant also to groups and metrics artifacts?

macros/edr/dbt_artifacts/upload_dbt_columns.sql
cedric-orange: It doesn't work without it. The goal here is to ensure that only nodes from the current project are returned.
In most cases, this is normal behavior. 
This is necessary when there are multiple projects information in the same manifest.

elazarlachkar: @cedric-orange I meant is the syntax `project_name` (rather that `context[project_name]`) works and the actual project name is being resolved.
I tested it myself and I see it does, so resolving :)

elazarlachkar: We need to filter the `graph.sources` as well, like you did in `macros/edr/dbt_artifacts/upload_dbt_sources.sql`

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
```diff
@@ -1,7 +1,7 @@
 {%- macro upload_dbt_exposures(should_commit=false, metadata_hashes=none) -%}
     {% set relation = elementary.get_elementary_relation('dbt_exposures') %}
     {% if execute and relation %}
-        {% set exposures = graph.exposures.values() | selectattr('resource_type', '==', 'exposure') %}
+        {% set exposures = graph.exposures.values() | selectattr('resource_type', '==', 'exposure') | selectattr("package_name", "equalto", project_name) %}
```
elazarlachkar: Let's use `==` instead of `equalto` for consistency with the first part.

macros/edr/dbt_artifacts/upload_dbt_columns.sql
cedric-orange: done

macros/edr/dbt_artifacts/upload_dbt_exposures.sql
cedric-orange: done

elazarlachkar:
@cedric-orange thanks for resolving my comments.

I now realised this PR changes the behavior a bit, as currently the artifacts include also artifacts from dbt package dependencies (for example, if you use [dbt_project_evaluator](https://hub.getdbt.com/dbt-labs/dbt_project_evaluator/latest/) as a dependency, its models will be included).
I'll check if this behavior change is Ok with out Product team.

If this change is not Ok, we'll need to disable your feature by default, and add a flag which enables it.
I'm thinking of a generic macro which filters "entities" from the Graph. This macro can check for this flag, and be called by the different "upload_x_artifacts". Let me know what you think.

alexandre-bernard:
Hello,

Any update on this issue ?

cedric-orange:
> I'll check if this behavior change is Ok with out Product team.

@elazarlachkar Do you have feed back of your product owner ?

haritamar:
Hi all, sorry for the long delay.
I've create a separate PR off this one, but changed this to be an opt-in flag so we won't break existing behavior.
https://github.com/elementary-data/dbt-data-reliability/pull/968

This will be in the next release.

Thanks,
Itamar

PR #803, opened 2025-04-03
Support athena in deprecated tests

GuyEshdat:


macros/edr/system/system_utils/full_names.sql
```diff
@@ -60,6 +60,19 @@
     trim(split_part(full_table_name,'.',{{ part_index }}),'"') as {{ part_name }}
 {% endmacro %}
 
+{% macro athena__full_name_split(part_name) %}
```
MikaKerman, line 63: Is there any way to reduce code duplication here?

macros/utils/cross_db_utils/contains.sql
```diff
@@ -30,4 +30,16 @@
             lower({{ string }}) like lower('%{{ string_to_search }}%') then true
         else false end
     {%- endif %}
+{% endmacro %}
+
+{% macro athena__contains(string, string_to_search, case_sensitive) %}
```
MikaKerman, line 35: here too

MikaKerman (approved):

PR #805, opened 2025-04-15
Ele 4291 increase minimum dbt version in oss tests

haritamar:


.github/workflows/test-all-warehouses.yml
```diff
@@ -37,29 +37,28 @@ jobs:
...
+          - dbt-version: "${{ inputs.dbt-version || '1.7.0' }}"
             warehouse-type: postgres
           - dbt-version: "${{ inputs.dbt-version || 'latest_pre' }}"
             warehouse-type: postgres
         exclude:
-          - dbt-version: "1.3.0"
+          - dbt-version: "1.7.0"
             warehouse-type: athena
-          - dbt-version: "1.3.0"
+          - dbt-version: "1.7.0"
```
ofek1weiss, line 61: wasnt the reason that they were excluded that the adapters did not exist in `dbt 1.3`? if so doesnt it make sense to remove the exclusion?

PR #812, opened 2025-05-12
Ele 4399 fix clear tests

michael-myaskovsky:


integration_tests/deprecated_tests/macros/e2e_tests/clear_tests.sql
```diff
@@ -1,32 +1,35 @@
 {% macro clear_tests() %}
     {% if execute %}
-        {% do drop_schema(elementary.target_database(), target.schema) %}
+
```
GuyEshdat, line 3: i guess that was the line of the debug :) remove

integration_tests/deprecated_tests/macros/e2e_tests/clear_tests.sql
```diff
@@ -1,32 +1,35 @@
 {% macro clear_tests() %}
     {% if execute %}
-        {% do drop_schema(elementary.target_database(), target.schema) %}
+    
+        {% set database_name = elementary.target_database() %}
+        {% set schema_name = target.schema %}
```
GuyEshdat, line 5: the schema name and database_name can be inline with the macro call, like it was before

GuyEshdat (approved):

PR #828, opened 2025-07-27
run_query: add metadata comment to all queries run with elementary.run.query

haritamar:


macros/utils/run_queries/run_query.sql
```diff
@@ -1,5 +1,11 @@
 {% macro run_query(query, lowercase_column_names=True) %}
-    {% set query_result = dbt.run_query(query) %}
+    {% set query_with_metadata %}
+        {{ query }}
+        /* --ELEMENTARY-METADATA-- {{ elementary.get_elementary_query_metadata() | tojson }} --END-ELEMENTARY-METADATA-- */
```
haritamar: Adding the comment after and not before, because Snowflake seems to remove comments before the query.

MikaKerman (approved):

PR #851, opened 2025-09-02
Fix datetime casting macro for ClickHouse

StevenReitsma:
Previously, when a `NULL` value was encountered in the `timestamp_field`, the `toString` function would convert this NULL value to a `Nullable(String)`. This would result in the split function not working properly, since it expects non-nullable input values. The datetime parsing can be simplified by using `parseDateTimeBestEffortOrNull` instead of doing manual parsing of the date, bypassing the issue of `Nullable` types entirely.

## Summary by CodeRabbit

- Bug Fixes
  - Improved timestamp parsing for ClickHouse sources: now uses best-effort, UTC-aware parsing to handle varied or unexpected formats and reduce ingestion failures.
  - Ensures a safe fallback to the Unix epoch in UTC when parsing fails.

- Other
  - No changes to public interfaces or configuration required.

### DIFF
--- a/macros/utils/data_types/cast_column.sql
+++ b/macros/utils/data_types/cast_column.sql
@@ -23,8 +23,7 @@
 
 {%- macro clickhouse__edr_cast_as_timestamp(timestamp_field) -%}
     coalesce(
-        toDateTime(
-            left(splitByChar('+', toString({{ timestamp_field }}))[1], 19)),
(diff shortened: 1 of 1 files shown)


michael-myaskovsky (approved):

michael-myaskovsky:
Approved! Thank you for your contribution!

PR #858, opened 2025-09-09
removed render_pure() when deleting table since quoting is needed

arbiv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of test table cleanup in Athena environments, ensuring DROP statements target the correct tables and preventing leftover artifacts or teardown failures during test runs.


arbiv:
This results in " which athena doesn't like.
The right solution is https://github.com/elementary-data/dbt-data-reliability/pull/859

PR #867, opened 2025-10-07
Ele 5120 fusion databricks bugfix

haritamar:
## Summary by CodeRabbit

* **New Features**
  * Improved compatibility with dbt-fusion on Databricks when creating temporary views, enabling the adapter to use fusion-specific view handling.
* **Bug Fixes**
  * Fixed temporary relation creation on Databricks under dbt-fusion to reduce workflow failures.
* **Chores**
  * Streamlined CI matrix by removing a forced Postgres + specific dbt version combo, aligning tests with latest/pre-release configurations.

### DIFF
--- a/.github/workflows/test-all-warehouses.yml
+++ b/.github/workflows/test-all-warehouses.yml
@@ -51,9 +51,6 @@ jobs:
             dremio,
           ]
         include:
-          # If we're not running on a specific dbt version, then always add postgres on 1.8.0
-          - dbt-version: "${{ inputs.dbt-version || '1.8.0' }}"
-            warehouse-type: postgres
           - dbt-version: "${{ inputs.dbt-version || 'latest_pre' }}"
             warehouse-type: postgres
           - dbt-version: "${{ inputs.dbt-version || 'fusion' }}"
--- a/macros/utils/table_operations/make_temp_relation.sql
+++ b/macros/utils/table_operations/make_temp_relation.sql
@@ -18,7 +18,12 @@
 
 {% macro databricks__edr_make_temp_relation(base_relation, suffix) %}
     {% set tmp_identifier = elementary.table_name_with_suffix(base_relation.identifier, suffix) %}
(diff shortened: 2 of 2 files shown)


.github/workflows/test-all-warehouses.yml
```diff
@@ -51,9 +51,6 @@ jobs:
             dremio,
           ]
         include:
-          # If we're not running on a specific dbt version, then always add postgres on 1.8.0
-          - dbt-version: "${{ inputs.dbt-version || '1.8.0' }}"
-            warehouse-type: postgres
```
haritamar, line 56: The tests are no longer compatible with `dbt<1.10.8` (though the package itself still is)

macros/utils/table_operations/make_temp_relation.sql
```diff
@@ -18,7 +18,12 @@
 
 {% macro databricks__edr_make_temp_relation(base_relation, suffix) %}
     {% set tmp_identifier = elementary.table_name_with_suffix(base_relation.identifier, suffix) %}
-    {% set tmp_relation = api.Relation.create(identifier=tmp_identifier, type='view') %}
+    {% if elementary.is_dbt_fusion() %}
+        {# dbt-fusion databricks adapter doesn't support temp views, for consistency we still use views but regular ones #}
+        {% set tmp_relation = base_relation.incorporate(path={"identifier": tmp_identifier}, type='view') %}
```
haritamar, line 24: Perhaps to clarify further, the previous bug was that in dbt-fusion even though the views are not temp, we created them without specifying the schema explicitly.
This was a problem in the tests because they use `generate_schema_name` and the "default" schema did not exist when running the tests in parallel.

ofek1weiss: so, from what i understand, the comment in the code is incorrect

both version seem to create regular views

ofek1weiss (approved):

PR #881, opened 2025-11-05
Raize error if columns can't be detected

kokorin:
When dbt user has no permissions for a table, Elementary `schema_changes_from_baseline` test fails with unclear message (at least in Snowflake):

```
001003 (42000): SQL compilation error:
syntax error line 44 at position 12 unexpected ')'.
```

It happens because dbt swallows DB error in `adapter.get_columns_in_relation()` when a user has not enough permissions.

It our case it was hard to detect because of Snowflake Secondary roles, which allowed human users to read from source DB, but not for CI users.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced schema monitoring validation to detect when model attributes cannot be properly identified, providing clear diagnostic messages instead of silent failures.

### DIFF
--- a/macros/edr/data_monitoring/schema_changes/get_columns_snapshot_query.sql
+++ b/macros/edr/data_monitoring/schema_changes/get_columns_snapshot_query.sql
@@ -14,6 +14,13 @@
 
     {% set columns = adapter.get_columns_in_relation(model_relation) %}
 
+    {% if not columns %}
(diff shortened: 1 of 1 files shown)


NoyaOffer:
Thanks @kokorin , this looks great!
I'll have one of our engineers check it out.

arbiv:
approved, pending test

PR #981, opened 2026-04-09
fix: allow _with_context tests to collect sample rows

joostboon:
## Summary
- `_with_context` tests are Elementary-namespaced but behave like regular dbt tests (they return failing rows rather than handling their own result row collection)
- The test materialization was early-returning for all `elementary`-namespaced tests, bypassing the sampling logic entirely
- This meant `_with_context` tests never wrote rows to `test_result_rows`, even when tagged with `show_sample_rows`

## Fix
Check if the test name ends with `_with_context` before early-returning, allowing these tests to go through the standard `handle_dbt_test` sampling path.

## Test plan
- [ ] Run a failing `_with_context` test (e.g. `not_null_with_context`) and verify rows appear in `test_result_rows`
- [ ] Verify non-`_with_context` Elementary tests (anomaly, schema change) still early-return and handle their own result rows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of elementary tests by applying specialized processing only to recognized test types.
  * Ensured other elementary-namespaced tests, including context-aware tests, follow the standard processing flow.
  * Fixed sampling for `not_null_with_context` tests so configured sample counts are respected and context values are preserved in sampled results.

### DIFF
--- a/macros/edr/materializations/test/test.sql
+++ b/macros/edr/materializations/test/test.sql
@@ -28,8 +28,11 @@
     {% endif %}
 
     {% set test_namespace = model.get("test_metadata", {}).get("namespace") %}
-    {% if test_namespace == "elementary" %}
-        {# Custom test materialization is needed only for non-elementary tests #}
+    {% set test_name = model.get("test_metadata", {}).get("name", "") %}
+    {% if test_namespace == "elementary" and not test_name.endswith("_with_context") %}
+        {# Custom test materialization is needed only for non-elementary tests.
+           _with_context tests are elementary-namespaced but behave like regular dbt
+           tests (they return failing rows) so they go through the standard sampling path. #}
         {% do return(materialization_macro()) %}
     {% endif %}
 
--- a/macros/edr/materializations/test/test.sql
+++ b/macros/edr/materializations/test/test.sql
@@ -28,12 +28,15 @@
     {% endif %}
 
     {% set test_namespace = model.get("test_metadata", {}).get("namespace") %}
-    {% set test_name = model.get("test_metadata", {}).get("name", "") %}
-    {% if test_namespace == "elementary" and not test_name.endswith("_with_context") %}
-        {# Custom test materialization is needed only for non-elementary tests.
-           _with_context tests are elementary-namespaced but behave like regular dbt
-           tests (they return failing rows) so they go through the standard sampling path. #}
-        {% do return(materialization_macro()) %}
+    {% if test_namespace == "elementary" %}
+        {% set short_name = model.get("test_metadata", {}).get("name", "") | lower %}
+        {% set elementary_test_type = elementary.get_elementary_test_type(
+            {"short_name": short_name, "test_namespace": "elementary"}
+        ) %}
+        {% if elementary_test_type %}
+            {# Anomaly detection and schema change tests handle their own result row collection #}
+            {% do return(materialization_macro()) %}
+        {% endif %}
     {% endif %}
 
     {% set flattened_test = elementary.flatten_test(model) %}
--- a/macros/edr/materializations/test/test.sql
+++ b/macros/edr/materializations/test/test.sql
@@ -34,9 +34,7 @@
             {"short_name": short_name, "test_namespace": "elementary"}
         ) %}
         {% if elementary_test_type and elementary_test_type != "with_context" %}
-            {# Anomaly detection and schema change tests handle their own result row collection.
-               with_context tests do not -- they return the failing rows and need this
-               materialization to sample them -- so they are exempted explicitly. #}
+            {# Anomaly detection and schema change tests handle their own result row collection #}
             {% do return(materialization_macro()) %}
         {% endif %}
     {% endif %}
--- a/macros/edr/materializations/test/test.sql
+++ b/macros/edr/materializations/test/test.sql
@@ -33,8 +33,10 @@
         {% set elementary_test_type = elementary.get_elementary_test_type(
             {"short_name": short_name, "test_namespace": "elementary"}
         ) %}
-        {% if elementary_test_type %}
-            {# Anomaly detection and schema change tests handle their own result row collection #}
+        {% if elementary_test_type and elementary_test_type != "with_context" %}
+            {# Anomaly detection and schema change tests handle their own result row collection.
+               with_context tests do not -- they return the failing rows and need this
+               materialization to sample them -- so they are exempted explicitly. #}
             {% do return(materialization_macro()) %}
         {% endif %}
     {% endif %}
--- a/integration_tests/tests/test_with_context_sampling.py
+++ b/integration_tests/tests/test_with_context_sampling.py
@@ -1,0 +1,46 @@
+import json
+
+from dbt_project import DbtProject
+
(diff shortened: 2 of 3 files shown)


joostboon:
Reopening — this was closed by stale-bot on 2026-07-10 after no human review, not on its merits. Added the integration test the contribution bot asked for, and independently re-verified the bug against a live Snowflake warehouse.

### The bug is in every release that has the feature

The namespace early-return landed in `0.23.0`; the `_with_context` tests landed in `0.24.0`. So they have never collected sample rows in any released version.

| version | `_with_context` tests | namespace early-return |
|---|---|---|
| 0.22.0 | absent | no |
| 0.23.0 | absent | **yes** |
| 0.24.0 → 0.25.1 | present | **yes** |

### Verified end to end

elementary `0.25.1`, dbt-snowflake `1.11.1`, using `elementary.not_null_with_context` exactly as documented on a column with 55 nulls and five `context_columns`:

| | sample rows in `test_result_rows` | `failed_row_count` |
|---|---|---|
| stock 0.25.1 | **0** | `null` |
| with this branch | **5** | **55** |

Note `failed_row_count` was also null before, since `calculate_failed_count` sits after the early return — so these tests were skipping that too.

A useful corroborating detail: the package already classifies these as regular dbt tests. `get_elementary_test_type` returns a type only for the anomaly and schema-change families, so `get_test_type` returns `dbt_test` for `_with_context` — which is what lands in `elementary_test_results.test_type`. The materialization branching on namespace instead of test type was inconsistent with the package's own classification, which is what this fixes.

### What I added

`integration_tests/tests/test_with_context_sampling.py` — asserts an elementary-namespaced `_with_context` test writes `test_sample_row_count` rows and that the requested context column travels with each failing row. There were no integration tests covering the `_with_context` family at all, which is why this shipped unnoticed. Modelled on `test_sampling.py` and uses the same harness calls. It fails against the unfixed materialization.

I have not run the integration suite myself — no warehouse credentials for the harness — so CI is the first real execution of that test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

joostboon:
CI is red, but none of it is this PR. Breakdown:

| job | cause | related? |
|---|---|---|
| `latest_official, snowflake` | `Database Error 251006: Password is empty` — failed at `dbt debug`, before any test ran | no |
| `fusion, snowflake` | `[Snowflake] 260002: password is empty` | no |
| `latest_official, athena` | `InvalidClientTokenId` / `UnrecognizedClientException` — expired AWS credentials | no |
| `latest_official, sqlserver` | errors in `test_exposure_schema_validity.py` | no |
| `fusion, bigquery` | `FAILED test_exposure_schema_validity_no_exposures` | no |

Two independent pre-existing issues: **stale CI credentials** for Snowflake and Athena (expected on a branch idle since April), and **`test_exposure_schema_validity.py` failing** on sqlserver and bigquery-fusion.

The new test passed on every adapter that could connect:

```
postgres  [gw2] [ 98%] PASSED test_with_context_sampling.py::test_with_context_test_is_sampled
duckdb    [gw6] [ 98%] PASSED test_with_context_sampling.py::test_with_context_test_is_sampled
redshift  [gw0] [ 98%] PASSED test_with_context_sampling.py::test_with_context_test_is_sampled
```

plus the other green adapters (fabric, dremio, vertica, clickhouse, databricks_catalog, latest_pre/postgres).

Snowflake CI couldn't run, but that is the adapter I verified the fix on by hand — elementary 0.25.1 / dbt-snowflake 1.11.1, `elementary.not_null_with_context` on a column with 55 nulls: 0 sample rows before, 5 after, with every requested context column attached, and `failed_row_count` going from `null` to 55.

Someone with access will need to refresh the Snowflake and Athena CI secrets to get a fully green run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

joostboon:
Both CodeRabbit suggestions applied in 4b176a86 — both were right, and the first is more load-bearing than it looks.

**`failed_row_count`** — applied. Worth noting *why* it matters here rather than being a generic nicety: `calculate_failed_count` sits after the same early return the sampling did, so it was skipped too. Measured on Snowflake with 0.25.1, `elementary.not_null_with_context` on a column with 55 nulls:

| | sample rows | `failed_row_count` |
|---|---|---|
| unfixed | 0 | `null` |
| fixed | 5 | 55 |

So this assertion covers a second broken path, not just the samples. Consistent with `test_failed_row_count.py`, which asserts the same key the same way.

**Exact context values** — applied. Used set membership rather than equality, since sampling returns an arbitrary `test_sample_row_count` of the `null_count` failing rows, so no single expected value would hold. This does prove the materialization preserved the source value, which the `startswith` check did not.

Also merged current `master` into the branch (68 commits behind). No conflicts, and the gate survived intact — net diff against `master` is still just `test.sql` (+10/-2) and the new test. I verified the harness API this test depends on (`dbt_project.test`, `samples_query`, `run_query`) is unchanged across those commits.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

macros/edr/materializations/test/test.sql
```diff
@@ -29,8 +29,14 @@
 
     {% set test_namespace = model.get("test_metadata", {}).get("namespace") %}
     {% if test_namespace == "elementary" %}
-        {# Custom test materialization is needed only for non-elementary tests #}
-        {% do return(materialization_macro()) %}
+        {% set short_name = model.get("test_metadata", {}).get("name", "") | lower %}
+        {% set elementary_test_type = elementary.get_elementary_test_type(
+            {"short_name": short_name, "test_namespace": "elementary"}
+        ) %}
+        {% if elementary_test_type %}
```
haritamar: Can we perhaps instead be more explicit, and return a type in `get_elementary_test_type` for the relevant tests, and then check here that it's not that type?

joostboon: Done in c33465fc — agreed, explicit is better. `get_elementary_test_type` now returns a `"with_context"` type for the six tests in the family, and the materialization checks `elementary_test_type != "with_context"` instead of leaning on the macro returning `none`. It also means a future elementary test family defaults to the existing skip behaviour rather than silently landing in this bug again.

One thing worth knowing, since it caught me while implementing it: the new type must **not** reach `test_type`. All three alerts models partition on that value exactly —

- `alerts_dbt_tests.sql:41` → `test_type = 'dbt_test'`
- `alerts_anomaly_detection.sql:41` → `test_type = 'anomaly_detection'`
- `alerts_schema_changes.sql:47` → `test_type = 'schema_change'`

so a fourth value would match none of them and `with_context` test failures would raise **no alerts at all**. `get_test_type` therefore maps `with_context` → `dbt_test`, keeping the reported type and the alerts behaviour exactly as before; the new type stays internal to the materialization gate. I've added `assert test_result["test_type"] == "dbt_test"` to the integration test so that can't regress silently.

Happy to change the type name, or to move the exemption list somewhere more visible than `get_elementary_test_type`, if you'd prefer it shaped differently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

haritamar (approved):

Issue #986, opened 2026-04-15
Would it be possible to add deprecation_date to dbt_models metadata model?


scrambldchannel:
As of dbt core 1.6 [(changelog](https://github.com/dbt-labs/dbt-core/issues/7433) / [github issue)](https://github.com/dbt-labs/dbt-core/issues/7433
), adding the top level key `deprecation_date` to a model's properties is supported as documented here:

https://docs.getdbt.com/reference/resource-properties/deprecation_date?version=1.12

We make use of this key and it would be great if it could be added to the `dbt_models` metadata model. I've had a quick look at the code and think I can see how this could be achieved and will create a draft PR to see if the idea makes sense.

PR #987, opened 2026-04-15
feat; add deprecation_date key to dbt_models metadata model

scrambldchannel:
## What

This is a quick attempt to address https://github.com/elementary-data/dbt-data-reliability/issues/986 by adding the column to the `upload_dbt_models` macro that populates the `dbt_models` model. 

I've also added the new column, and several other missing columns, the properties for the model.

## Why

We, and presumably others, leverage this column as a way to flag upcoming model deprecation to developers and also to automatically remove models from our project. If we could have this column in `dbt_models`, it would make analysis, and automated notifications to users easier. As far as I'm aware, there isn't a way to currently do this and I don't see a record of this being discussed previously.

## TODO / Questions

~~I still need to do some testing but thought I might as well create a draft PR to accompany the issue I raised in case this is a non starter.~~

I've tested this locally and the resulting `deprecation_date` column picks up the value as expected. I've also run the integration tests against a postgres target successfully and will attach a log file.

I will do some functional testing against Snowflake using a branch in my repo too.

~~I also need to think about the appropriate data type to use for this column~~

I think leaving this as is the right approach but happy to be contradicted

## Summary by CodeRabbit

* **New Features**
  * Extended the dbt_models artifact table with seven new metadata fields: `deprecation_date`, `patch_path`, `metadata_hash`, `unique_key`, `incremental_strategy`, `group_name`, and `access`. These additions enhance model metadata tracking and capture additional configuration properties from dbt graph sources.

### DIFF
--- a/macros/edr/dbt_artifacts/upload_dbt_models.sql
+++ b/macros/edr/dbt_artifacts/upload_dbt_models.sql
@@ -40,6 +40,7 @@
         ("incremental_strategy", "string"),
         ("group_name", "string"),
         ("access", "string"),
+        ("deprecation_date", "string"),
     ] %}
     {% if target.type == "bigquery" or elementary.get_config_var(
         "include_other_warehouse_specific_columns"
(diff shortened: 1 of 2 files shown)


scrambldchannel:
[postgres_test_log.txt](https://github.com/user-attachments/files/26763686/postgres_test_log.txt)

PR #1012, opened 2026-05-22
Support BigQuery nested STRUCT fields in anomaly tests

tlangton3:
Allows `column_anomalies` and `dimension_anomalies` to reference nested STRUCT leaves on BigQuery (e.g. `user.address.city`) instead of only top-level columns.

A single column-discovery wrapper segment-quotes nested references (`` `a`.`b`.`c` ``) and projects the monitored column with a dot-free CTE alias so the path survives into downstream aggregates. Non-nested columns and non-BigQuery adapters compile byte-identically to today's behaviour. REPEATED ancestors are out of scope (would require `UNNEST`). `test_all_columns_anomalies` is unchanged — users opt in by passing `column_name=user.address.city` explicitly to avoid ballooning the test surface on wide STRUCT schemas.

### What changes

- `get_column_obj_and_monitors` flattens BigQuery STRUCT columns via `BigQueryColumn.flatten()`, filtered through an ancestor-aware walker (`bq_safe_leaf_names`) so leaves under REPEATED ancestors are excluded. Each discovered column is wrapped with a dict carrying `.name` (dotted display form), `.quoted` (segment-quoted SQL ref), `.safe_alias` (dot-free identifier) and `.is_nested`. Top-level STRUCTs are kept alongside their leaves so existing `column_name=user` behaviour is preserved.
- `column_monitoring_query` projects nested columns as `<quoted> as <adapter-quoted safe_alias>` and references the quoted alias in metric aggregates. Non-nested columns keep today's projection and `.quoted` references untouched, so identifier quoting (reserved words, case-sensitive names) is never lost.
- `bq_is_nested_identifier` matches only plain dotted identifier paths (`^\w+(\.\w+)+$`) on BigQuery. Since `dimensions` accepts arbitrary SQL expressions, segment-quoting in `dimension_monitoring_query` and `select_dimensions_columns` is gated on this predicate — expressions and plain identifiers pass through byte-identically to master.

### Why two representations

`BigQueryColumn.quoted` wraps the whole string in one set of backticks, so a flattened nested column's `.quoted` is `` `user.address.city` `` — which BigQuery treats as a single column literally named `user.address.city`. Even with correct segment-quoting, projecting `select user.address.city from t` into a CTE without an alias names the resulting column `city`, losing the path. The wrapper exposes both `.quoted` (segment-quoted source ref) and `.safe_alias` (dot-free CTE alias) so the projection-alias pattern composes cleanly and downstream macros stay nesting-agnostic. The alias is only emitted when `.is_nested` is true; otherwise `safe_alias` mirrors `.quoted`.

### Testing

Local validation via `dbt parse` and a `run-operation` harness against the BigQuery adapter confirmed every SQL fingerprint:

- Segment-quoting: `user.address.city` → `` `user`.`address`.`city` ``
- Projection: `` select `user`.`address`.`city` as `user__address__city` from t ``; non-nested projection byte-identical to master
- Downstream aggregate references `` `user__address__city` `` when nested, `.quoted` otherwise (byte-identical to master)
- Expression dimensions (`case when amount > 100 then 'high' end`) and dotted expressions (`coalesce(user.a, user.b)`) pass through unchanged
- Stored `column_name`: `user.address.city` (dotted display preserved for alerts)
- `get_column_data_type` BigQuery dispatch works on the wrapped dict via subscript access

End-to-end execution against BigQuery to follow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved monitoring for nested BigQuery STRUCT fields, including dotted column names and nested dimensions.
  * Improved handling of repeated fields and nested leaf values.
  * Added safer column and dimension aliases for complex names in monitoring results.

* **Tests**
  * Added integration coverage for stable and anomalous nested-column and nested-dimension scenarios.

### DIFF
--- a/macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
@@ -412,6 +412,10 @@
         {%- set quoted = column_obj.quoted -%}
         {%- set safe_alias = column_obj.column -%}
     {%- endif -%}
+    {# `fields` only exists on BigQueryColumn; guard so non-BigQuery
+       adapters (Snowflake, Postgres, Redshift, ...) don't trip on the
+       attribute access. #}
+    {%- set fields = column_obj.fields if column_obj.fields is defined else [] -%}
     {{ return({
         'name': name,
         'column': column_obj.column,
@@ -419,6 +423,44 @@
         'safe_alias': safe_alias,
         'dtype': column_obj.dtype,
         'data_type': column_obj.data_type,
-        'fields': column_obj.fields,
+        'fields': fields,
     }) }}
+{% endmacro %}
+
+{# Walk a BigQuery STRUCT tree and collect dotted leaf names that are safe to
+   monitor without UNNEST — i.e. no REPEATED ancestor anywhere in the path,
+   and the leaf itself is not REPEATED. `BigQueryColumn.flatten()` returns leaf
+   columns with the leaf's own mode but discards ancestor modes, so this walker
+   is the source of truth for "which leaves can we project directly?". #}
+{% macro bq_safe_leaf_names(column_obj) %}
+    {%- set safe_names = [] -%}
+    {%- if column_obj.mode != 'REPEATED'
+            and column_obj.fields is defined
+            and column_obj.fields | length > 0 -%}
+        {%- for child in column_obj.fields -%}
+            {%- do elementary._bq_walk_collect(
+                child, [column_obj.column], false, safe_names
+            ) -%}
+        {%- endfor -%}
+    {%- endif -%}
+    {{ return(safe_names) }}
+{% endmacro %}
+
+{# Recursive helper: walks a google.cloud.bigquery.SchemaField subtree,
+   propagating whether any ancestor was REPEATED. Append safe leaf names to
+   `safe_names`. #}
+{% macro _bq_walk_collect(field, prefix, has_repeated_ancestor, safe_names) %}
+    {%- set new_prefix = prefix + [field.name] -%}
+    {%- if field.fields | length == 0 -%}
+        {%- if not has_repeated_ancestor and field.mode != 'REPEATED' -%}
+            {%- do safe_names.append(new_prefix | join('.')) -%}
+        {%- endif -%}
+    {%- else -%}
+        {%- set new_has_repeated = has_repeated_ancestor or (field.mode == 'REPEATED') -%}
+        {%- for child in field.fields -%}
+            {%- do elementary._bq_walk_collect(
+                child, new_prefix, new_has_repeated, safe_names
+            ) -%}
+        {%- endfor -%}
+    {%- endif -%}
 {% endmacro %}
--- a/macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
+++ b/macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
@@ -15,7 +15,9 @@
     {%- set timestamp_column = metric_properties.timestamp_column %}
     {% set prefixed_dimensions = [] %}
     {% for dimension_column in dimensions %}
-        {% do prefixed_dimensions.append("dimension_" ~ dimension_column) %}
+        {% do prefixed_dimensions.append(
+            "dimension_" ~ elementary.bq_safe_alias(dimension_column)
+        ) %}
     {% endfor %}
 
     {% set metric_types = [] %}
@@ -53,7 +55,7 @@
             ),
             filtered_monitored_table as (
                 select
-                    {{ column_obj.quoted }},
+                    {{ column_obj.quoted }} as {{ column_obj.safe_alias }},
                     {%- if dimensions -%}
                         {{
                             elementary.select_dimensions_columns(
@@ -78,7 +80,7 @@
         {%- else %}
             filtered_monitored_table as (
                 select
-                    {{ column_obj.quoted }},
+                    {{ column_obj.quoted }} as {{ column_obj.safe_alias }},
                     {%- if dimensions -%}
                         {{
                             elementary.select_dimensions_columns(
@@ -94,7 +96,7 @@
         column_metrics as (
 
             {%- if column_metrics %}
-                {%- set column = column_obj.quoted -%}
+                {%- set column = column_obj.safe_alias -%}
                 select
                     {%- if timestamp_column %}
                         edr_bucket_start as bucket_start, edr_bucket_end as bucket_end,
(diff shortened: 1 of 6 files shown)


tlangton3:
End-to-end validated against a real BigQuery dataset.

- `column_anomalies` on a three-level nested STRUCT field (`<parent>.<intermediate>.<leaf>`) compiles with segment-quoted SQL, executes against real data, and writes a row to `data_monitoring_metrics` with the dotted `column_name` preserved.
- Discovery layer correctly flattens parent STRUCTs via `BigQueryColumn.flatten()`; the wrapper exposes `.name` (dotted display), `.quoted` (segment-quoted SQL ref), and `.safe_alias` (dot-free CTE alias) as designed.
- Ran the new nested test alongside 10+ existing non-nested `column_anomalies` tests in a single `dbt test` invocation — all 15 PASS with no interference, confirming the projection-alias pattern is backwards-compatible.
- Re-ran with `--defer --favor-state` against a prod manifest so the non-nested tests had data and history; metrics for nested and non-nested columns land in `data_monitoring_metrics` and `elementary_test_results` with identical schema. The dotted `column_name` is just a longer string in an otherwise unchanged structure.
- `elementary.on_run_end` upload hook works unchanged with the override — metric history persists correctly.

Tested against:
- dbt-core 1.11.8 / dbt-bigquery 1.11.1
- elementary package version 0.23.x (this branch)

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
```diff
@@ -341,17 +347,135 @@
...
+      {%- if not loop.last -%}{{ ", " }}{%- endif -%}
     {%- endfor -%}
     {% endset %}
     {{ return(select_statements) }}
 {% endmacro %}
+
+
+{# ---------------------------------------------------------------------- #}
+{# BigQuery STRUCT nested-field helpers.                                  #}
+{# ---------------------------------------------------------------------- #}
```
haritamar: I think all the utilities here should not be in this file, but in dedicated files

macros/edr/data_monitoring/monitors_query/column_monitoring_query.sql
```diff
@@ -15,7 +15,13 @@
     {%- set timestamp_column = metric_properties.timestamp_column %}
```
haritamar, line 15: General note - unless very complicated, we need integration tests of column + dimension tests with struct fields.

haritamar:
Thanks for this @tlangton3 — the approach is sound and the REPEATED-ancestor walker in particular is a nice catch. I've pushed three commits on top of your branch rather than leaving a long list of review comments; happy to revert any of it if you disagree.

_(This review and the follow-up commits were generated by Claude Code.)_

### What changed

**1. Helpers moved to a dedicated file** (`macros/utils/sql_utils/bigquery_nested_columns.sql`) — following up on my inline note. `column_monitoring_query.sql` shouldn't own utilities that `dimension_monitoring_query.sql` and `get_column_monitors.sql` also call.

**2. `wrap_column_for_struct_support` removed.** This was the main concern. It replaced the adapter `Column` object with a plain dict for *every* adapter, not just BigQuery, which silently drops `char_size`, `numeric_precision`, `numeric_scale`, `mode` and every `Column` method (`is_string()`, `is_numeric()`, …). Nothing breaks today because the only consumers read `.name`/`.quoted`/dtype, but it changes the contract of a macro shared by ~12 adapters and the next person to write `column_obj.is_string()` gets a silent `Undefined`. It also eagerly evaluated `.data_type` on adapters where only `.dtype` is ever read.

`column_monitoring_query` now derives what it needs from `column_obj.name` directly, so `column_obj` stays a real `Column`:

```jinja
{%- if elementary.bq_is_nested_identifier(column_obj.name) %}
    {%- set nested_alias = adapter.quote(elementary.bq_safe_alias(column_obj.name)) %}
    {%- set monitored_column_projection = elementary.bq_segment_quote(column_obj.name) ~ " as " ~ nested_alias %}
    {%- set monitored_column_expr = nested_alias %}
{%- else %}
    {%- set monitored_column_projection = column_obj.quoted %}
    {%- set monitored_column_expr = column_obj.quoted %}
{%- endif %}
```

This also removed the `safe_alias` → `.quoted` fallback, which would have produced `` `\`col\`` `` if anything had ever used it unguarded.

**3. Smaller cleanups**
- `bq_alias_safe_dimension` shared by `prefixed_dimensions` and `select_dimensions_columns`, instead of the same `is_nested ? safe_alias : column` ternary in both places, and instead of `wrap_column_for_struct_support` re-implementing `bq_segment_quote`/`bq_safe_alias` inline.
- `bq_flatten_nested_columns` owns the flatten loop, so `get_column_monitors.sql` is a two-line call. The pass is now skipped unless `column_name` is dotted — no point walking wide STRUCT schemas to look up an ordinary column.
- `_bq_walk_collect`'s comment said it walks `google.cloud.bigquery.SchemaField`; `BigQueryColumn.fields` is actually a list of `BigQueryColumn` (via `wrap_subfields`). Same duck-typed attributes, so the code was fine — just the comment.
- **`bq_is_nested_identifier` regex tightened** to `^[A-Za-z_]\w*(\.[A-Za-z_]\w*)+$`. `^\w+(\.\w+)+$` also matches bare decimals, so a dimension of `0.99` would have been rewritten to `` `0`.`99` `` — a syntax error, and a behaviour change from master since `dimension_monitoring_query` now applies `bq_segment_quote` to every dimension on BigQuery.

Net effect on the three files you touched: **+56/-11** instead of +170/-15.

### Integration tests

Added `integration_tests/tests/test_nested_struct_anomalies.py` — six tests, `@pytest.mark.only_on_targets(["bigquery"])`, so they run in the BigQuery CI job.

Structs can't be seeded from CSV, so they build a real STRUCT table via `create_temp_model_for_existing_table(raw_code=…)` + `dbt run`, then use the existing `as_model=True` path. The fixture deliberately includes a `REPEATED` struct (`orders`) and a `REPEATED` scalar (`tags`) so discovery has to skip them.

| Test | Covers |
|---|---|
| `test_anomalyless_column_anomalies_on_struct_field` | nested column passes; asserts the stored `column_name` is the dotted path |
| `test_anomalous_column_anomalies_on_struct_field` | null spike on a nested leaf → fail |
| `test_column_anomalies_with_struct_dimension` | plain column + nested dimension; asserts `dimension_value` is `{Metropolis, Gotham}` |
| `test_column_anomalies_on_struct_field_with_struct_dimension` | nested column **and** nested dimension |
| `test_anomalyless_dimension_anomalies_on_struct_field` | `dimension_anomalies` on a nested leaf |
| `test_anomalous_dimension_anomalies_on_struct_field` | anomalous dimension; asserts the anomalous `dimension_value` |

The `dimension_value` assertions matter — without them, a query that silently resolved the dimension to NULL would still report "pass".

### Verification

`dbt parse --target bigquery` is clean, and I rendered the actual Jinja (the `{%- if elementary.bq_is_nested_identifier(...) %}` block extracted from the file, `select_dimensions_columns`, and the flatten walker) against a real `BigQueryColumn` tree. Every fragment is byte-identical to what your version emitted, and the non-nested cases are byte-identical to master:

```
flatten():          ['user_info.address.city', 'user_info.address.country', 'user_info.name',
                     'user_info.orders.amount', 'user_info.tags']
bq_safe_leaf_names: ['user_info.address.city', 'user_info.address.country', 'user_info.name']

nested  -> `user_info`.`address`.`city` as `user_info__address__city`   (aggregates ref `user_info__address__city`)
plain   -> `superhero`      (unchanged from master)
struct  -> `user_info`      (unchanged from master)
dim     -> `user_info`.`address`.`city` as dimension_user_info__address__city
expr    -> case when amount > 100 then 'high' end as dimension_case when …   (unchanged from master)
```

I could not run the integration tests locally — my BigQuery credentials are stale — so CI is the first real execution. I'll approve the workflow run.

### One open question

Is the `bq_segment_quote` in `dimension_monitoring_query` load-bearing, i.e. did you hit an actual failure without it? Plain struct paths resolve fine unquoted in that position, so as far as I can tell the only thing it genuinely buys is path segments that *need* quoting — a field named `order`, `from`, `hash`, etc. That's a real case and a good reason to keep it, I'd just like to confirm that's the intent rather than defensive quoting. (It doesn't help with the one genuine ambiguity — a table alias colliding with a struct column name — since range variables win over columns regardless of backticks.)

Docs: skipping for now, our docs source of truth is on a separate branch that has diverged. Tracking that separately.

tlangton3:
Hi @haritamar

Thanks for the review!

No, I never hit a runtime failure in that position -> plain struct paths resolve fine unquoted there, as you say. The intent was reserved-word segments (user_info.order fails to compile without backticks) plus keeping every nested-path site behind the same bq_is_nested_identifier predicate rather than special-casing this one. So: deliberate, but defensive rather than fixing an observed bug.
  
Agreed on the range-variable caveat, backticks can't win against a table alias shadowing the first segment, and that quirk predates this PR.
  
Happy to make it load-bearing-by-test rather than by argument: I can add a reserved-word leaf (e.g. order) to the integration fixture and use it as a dimension, which pins the behaviour for both the dimension and column paths.

macros/utils/sql_utils/bigquery_nested_columns.sql
```diff
@@ -0,0 +1,126 @@
...
+{% endmacro %}
+
+{% macro bq_safe_alias(name) %}
+    {#- Convert a dotted identifier path into a dot-free SQL identifier.
+        Projecting `select user.address.city from t` into a CTE without an alias
+        names the resulting column `city`, losing the path, so nested columns
+        must be aliased on the way in. Only call this for names that satisfy
+        `bq_is_nested_identifier` — on arbitrary SQL expressions it produces
+        nonsense. -#}
+    {{- name | replace(".", "__") -}}
```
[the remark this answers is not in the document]
tlangton3: Deliberately leaving this as-is: a collision requires two dimensions on the same test whose segments themselves contain `__` (e.g. `a.b__c` + `a__b.c`), and BigQuery then rejects the duplicate CTE column names loudly rather than producing wrong data. The length-prefixed encoding makes every nested alias unreadable to defend against that. Documented the limitation on the macro in f6eb2f0.

haritamar (approved):
Thanks a lot for the contribution! Sorry it took us a long time to get to.
I did a couple of fixes before merging (mostly avoiding calling BQ-specific macros in generic cross DWH ones)

PR #1019, opened 2026-06-02
feat: add `elementary_extra_indexes` config var for custom index injection

haritamar:
## Summary

Adds a new `elementary_extra_indexes` dbt var (default `{}`) that lets downstream consumers inject additional PostgreSQL indexes into Elementary models without modifying this package.

**New macro**: `elementary.get_indexes_for_model(model_name, base_indexes)` — merges the model's built-in indexes with any extras from `elementary_extra_indexes`, deduplicating by sorted column set. Returns `[]` on non-postgres targets.

**Usage** — pass a dict mapping model names to index configs via `--vars`:

```yaml
elementary_extra_indexes:
  dbt_run_results:
    - columns: [\"created_at\"]
  elementary_test_results:
    - columns: [\"created_at\"]
```

All 5 models with postgres indexes now use `get_indexes_for_model`: `dbt_run_results`, `elementary_test_results`, `data_monitoring_metrics`, `dbt_source_freshness_results`, `test_result_rows`.

**Motivation**: Elementary Cloud needs cloud-specific indexes (e.g. `(created_at)` for time-range queries) that don't belong in the OSS defaults. This avoids polluting the base package while keeping index definitions co-located with models."

Link to Devin session: https://app.devin.ai/sessions/278ceddbe6d24dbea36a83ecc232f3b6
Requested by: @haritamar

## Summary by CodeRabbit

* **New Features**
  * Added a centralized index helper and a new default config key for extra indexes.
* **Refactor**
  * Updated multiple models to use the centralized index helper for index definitions.
  * Centralized index merging and deduplication, keeping Postgres-specific handling intact.
* **Chores**
  * Reduced inline DB-specific conditionals in model configurations for consistency.

### DIFF
--- a/macros/edr/system/system_utils/get_indexes_for_model.sql
+++ b/macros/edr/system/system_utils/get_indexes_for_model.sql
@@ -1,4 +1,5 @@
 {% macro get_indexes_for_model(model_name, base_indexes) %}
+    {# dbt's indexes config is Postgres-only. Returns [] for all other targets. #}
     {%- if target.type != "postgres" -%} {{ return([]) }} {%- endif -%}
 
     {% set extra_indexes_config = elementary.get_config_var(
@@ -6,6 +7,7 @@
     ) %}
     {% set extra_indexes = extra_indexes_config.get(model_name, []) %}
 
+    {# Dedup by sorted column set only. Base indexes take precedence on conflicts. #}
     {% set seen_column_sets = [] %}
     {% set merged_indexes = [] %}
 
@@ -18,6 +20,16 @@
     {% endfor %}
 
     {% for idx in extra_indexes %}
+        {% if idx.columns is not defined %}
+            {{
+                exceptions.raise_compiler_error(
+                    "elementary_extra_indexes entry for '"
+                    ~ model_name
+                    ~ "' is missing 'columns': "
+                    ~ idx
+                )
+            }}
+        {% endif %}
         {% set sorted_cols = idx.columns | sort | join(",") %}
         {% if sorted_cols not in seen_column_sets %}
             {% do seen_column_sets.append(sorted_cols) %}
--- a/macros/edr/system/system_utils/get_indexes_for_model.sql
+++ b/macros/edr/system/system_utils/get_indexes_for_model.sql
@@ -1,7 +1,9 @@
 {% macro get_indexes_for_model(model_name, base_indexes) %}
     {%- if target.type != "postgres" -%} {{ return([]) }} {%- endif -%}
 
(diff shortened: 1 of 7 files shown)


macros/edr/system/system_utils/get_indexes_for_model.sql
```diff
@@ -0,0 +1,29 @@
+{% macro get_indexes_for_model(model_name, base_indexes) %}
+    {%- if target.type != "postgres" -%} {{ return([]) }} {%- endif -%}
```
MikaKerman, line 3: Nit: now that the Postgres gating moved from the individual model files into this macro, it would be good to document why — something like:

```jinja2
{# dbt's indexes config is Postgres-only. Returns [] for all other targets. #}
```

macros/edr/system/system_utils/get_indexes_for_model.sql
```diff
@@ -0,0 +1,29 @@
...
+
+    {% set extra_indexes_config = elementary.get_config_var(
+        "elementary_extra_indexes"
+    ) %}
+    {% set extra_indexes = extra_indexes_config.get(model_name, []) %}
```
MikaKerman, line 8: Nit: if a user passes an entry missing the `columns` key (e.g. a typo like `{"column": ["created_at"]}`), `idx.columns` will cause a cryptic Jinja compilation error. A small guard would make the error much clearer:

```jinja2
{% for idx in extra_indexes %}
    {% if idx.columns is not defined %}
        {{ exceptions.raise_compiler_error(
            "elementary_extra_indexes entry for '" ~ model_name ~ "' is missing 'columns': " ~ idx
        ) }}
    {% endif %}
    ...
{% endfor %}
```

Not a blocker — the compile error would still point to the right area — but would improve DX for anyone configuring this var.

macros/edr/system/system_utils/get_indexes_for_model.sql
```diff
@@ -0,0 +1,29 @@
...
+
+    {% set seen_column_sets = [] %}
```
MikaKerman, line 11: Nit: the dedup logic only considers the sorted column set, so an extra index with `{"columns": ["id"], "unique": true}` would be silently dropped if the base already has `{"columns": ["id"]}`. I see Devin's response that this is intentional for now — makes sense. A one-line comment would help future readers understand the precedence:

```jinja2
{# Dedup by sorted column set only. Base indexes take precedence on conflicts. #}
```

MikaKerman (approved):
Looks good! A few minor nits posted as inline comments — none are blockers.

PR #1030, opened 2026-06-24
APP-1316: add BigQuery expiration to elementary test temp tables

MikaKerman:
## Summary

Adds a 1-hour `expiration_timestamp` to BigQuery test temp tables at CREATE time as a safety net against orphan accumulation when `on_run_end` cleanup fails.

Instead of a separate ALTER TABLE macro, threads an `expiration_hours` parameter through the existing infrastructure:

```
create_elementary_test_table
  → create_or_replace(false, relation, sql, expiration_hours=1)
    → edr_create_table_as(..., expiration_hours=1)
      → bigquery__edr_get_create_table_as_sql(...)
        → CREATE TABLE ... OPTIONS(expiration_timestamp=... INTERVAL 1 hour)
```

On BigQuery, when `expiration_hours` is set and `temporary` is false, the CREATE statement includes the expiration option — single API call, no separate ALTER needed. Other adapters silently ignore the parameter via Jinja kwargs.

**Why not just change `temporary=false` to `temporary=true`?** On Snowflake/Postgres/Redshift, `temporary=true` creates session-scoped temp tables that aren't visible across dbt threads — test tables created during parallel test execution wouldn't be readable from `on_run_end` which runs on a different connection.

Resolves [APP-1316](https://linear.app/elementary/issue/APP-1316)

Link to Devin session: https://app.devin.ai/sessions/2a53f516ec2142da97e0e3e0f75f0489
Requested by: @MikaKerman

## Summary by CodeRabbit

* **New Features**
  * Added `expiration_hours` support for creating tables, including through create-or-replace, and threaded through adapter implementations.
  * Where supported, table creation can include an expiration clause (notably BigQuery).
* **Tests**
  * Temporary test tables are now created with `expiration_hours=1`, ensuring they expire after 1 hour.


macros/edr/tests/test_utils/create_elementary_test_table.sql
```diff
@@ -32,7 +32,9 @@
         ) -%}
 
         {# Create the table if it doesnt exist #}
-        {%- do elementary.create_or_replace(false, temp_table_relation, sql_query) %}
+        {%- do elementary.create_or_replace(
+            false, temp_table_relation, sql_query, expiration_hours=1
```
haritamar, line 37: Add a comment here that this table needs to survive cross-session, so it can't be temp, but we can set expiration hours for adapters that support it (currently only BQ)

haritamar (approved):

PR #1041, opened 2026-07-23
CORE-1210: optimize insert_rows / render_row_to_sql Jinja path (Fusion artifact upload slowdown)

haritamar:
## Summary

Artifact upload (`DBT_COLUMNS`, ~65k rows) spends minutes in Jinja building `INSERT … VALUES`, dramatically worse under dbt Fusion (MiniJinja FFI overhead per macro/dict boundary). Customer trace: ~8m15s in `render_row_to_sql` across 65,126 rows while the warehouse inserts took only ~45s.

This reworks the hot path in `macros/utils/table_operations/insert_rows.sql` so the per-row loop does only a dict lookup + value render, moving all repeated work out of the loop. No change to the emitted SQL shape (still `INSERT … VALUES`) or to insert semantics across adapters.

### What changed

**(1) Remove per-row duration instrumentation.** `render_row_to_sql` no longer calls `begin/end_duration_measure_context` (was ~65k stack push/pop + `utcnow()` + long name concat per row). Measurement stays at `get_insert_rows_queries` / `insert_rows` level.

**(2) Hoist column-level work out of the loop** into a new `get_row_render_metadata(columns, sample_row)`, computed once per insert:
```
column_meta[i] = {
  is_created_at:  column.name|lower == "created_at",
  row_key:        <actual dict key resolved once from the first row>,
  normalized_type: normalize_data_type(column.dtype),   # was called per cell
}
created_at_sql = edr_current_timestamp()                 # was called per created_at cell
```
This removes ~900k `normalize_data_type → data_type_list` `adapter.dispatch` calls for the customer case.

**(3) Drop `insensitive_get_dict_value` in this path.** `row_key` is resolved once (exact / lower / upper against the first row — flatten emits consistent keys), then the loop does a direct `row.get(row_key)`.

**(4) Resolve `adapter.dispatch` once per insert.** `render_value` and `escape_special_chars` impls are resolved once and reused. `render_value` / `default__render_value` gained an optional `escaper` param so the pre-resolved escaper is threaded through instead of re-dispatching per string cell (default behavior unchanged when `escaper` is omitted, for other callers).

**(5) Fix quadratic chunk assembly.** Instead of `current_query.data + "," + row_sql` on a growing multi-MB string, each chunk accumulates row SQL in a list and joins once:
```
base_insert_query + (chunk.rows | join(","))
```
Length tracking is done incrementally (namespace counter) so no giant string is rebuilt to check `query_max_size`. Existing chunking semantics preserved: `query_max_size`, `chunk_size`, and `on_query_exceed` oversized-row handling.

### Correctness / tests

- Existing `integration_tests/tests/test_dbt_artifacts` suite passes on Postgres (46 passed, 4 skipped), including `test_artifacts_collection_in_multiple_row_batches` (forces small `query_max_size` → multiple chunks).
- Added `test_insert_rows.py` + `test_render_insert_rows_queries` macro covering the rendered VALUES output: single-quote escaping, numbers, nulls, nested mapping/sequence via `tojson`, and chunk splitting.
- The full warehouse matrix (incl. Fusion on Snowflake/BigQuery/Databricks) runs in CI on this PR.

Link to Devin session: https://app.devin.ai/sessions/11d70ed2cdfe4ae182c112e4135de8d3
Requested by: @haritamar

## Summary by CodeRabbit

- **New Features**
  - Improved insert-row SQL rendering for mixed values, including adapter-aware quoting/escaping and proper handling of nulls and nested/JSON-like structures.
  - Added/expanded macro coverage to support configurable escaping samples and consistent chunking behavior for large inserts.

- **Bug Fixes**
  - Strengthened escaping/value rendering to ensure correct string literals (including special characters) and consistent literal serialization across adapters.
  - More reliable chunk splitting, producing correct multiple-query output when row size limits are exceeded.

- **Tests**
  - Added integration tests validating rendered values, escaping behavior, and chunked query splitting.

### DIFF
--- a/integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
+++ b/integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
@@ -1,0 +1,37 @@
+{% macro test_render_insert_rows_queries(query_max_size=none, chunk_size=5000) %}
+    {% set columns = [
+        {"name": "name", "dtype": "varchar"},
+        {"name": "int_col", "dtype": "integer"},
+        {"name": "float_col", "dtype": "float"},
+        {"name": "bool_col", "dtype": "boolean"},
+        {"name": "json_col", "dtype": "varchar"},
+        {"name": "null_col", "dtype": "varchar"},
+        {"name": "created_at", "dtype": "timestamp"},
+    ] %}
+    {% set rows = [
+        {
+            "name": "O'Brien",
+            "int_col": 42,
+            "float_col": 3.14,
+            "bool_col": true,
+            "json_col": {"a": 1, "b": [1, 2]},
+            "null_col": none,
+        },
+        {
+            "name": "second",
+            "int_col": 0,
+            "float_col": -1.5,
+            "bool_col": false,
+            "json_col": [1, "x"],
+            "null_col": "here",
+        },
+    ] %}
+    {% set queries = elementary.get_insert_rows_queries(
+        "my_table",
+        columns,
+        rows,
+        query_max_size=query_max_size,
+        chunk_size=chunk_size,
+    ) %}
+    {% do return(queries) %}
+{% endmacro %}
--- a/integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
+++ b/integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
@@ -33,5 +33,12 @@
         query_max_size=query_max_size,
         chunk_size=chunk_size,
     ) %}
-    {% do return(queries) %}
+    {# Return the adapter-escaped literal alongside the queries so assertions can
+       stay adapter-aware (escaping differs per warehouse, e.g. '' vs \'). #}
+    {% do return(
+        {
+            "queries": queries,
+            "escaped_quote_name": elementary.escape_special_chars("O'Brien"),
+        }
+    ) %}
 {% endmacro %}
--- a/integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
+++ b/integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
@@ -1,31 +1,6 @@
-{% macro test_render_insert_rows_queries(query_max_size=none, chunk_size=5000) %}
-    {% set columns = [
-        {"name": "name", "dtype": "varchar"},
-        {"name": "int_col", "dtype": "integer"},
-        {"name": "float_col", "dtype": "float"},
-        {"name": "bool_col", "dtype": "boolean"},
-        {"name": "json_col", "dtype": "varchar"},
-        {"name": "null_col", "dtype": "varchar"},
-        {"name": "created_at", "dtype": "timestamp"},
-    ] %}
-    {% set rows = [
-        {
-            "name": "O'Brien",
-            "int_col": 42,
-            "float_col": 3.14,
-            "bool_col": true,
-            "json_col": {"a": 1, "b": [1, 2]},
-            "null_col": none,
-        },
-        {
-            "name": "second",
-            "int_col": 0,
-            "float_col": -1.5,
-            "bool_col": false,
-            "json_col": [1, "x"],
-            "null_col": "here",
-        },
-    ] %}
+{% macro test_render_insert_rows_queries(
+    columns, rows, escape_sample, query_max_size=none, chunk_size=5000
+) %}
     {% set queries = elementary.get_insert_rows_queries(
         "my_table",
         columns,
@@ -38,7 +13,9 @@
     {% do return(
         {
             "queries": queries,
-            "escaped_quote_name": elementary.escape_special_chars("O'Brien"),
+            "escaped_quote_name": elementary.escape_special_chars(
+                escape_sample
+            ),
         }
     ) %}
 {% endmacro %}
--- a/macros/utils/table_operations/insert_rows.sql
+++ b/macros/utils/table_operations/insert_rows.sql
@@ -124,20 +124,13 @@
     {% endset %}
     {% do elementary.end_duration_measure_context("base_query_calc") %}
 
-    {# Precompute per-column rendering metadata and adapter implementations once,
-       hoisted out of the per-row loop (dtypes / created_at handling / dispatch
-       results are identical for every row). #}
-    {% set column_meta = elementary.get_row_render_metadata(
+    {% set column_meta = elementary.get_columns_metadata(
         columns, rows[0] if rows else none
     ) %}
     {% set created_at_sql = elementary.edr_current_timestamp() %}
     {% set render_value_impl = adapter.dispatch("render_value", "elementary") %}
     {% set escaper = adapter.dispatch("escape_special_chars", "elementary") %}
 
-    {# Accumulate each chunk's rendered rows in a list and join once when the
-       chunk is finalized, instead of repeatedly concatenating a growing query
-       string (which is O(n^2)). A namespace is required so reassignments made
-       inside the loop persist across iterations. #}
     {% set base_len = base_insert_query | length %}
     {% set chunk = namespace(rows=[], len=base_len, count=0) %}
     {% for row in rows %}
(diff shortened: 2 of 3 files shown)


macros/utils/table_operations/insert_rows.sql
```diff
@@ -160,50 +183,82 @@
...
+                "is_created_at": is_created_at,
+                "row_key": row_key,
+                "normalized_type": elementary.normalize_data_type(
+                    column.dtype
+                ),
+            }
+        ) %}
+    {% endfor %}
+    {% do return(column_meta) %}
+{% endmacro %}
```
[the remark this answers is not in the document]
haritamar: aside - CodeRabbit - does the above response make sense?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -160,50 +183,82 @@
...
     {{ return(insert_queries) }}
 {%- endmacro %}
 
-{% macro render_row_to_sql(row, columns) %}
-    {% do elementary.begin_duration_measure_context("render_row_to_sql") %}
+{# Precompute per-column rendering metadata once per insert so the hot per-row
+   loop only does a dict lookup + value render. `sample_row` (the first row) is
+   used to resolve the actual dict key for each column, replacing the per-cell
+   case-insensitive lookup - flatten macros emit consistent keys across rows. #}
+{% macro get_row_render_metadata(columns, sample_row) %}
```
haritamar: should it be `get_columns_metadata`?
also I think the comment can be removed - it's pretty clear what this macro does.

macros/utils/table_operations/insert_rows.sql
```diff
@@ -290,11 +345,17 @@
     -}}
 {%- endmacro -%}
 
-{%- macro render_value(value, data_type) -%}
-    {{- adapter.dispatch("render_value", "elementary")(value, data_type) -}}
+{# `escaper` lets callers pass a pre-resolved escape_special_chars implementation
+   so the hot insert path avoids an adapter.dispatch per rendered cell. When it's
+   not provided (other callers) it is resolved once here, matching prior behavior. #}
```
haritamar: Remove "matching prior behavior", don't document past code

macros/utils/table_operations/insert_rows.sql
```diff
@@ -124,34 +124,57 @@
...
+        columns, rows[0] if rows else none
+    ) %}
+    {% set created_at_sql = elementary.edr_current_timestamp() %}
+    {% set render_value_impl = adapter.dispatch("render_value", "elementary") %}
+    {% set escaper = adapter.dispatch("escape_special_chars", "elementary") %}
+
+    {# Accumulate each chunk's rendered rows in a list and join once when the
+       chunk is finalized, instead of repeatedly concatenating a growing query
+       string (which is O(n^2)). A namespace is required so reassignments made
+       inside the loop persist across iterations. #}
```
haritamar: remove this comment

macros/utils/table_operations/insert_rows.sql
```diff
@@ -124,34 +124,57 @@
     {% endset %}
     {% do elementary.end_duration_measure_context("base_query_calc") %}
 
-    {% set current_query = namespace(data=base_insert_query) %}
-    {% set current_chunk_size = namespace(data=0) %}
+    {# Precompute per-column rendering metadata and adapter implementations once,
+       hoisted out of the per-row loop (dtypes / created_at handling / dispatch
+       results are identical for every row). #}
```
haritamar: remove this comment

macros/utils/table_operations/insert_rows.sql
```diff
@@ -124,34 +124,50 @@
     {% endset %}
     {% do elementary.end_duration_measure_context("base_query_calc") %}
 
-    {% set current_query = namespace(data=base_insert_query) %}
-    {% set current_chunk_size = namespace(data=0) %}
+    {% set column_meta = elementary.get_columns_metadata(
```
haritamar, line 130: aside - getting column metadata once, instead of re-computing it over and over on every row

macros/utils/table_operations/insert_rows.sql
```diff
@@ -160,50 +176,78 @@
...
+                render_value_impl(
+                    column_value, column.normalized_type, escaper
+                )
             ) %}
-
         {% endif %}
     {% endfor %}
-    {% set row_sql = "({})".format(rendered_column_values | join(",")) %}
-
-    {% do elementary.end_duration_measure_context("render_row_to_sql") %}
```
haritamar, line 205: aside - These begin/end measurements were very expensive by themselves, so we removed them

macros/utils/table_operations/insert_rows.sql
```diff
@@ -160,50 +176,78 @@
...
     {% set rendered_column_values = [] %}
-    {% for column in columns %}
-        {% if column.name.lower() == "created_at" %}
-            {% set column_value = elementary.edr_current_timestamp() %}
-            {% do rendered_column_values.append(column_value) %}
+    {% for column in column_meta %}
+        {% if column.is_created_at %}
+            {% do rendered_column_values.append(created_at_sql) %}
         {% else %}
-            {% set column_value = elementary.insensitive_get_dict_value(
```
haritamar, line 191: aside - `insensitive_get_dict_value` replaced with a simple get, as the flattened rows already lower all keys

macros/utils/table_operations/insert_rows.sql
```diff
@@ -124,34 +124,50 @@
...
     {% do elementary.end_duration_measure_context("base_query_calc") %}
 
-    {% set current_query = namespace(data=base_insert_query) %}
-    {% set current_chunk_size = namespace(data=0) %}
+    {% set column_meta = elementary.get_columns_metadata(
+        columns, rows[0] if rows else none
+    ) %}
+    {% set created_at_sql = elementary.edr_current_timestamp() %}
+    {% set render_value_impl = adapter.dispatch("render_value", "elementary") %}
+    {% set escaper = adapter.dispatch("escape_special_chars", "elementary") %}
```
haritamar, line 135: aside - smaller optimization - pre-compute once and then pass `adapter.dispatch` to avoid doing this logic on every row.

macros/utils/table_operations/insert_rows.sql
```diff
@@ -124,34 +124,50 @@
...
+
+    {% set base_len = base_insert_query | length %}
+    {% set chunk = namespace(rows=[], len=base_len, count=0) %}
     {% for row in rows %}
-        {% set row_sql = elementary.render_row_to_sql(row, columns) %}
-        {% set query_with_row = (
-            current_query.data + ("," if not loop.first else "") + row_sql
```
haritamar, line 132: aside - replaced this with collecting an array of all the rows in the chunk and joining once (makes this O(n) instead of O(n^2))

macros/utils/table_operations/insert_rows.sql
```diff
@@ -160,50 +176,78 @@
...
+{% macro get_columns_metadata(columns, sample_row) %}
+    {% set column_meta = [] %}
+    {% for column in columns %}
+        {% set is_created_at = column.name | lower == "created_at" %}
+        {% set row_key = none %}
+        {% if not is_created_at and sample_row is not none %}
+            {% if column.name in sample_row %} {% set row_key = column.name %}
+            {% elif column.name | lower in sample_row %}
+                {% set row_key = column.name | lower %}
+            {% elif column.name | upper in sample_row %}
```
elazarlachkar, line 216: This assumes that the same "row_key" persists across different rows. Is it always the case?

macros/utils/table_operations/insert_rows.sql
```diff
@@ -290,11 +334,17 @@
     -}}
 {%- endmacro -%}
 
-{%- macro render_value(value, data_type) -%}
-    {{- adapter.dispatch("render_value", "elementary")(value, data_type) -}}
+{# `escaper` lets callers pass a pre-resolved escape_special_chars implementation
+   so the hot insert path avoids an adapter.dispatch per rendered cell. When it's
+   not provided it is resolved once here. #}
```
elazarlachkar: > When it's   not provided it is resolved once here.

This is resolved at line 131, not here, right?

integration_tests/dbt_project/macros/test_render_insert_rows_queries.sql
elazarlachkar, line 1: Why is this macro needed?
Doesn't it make more sense to call `elementary.get_insert_rows_queries` directly from the tests, with Python generated columns and rows, so the tests are more reabable?

macros/utils/table_operations/insert_rows.sql
elazarlachkar: Cool. Worth adding a small comment before line 127 saying that we expect every row to have the same keys casing.

elazarlachkar (approved):

NoyaArie (approved):

PR #1046, opened 2026-08-30
feat: expand _with_context tests and drop the dbt_expectations dependency

joostboon:
Docs: [elementary-data/elementary#2338](https://github.com/elementary-data/elementary/pull/2338)

## What

Adds five `_with_context` variants, extracts the duplicated context-column logic into one macro, and replaces the `dbt_expectations` regex call with a native cross-database implementation.

New tests, all `_with_context`: `expression_is_true`, `not_empty_string`, `expect_column_pair_values_A_to_be_greater_than_B`, `expect_compound_columns_to_be_unique`, `expect_column_values_to_match_regex_list`.

`get_context_select_clause` replaces six copies of the same ~20 line block, which would otherwise have become eleven. It also gates on `execute`: at parse time `get_columns_in_relation` is stubbed to `[]`, so the old code warned that every context column was missing, once per column per test, on every full parse.

## Why the regex rewrite

Both regex tests called `dbt_expectations.regexp_instr`, but `dbt_expectations` is not in `packages.yml`, so they only compiled for users who happened to install it themselves. `dbt_expectations` overrides `regexp_instr` for 7 adapters; everything else fell through to a `default__` calling a function that does not exist on Athena, ClickHouse, Dremio, SQL Server or Fabric.

`elementary.regexp_match` covers all 15 adapter families using each engine's own primitive, returns a boolean rather than a string position, and raises a clear compile error on SQL Server and Fabric instead of emitting SQL that cannot run.

> [!IMPORTANT]
> Snowflake's and Dremio's `regexp_like` implicitly anchor at both ends and are **not** substring searches. Both are handled, with a regression test.

## Breaking: `row_condition` is gone, use dbt's `where`

Review feedback was that `row_condition` duplicated something dbt already handles. Correct: dbt wraps the relation as `(select * from rel where cond) dbt_subquery`, and the uniqueness tests read `from {{ model }}` inside the window subquery, so rows are filtered before the window either way. It is dropped from all six tests that had it.

It was only reachable because `where` was broken on this family. With `where` set dbt passes `model` as a subquery string rather than a relation, and `get_context_select_clause` introspects it, so `where` plus `context_columns` failed with `'str object' has no attribute 'render'`; the two uniqueness tests failed on `where` alone, because `default_clause=none` introspects unconditionally. This reproduces on pre-PR master, so it predates the branch and has been broken since 0.23.1. Fixed by resolving the relation through `elementary.get_model_relation_for_test`, which exists for exactly this and is already used by 14 other elementary tests.

> [!WARNING]
> `expect_column_values_to_match_regex_with_context`, `expect_column_values_to_be_unique_with_context` and `expect_column_values_to_not_be_null_with_context` shipped with `row_condition` in 0.23.1. A project still passing it now fails to parse with an unknown argument. The replacement is dbt's `where` config.

## Behaviour changes a reviewer should check

The six existing `_with_context` tests shipped in 0.23.1. Everything below is a fix, but the first three change a test's result on unchanged data.

| Where | Change |
|---|---|
| `match_regex_with_context` on **Postgres** | A NULL value is no longer a failure. The old `dbt_expectations` path wrapped the match in `coalesce(..., 0)`; every other adapter already ignored NULL. |
| `match_regex_with_context` on **Spark / Databricks / Fabric Spark** | A pattern that can match the empty string (`\d*`, `.*`, `a?`) goes from reporting every row to reporting none. The old `length(regexp_extract(...)) = 0` could not tell "no match" from "zero-length match". |
| `expect_column_values_to_be_unique_with_context` | NULL values are no longer reported as duplicates of each other, matching dbt's own `unique`. |
| all | Mixed-case and reserved column names now work: names resolve to the warehouse's own casing and are quoted. |
| all | dbt's `where` config now works. It was a hard compile error whenever `context_columns` was set, and on the uniqueness tests even without it. |
| all | `context_columns` as a bare string is now honoured (it was silently ignored, so the whole row was sampled), and `context_columns: []` now means "all columns" as documented. |

## Deprecation

`accepted_range_with_context` is deprecated and will be removed in the next release. `dbt_utils.accepted_range` already selects `*`, so this variant can never enrich a sample, only narrow one. It still works and still honours `context_columns`; the notice goes through `edr_log_warning` rather than `exceptions.warn`, so upgrading cannot fail a run under `--warn-error`.

The trade-off is real: narrowing is a legitimate use case, and anyone using it to keep PII out of stored samples loses it. The remaining levers are the `show_sample_rows` and PII tags, `disable_test_samples`, and `test_sample_row_count`. Deleting it outright would fail `dbt parse` for any project still listing it, so it is carried for one release.

## Testing

14 integration tests in `test_with_context_sampling.py`. Full integration suite on duckdb: 231 passed, 13 skipped.

Coverage is uneven by design: the four regex tests skip `sqlserver` and `fabric`, where `regexp_match` raises because T-SQL has no regex; `not_empty_string_with_context` runs on 3 of 14 targets, because `dbt seed` reads an all-whitespace cell as NULL and only the direct seeders preserve a genuine empty string; the rest run everywhere.

> [!NOTE]
> The Spark CI job is hitting its 60 minute limit and producing no result. It does the same on other branches that add tests, and takes 31 minutes on one that adds none, so it predates this PR. Flagging it because Spark is where the `is_raw` handling lands.


macros/edr/tests/test_expect_column_values_to_match_regex_list_with_context.sql
```diff
@@ -0,0 +1,46 @@
...
+    model,
+    column_name,
+    regex_list,
+    match_on="any",
+    row_condition=none,
+    is_raw=false,
+    flags="",
+    context_columns=none
+) %}
+    {%- if not regex_list %}
```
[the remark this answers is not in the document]
joostboon: Instead of rejecting strings, the test now turns a single string into a list with one item. This matches how expect_compound_columns_to_be_unique_with_context handles column_list. This happens after the check for empty strings. This means "" still causes an error instead of becoming [""]

macros/utils/cross_db_utils/regexp_match.sql
```diff
@@ -0,0 +1,197 @@
...
+{% macro trino__regexp_supported_flags() %} {%- do return("imsU") %} {% endmacro %}
+{% macro athena__regexp_supported_flags() %} {%- do return("imsU") %} {% endmacro %}
+{% macro clickhouse__regexp_supported_flags() %} {%- do return("imsU") %} {% endmacro %}
+{% macro vertica__regexp_supported_flags() %} {%- do return("bcgimnx") %} {% endmacro %}
+{% macro dremio__regexp_supported_flags() %} {%- do return("imsx") %} {% endmacro %}
+
+{# Fallback for adapters we have no override for. `regexp_instr` is the most
+   widely implemented position function, and > 0 makes it a search. #}
+{% macro default__regexp_match(string, regex, is_raw, flags) %}
+    regexp_instr({{ string }}, '{{ regex }}') > 0
```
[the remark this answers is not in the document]
joostboon: Confirmed: regex: "O'Reilly" emits col ~ 'O'Reilly', and is_raw on Snowflake terminates early on a pattern containing "$$". Deferring to a follow-up rather than doing it here, because a correct fix needs per-adapter literal handling and the obvious shortcut is harmful: elementary.escape_special_chars maps \ → \\, which turns a pattern of \d+ into a literal backslash followed by d. The right fix escapes only the quote character in each dialect's own way ('', not \'), leaves backslashes alone, and handles $$ separately on the raw path. Also pre-existing: the previous dbt_expectations.regexp_instr call interpolated the pattern the same way

macros/edr/tests/test_utils/get_context_select_clause.sql
```diff
@@ -0,0 +1,117 @@
...
+    Args:
+        model: the relation under test.
+        tested_columns: columns the test itself needs, always selected first.
+            `none` entries are ignored, so table-level tests can pass [].
+        context_columns: user-requested extra columns. Anything that is not a
+            list (including `none`) means "no context requested".
+        test_name: used in the skipped-column warning.
+        default_clause: what to select when no context is requested. Pass
+            `none` to list every column explicitly, which callers need when a
+            bare `*` would leak a helper column such as `elementary_n_records`.
```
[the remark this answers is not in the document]
joostboon: Applied in ca0ff338. Quoting only the introspected names, since those are already in the relation's real case; the user-supplied context and tested column names stay unquoted because callers write them in arbitrary case and quoting would stop them matching.

macros/utils/cross_db_utils/regexp_match.sql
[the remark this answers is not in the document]
joostboon: Fixed in fe14af0b rather than deferred, since the follow-up turned out to be small once the escaping was kept per adapter.

`regexp_pattern_literal` now renders the pattern and escapes the delimiter only, so the pattern's backslashes are untouched: doubling (`''`) by default, and `\'` on BigQuery, ClickHouse and the Spark family, none of which take the doubled form. ClickHouse moved off doubling deliberately: `ClickHouseDirectSeeder` in integration_tests already proves `\'` works there, and I could not confirm `''` does.

The raw paths get the opposite treatment, since a raw literal escapes nothing. BigQuery moves the delimiter to whichever quote the pattern does not use. Snowflake has no alternative dollar tag, so a raw pattern containing `$$` is refused rather than requoted: requoting would start consuming the very backslashes `is_raw` was passed to preserve. Both refusals are gated on `execute`, so a stray pattern fails its own node instead of aborting parse for the project.

Verified by rendering the real macros for all fourteen adapters under a dbt-like Jinja environment: right escape form per adapter, `\d+` untouched everywhere, both inexpressible raw shapes raising at run time and still rendering at parse time. The integration test added alongside runs the `O'Reilly` case on the full warehouse matrix, since only real engines can confirm which escape each one accepts.

No follow-up issue needed.

macros/edr/tests/with_context/test_expect_column_pair_values_A_to_be_greater_than_B_with_context.sql
```diff
@@ -0,0 +1,21 @@
+{% test expect_column_pair_values_A_to_be_greater_than_B_with_context(
+    model,
+    column_A,
+    column_B,
+    or_equal=false,
+    row_condition=none,
+    context_columns=none
+) %}
+    {%- set select_clause = elementary.get_context_select_clause(
```
haritamar, line 4: Can we change this call and other calls to `get_context_select_clause` to pass kwargs rather than args? I think it will be a bit clearer

macros/edr/tests/test_expect_column_pair_values_A_to_be_greater_than_B_with_context.sql
```diff
@@ -0,0 +1,21 @@
+{% test expect_column_pair_values_A_to_be_greater_than_B_with_context(
+    model,
+    column_A,
+    column_B,
+    or_equal=false,
+    row_condition=none,
```
haritamar: Isn't dbt's [where](https://docs.getdbt.com/reference/resource-configs/where?version=2) enough? Do we need `row_condition`?

macros/edr/tests/with_context/test_expect_column_pair_values_A_to_be_greater_than_B_with_context.sql
```diff
@@ -0,0 +1,21 @@
+{% test expect_column_pair_values_A_to_be_greater_than_B_with_context(
```
haritamar, line 1: General comment (just didn't have anywhere to put it) - maybe we should put all the "with_context" tests in a dedicated folder?
as there are a lot of them.

macros/utils/cross_db_utils/regexp_match.sql
```diff
@@ -0,0 +1,403 @@
...
+            literals: ClickHouse and Redshift. Snowflake is exposed too whenever
+            `is_raw` is false, which is the default. A raw literal escapes
+            nothing, so a pattern that contains its delimiter cannot be
+            expressed as one and is refused rather than silently requoted.
+        flags: regex flags. `i` (case-insensitive) is honored on every adapter
+            that supports flags at all. Anything the adapter does not accept is
+            dropped with a warning, by `regexp_sanitize_flags` below. `-` clears
+            the flags after it, and is accepted only on adapters whose alphabet
+            lists it, which is not the same as taking inline flags: Postgres
+            takes them inline and still cannot express a negation.
```
haritamar: This docstring is a bit too long and technical IMO, please make it more concise.

haritamar: In general this file has too many comment lines, please make them more concise / remove ones where the meaning in the code is clear enough.
Keep short context where it is indeed valuable and not clear from the code.
Avoid keeping here context that is a result of an AI conversation, and assume anyone reading this has no context of the specific PR / session in which the feature was added.

macros/edr/tests/test_expect_column_pair_values_A_to_be_greater_than_B_with_context.sql
joostboon: We initially added `row_condition` to match the tests we are replacing, but you're right that it's old configuration and dbt's `where` is cleaner. Updated all the tests to use it.

macros/edr/tests/with_context/test_expect_column_pair_values_A_to_be_greater_than_B_with_context.sql
joostboon: Good call, done. All eleven call sites now pass kwargs.

macros/edr/tests/with_context/test_expect_column_pair_values_A_to_be_greater_than_B_with_context.sql
joostboon: Agreed, moved all eleven to `macros/edr/tests/with_context/`. `macro-paths` is recursive so nothing else had to change.

macros/utils/cross_db_utils/regexp_match.sql
joostboon: You're right, trimmed both. For now I kept only the notes where getting it wrong causes a real bug, like Snowflake's and Dremio's `regexp_like` anchoring at both ends.

haritamar (approved):

PR #1050, opened 2026-09-02
feat: detect changes to settled historical metrics

joostboon:
Detect historical restatements: if a settled day's revenue changes from 100 to 120 after a merge or full refresh, `elementary.metric_stability` flags it by comparing that bucket with its own earlier measurements. Comparing different days with anomaly detection does not directly check this expectation.

## Behavior

- Compare selected column metrics against the previous (`last_check`, default) or earliest retained post-settlement measurement (`first_check`), or both.
- Ignore measurements taken before `min_bucket_age`. The first eligible run establishes a baseline.
- Fail above `max_change_percent` (1 means 1%). The default is zero with a tiny relative floor for floating-point noise; movement away from a zero baseline always fails.
- Report previously measured buckets or dimensions that disappear from the actual rescan window as `missing_bucket`, with a NULL current value rather than an invented zero.
- With Elementary's test materialization enabled, persist failure samples containing bucket/dimensions, old and new values, baseline timestamps, and deltas. Existing sample limits and privacy settings apply.

```yaml
models:
  - name: orders
    tests:
      - elementary.metric_stability:
          columns: [revenue_amount]
          metrics: [sum]
          timestamp_column: order_ts
          time_bucket: {count: 1, period: day}
          min_bucket_age: {count: 4, period: week}
          days_back: 90
          change_since: [first_check]
          max_change_percent: 1
```

This example watches daily buckets roughly 28–90 days old. It does not protect all history. Longer windows increase rescanning and history-storage costs; incremental models and sources also use `backfill_days`, which defaults to `days_back`.

## Review guide

1. [`test_metric_stability.sql`](https://github.com/elementary-data/dbt-data-reliability/blob/a2328eae47cf75d62394cad1f0f6f6c10b706d01/macros/edr/tests/test_metric_stability.sql): validate settings, resolve column names and scan windows, collect per-column metrics, and materialize failures for sampling.
2. [`metric_stability_query.sql`](https://github.com/elementary-data/dbt-data-reliability/blob/a2328eae47cf75d62394cad1f0f6f6c10b706d01/macros/edr/data_monitoring/monitors_query/metric_stability_query.sql): combine persisted/current measurements, choose baselines, and detect changes or missing buckets. History is restricted to each column's actual scan bounds to avoid treating unscanned data as deleted.
3. [`test_metric_stability.py`](https://github.com/elementary-data/dbt-data-reliability/blob/a2328eae47cf75d62394cad1f0f6f6c10b706d01/integration_tests/tests/test_metric_stability.py): integration coverage for restatements, gradual drift, settling, thresholds, history persistence, weekly buckets, multiple/quoted columns, failure samples, and missing buckets/dimensions inside and outside coverage.

## Operational limits

`last_check` automatically advances: 100 → 120 fails, then another 120 passes. `first_check` continues failing against 100 while that baseline remains retained and the bucket remains in coverage. There is no explicit baseline acceptance/reset operation. Stable aggregates also cannot detect offsetting row changes. See [usage and baseline guidance](https://github.com/elementary-data/dbt-data-reliability/blob/a2328eae47cf75d62394cad1f0f6f6c10b706d01/docs/metric_stability.md).

## Validation

- **17 integration cases passed on DuckDB**, with Elementary sampling enabled, including assertions on persisted failure details.
- SQL/Python formatting and whitespace checks passed.
- The updated code still needs the PR's cross-adapter CI; the local run does not establish portability to every warehouse.


macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+
+    {%- if execute and elementary.is_test_command() and elementary.is_elementary_enabled() %}
+
+        {#- yaml lets a single value be written as a scalar, and iterating a
+            string in jinja walks it character by character. -#}
+        {%- if columns is string %} {% set columns = [columns] %} {%- endif %}
+        {%- if metrics is string %} {% set metrics = [metrics] %} {%- endif %}
+        {%- if change_since is string %}
+            {% set change_since = [change_since] %}
+        {%- endif %}
```
haritamar: What is all this commented out code? is it relevant?

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+    max_change_percent=0,
+    time_bucket=none,
+    where_expression=none,
+    days_back=none,
+    backfill_days=none,
+    dimensions=none
+) %}
+    {{ config(tags=["elementary-tests"]) }}
+
+    {%- if execute and elementary.is_test_command() and elementary.is_elementary_enabled() %}
```
haritamar: It's better to do:

```
{%- if not(execute and elementary.is_test_command() and elementary.is_elementary_enabled()) %}
     {% do return() %}
{% endif %}
```
So we don't have to ident all the code below

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
```diff
@@ -0,0 +1,234 @@
...
+  zero and the score is forced to zero, so it never fires. With the value
+  included, n unchanged observations followed by one value v give mean v/(n+1)
+  and stddev v/sqrt(n+1), so the score is n/sqrt(n+1): the v cancels and the
+  score reflects how long the history is rather than how large the change was.
+
+  The version history this reads is already collected. `data_monitoring_metrics`
+  is append-only (rows are inserted by the on-run-end hook), and a metric `id`
+  hashes the table, column, metric name and bucket_end while deliberately
+  excluding `updated_at` and `metric_value`. So re-measuring a bucket appends a
+  new row, and the earlier measurements remain.
```
haritamar: I think this comment is a bit too long, can we shorten it a bit?

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+            {%- set columns = deduped_columns %}
+        {%- endif %}
+
+        {%- if not change_since %}
+            {{
+                exceptions.raise_compiler_error(
+                    "metric_stability requires at least one baseline in `change_since`: 'last_check', 'first_check', or both."
+                )
+            }}
+        {%- endif %}
```
haritamar: Please create a `_validate_metric_stability_arguments` macro, and move all the argument validation logic there.
I think it will make the test more readable.

(also will be good that this macro will be palced after the test, at the bottom)

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+            would create a table per column and leave the cache pointing at the
+            last one, so all but the final column would be compared against
+            stale measurements.
+
+            Each column gets its own table, created directly from its select.
+            Creating one table empty and filling it with INSERT statements
+            loses the rows on adapters where dbt rolls back the test's
+            transaction, which leaves data_monitoring_metrics with no history
+            at all and makes this test a silent permanent pass. The tables are
+            unioned at read time instead. -#}
```
haritamar: I think this comment should be removed, or at least simplified.
For example it mentions `collect_column_metrics` even though it's not even called here - why is it relevant?

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+  since a previous run.
+
+  Regular anomaly detection compares one bucket against neighbouring buckets, so
+  it cannot see this: a restatement that shifts many historical buckets together
+  moves the training baseline along with the data, and normal period-to-period
+  variation is usually far wider than the change being looked for.
+
+  Arguments:
+    columns            - columns to monitor.
+    metrics            - metric types to monitor per column (e.g. [sum]).
```
haritamar: Do we want to support dimensions in this test?
Not critical for a first version I think - just making sure it's not a requirement.

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+    timestamp_column,
+    min_bucket_age,
+    change_since=["last_check"],
+    max_change_percent=0,
+    time_bucket=none,
+    where_expression=none,
+    days_back=none,
+    backfill_days=none,
+    dimensions=none
+) %}
```
haritamar, line 17: General note - this macro is very long, think if it's possible to split it further.
(I added below a comment specifically about logic for validating arguments, which seems to dominate many of the lines)

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+            time_bucket,
+            dimensions,
+            collected_by="metric_stability",
+        ) %}
+        {% set metric_names = metrics %}
+
+        {#- The measurement window has to extend past min_bucket_age and has to
+            be wide enough to hold whole buckets, so it is derived from both
+            rather than falling back to defaults that would leave nothing to
+            compare. -#}
```
haritamar: This comment is not clear enough - "derived from both" - both what?

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+  get_metric_buckets_min_and_max takes its incremental branch for sources and
+  incremental models, where backfill_days sets the window; every other model
+  takes the regular branch, which re-measures the whole days_back window and
+  ignores backfill_days entirely. days_back additionally bounds the comparison
+  itself, so it always matters.
+
+  Unset parameters are derived from min_bucket_age, because the package defaults
+  (days_back 14, backfill_days 2) are unrelated to how long a bucket needs
+  watching and would silently leave nothing to compare. An explicit value too
+  small to ever produce a comparison raises instead.
```
haritamar: This comment is not very clear. Explain in a concise manner what is this macro's purpose, what it returns, etc.

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+#}
+{% macro resolve_metric_stability_window(
+    model_graph_node, min_bucket_age, time_bucket, days_back, backfill_days
+) %}
+    {%- set age_kwargs = {min_bucket_age.period ~ "s": min_bucket_age.count} %}
+    {#- Kept as a fraction of a day. Ceiling it first would turn a sub-day age
+        into a whole day and reject a days_back that in fact covers many
+        settled buckets. -#}
```
haritamar: Remove this comment

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,563 @@
...
+            "Unsupported time_bucket period for metric_stability: '"
+            ~ time_bucket.period
+            ~ "'."
+        ) %}
+    {%- endif %}
+
+    {#- Twice the age, so a bucket is observed over a stretch rather than for a
+        single run, which is what lets 'first_check' see drift accumulate; and
+        at least two whole buckets past the age, or the settled band is narrower
+        than one bucket and nothing is ever both settled and still measured. -#}
```
haritamar: This doesn't explain the `age_days + 1` line - is it actually needed?

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
```diff
@@ -0,0 +1,234 @@
...
+    {#- Repeating a float aggregate can differ in the last bits when the scan is
+        partitioned differently between runs, since floating point addition is
+        not associative. That is a relative change around 1e-14, which a strict
+        comparison against the default of 0 reports as a failure on data nobody
+        touched. The floor sits far above that and far below any real movement,
+        and leaves the zero-crossing rule below untouched. -#}
+    {%- set change_percent_noise_floor = 0.000000001 %}
+    {%- set change_threshold = "%.10f" | format(
+        [max_change_percent, change_percent_noise_floor] | max
+    ) %}
```
haritamar, line 59: Is all this commented out code needed?

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
```diff
@@ -0,0 +1,234 @@
...
+                ~ " is not null and (("
+                ~ baseline_column
+                ~ " = 0 and metric_value != 0) or ("
+                ~ baseline_column
+                ~ " != 0 and "
+                ~ elementary.metric_stability_change_percent(baseline_column)
+                ~ " > "
+                ~ change_threshold
+                ~ ")))"
+            ) %}
```
haritamar: This condition is not readable.
Use perhaps a {% set %} ... {% endset %} block to create each condition with Jinja.

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
```diff
@@ -0,0 +1,234 @@
...
+  WHERE predicate and the reported columns so the two cannot drift apart.
+#}
+{% macro metric_stability_change_percent(baseline_column) %}
+    {%- do return(
+        "abs(metric_value - "
+        ~ baseline_column
+        ~ ") / abs("
+        ~ baseline_column
+        ~ ") * 100.0"
+    ) %}
```
haritamar: I think formatting with Jinja is more readable:

{%- macro metric_stability_change_percent(baseline_column) -%}
     abs( {{metric_value}} - ....)
{%- endmacro -%}

(this is also true in a couple of other places)

macros/edr/tests/test_metric_stability.sql
joostboon: Done, switched to the early return so the body is no longer indented.

macros/edr/tests/test_metric_stability.sql
joostboon: Done. `_validate_metric_stability_arguments` now holds all of it and sits at the bottom of the file.

macros/edr/tests/test_metric_stability.sql
joostboon: Split into `_validate_metric_stability_arguments` and `_collect_metric_stability_metrics`, both below the test. The test body is down from ~440 lines to ~100.

macros/edr/tests/test_metric_stability.sql
joostboon: Those were Jinja `{# #}` comments, not disabled code, but they were far too long. Removed here and in the other spots you flagged.

macros/edr/tests/test_metric_stability.sql
joostboon: Removed. It described `collect_column_metrics`, which this code does not call, so it was misleading on top of being long. The macro now has a two-line comment saying what it returns.

macros/edr/tests/test_metric_stability.sql
joostboon: Rewritten: "Size the scan from the settling age and bucket duration."

macros/edr/tests/test_metric_stability.sql
joostboon: Rewritten to one sentence on what it returns and when it raises.

macros/edr/tests/test_metric_stability.sql
joostboon: Removed.

macros/edr/tests/test_metric_stability.sql
joostboon: It is needed, and the code now says why. The three floors are named: `twice_the_age`, `two_buckets_past_age` and `one_day_past_age`. The third only binds for sub-day ages, where the other two can round the whole window down to a single day even though `days_back` is counted in whole days and the metrics scan starts at midnight. It is also what keeps the floor at 2 or more for every age, so a `days_back` the query would truncate to 0 or 1 days gets rejected rather than silently producing an empty comparison window. Added a test for that (`days_back` of 1 and of 0.5 against an hourly age both error).

macros/edr/tests/test_metric_stability.sql
joostboon: Yes, dimensions work. The metric `id` in `column_monitoring_query.sql` already hashes `dimension` and `dimension_value`, so each combination is its own series with its own history and baseline, and the window functions partitioning by `id` are per-combination. Covered by `test_metric_stability_reports_disappearing_bucket` (parametrized with and without a dimension) and `test_metric_stability_reports_disappearing_dimension`. Written down in docs/metric_stability.md now, since the macro comment is just a pointer there.

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
joostboon: Shortened to two lines, with the details in docs/metric_stability.md.

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
joostboon: Same as the other one: Jinja comments rather than disabled code, but too long. Both blocks are gone, replaced by a single line about the noise floor.

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
joostboon: Done, it is a `{% set %}...{% endset %}` block now.

macros/edr/data_monitoring/monitors_query/metric_stability_query.sql
joostboon: Done. `metric_stability_change_percent` emits the SQL as Jinja instead of returning a concatenated string, and the settled-history window is built the same way (it was the other place you meant, I think).

joostboon:
Pushed `84881217` addressing the review round. Replies are on each thread; the structural ones are the two extracted macros (`_validate_metric_stability_arguments` and `_collect_metric_stability_metrics`, both below the test), the early return, and the comment cleanup.

Two things in that commit are not from review comments, but the PR needed them:

- **Postgres** was failing six of these tests. Relation names are capped at 63 characters and the seed table is named after the test function, so four over-long test names never seeded. Shortened.
- **ClickHouse** was failing `test_metric_stability_reports_disappearing_bucket` with `float() argument must be ... not 'NoneType'`. ClickHouse resolves a select alias anywhere in the same select list, so `case when is_current = 1 then metric_value end as metric_value` shadowed the source column for its sibling expressions and a missing bucket reported a NULL baseline instead of the last observed value. The compared values are now carried as `measured_value` / `last_check_value` / `first_check_value` / `last_check_at` / `first_check_at`, so no output alias repeats a source column name on any adapter. Output columns are unchanged.
- `code-quality` was failing on black (integration test) and prettier (docs). Fixed.

19 integration cases pass locally on DuckDB and on Postgres. I could not run ClickHouse locally, so that one is on CI.

macros/edr/tests/test_metric_stability.sql
```diff
@@ -0,0 +1,560 @@
...
+
+    {% if not (
+        execute
+        and elementary.is_test_command()
+        and elementary.is_elementary_enabled()
+    ) %}
+        {% do return(elementary.no_results_query()) %}
+    {% endif %}
+
+    {% set arguments = elementary._validate_metric_stability_arguments(
```
haritamar: I'd maybe call it `_parse_and_validate_metric_stability_arguments`

docs/metric_stability.md
```diff
@@ -0,0 +1,91 @@
+# Metric stability
```
haritamar: The docs directory contains our mintlify docs - I think this md file is a bit out-of-place.
I think instead we should have a separate PR, to the "docs" branch, covering this test.

(Unless @joostboon you don't want it to be in the public docs, but anyway not sure an MD file should be added to this specific folder)

macros/edr/tests/test_metric_stability.sql
joostboon: Renamed. It normalizes too (scalar to list, column dedupe), so the old name undersold it.

docs/metric_stability.md
joostboon: Agreed, wrong home for it. Our docs live on the `docs` branch of elementary-data/elementary, so it is now elementary-data/elementary#2343: a `data-tests/metric-stability` page under Other Tests next to `volume-threshold`.

README and the test header now point at the published page. Have a look at how the limits are worded there, since it is public facing.

haritamar (approved):

PR #1051, opened 2026-09-03
fix: preserve newlines in Spark/Databricks compiled_code (create Spark temp tables with an explicit file format)

haritamar:
## Summary

Since 0.23.0 (#946), `spark__escape_special_chars` replaced `\n`/`\r` with spaces. dbt-databricks dispatches `databricks → spark → default`, so every Databricks customer's `compiled_code` in `dbt_run_results` was flattened to one line; any model starting with a `--` comment then fails to parse in Elementary Cloud column-level lineage (`ParseError: No expression was parsed from '-- ...'`), and inline `--` comments silently truncate lineage.

#946 attributed the OSS Spark row corruption to `INSERT VALUES` not supporting `\n` escapes. That was a misdiagnosis. The actual cause is how the intermediate table is created:

```sql
-- spark__edr_get_create_table_as_sql (non-temporary, used for artifact staging on dbt-spark)
create table <tmp> as select ... where 1=0     -- no USING
```

With Spark 3's default `spark.sql.legacy.createHiveTableByDefault=true` this is a Hive **TEXTFILE** table (`LazySimpleSerDe`/`TextInputFormat`, confirmed via `describe extended`). A text serde cannot hold a newline inside a value, so every escaped `\n` splits a row on read (the 37-tuple insert → 55 rows). The same insert into a `using parquet`/`using delta` table round-trips newlines correctly.

Fix (root cause, not a workaround):

```diff
- create table {{ relation }} as {{ sql_query }}
+ create table {{ relation }} using {{ elementary.get_config_var("spark_file_format") }} as {{ sql_query }}
```

with a new `spark__get_default_config` (`spark_file_format: parquet`, dbt-spark's own default; `fabricspark` delegates). `spark__escape_special_chars` now flattens newlines **only when** `spark_file_format == "hive"` (the one format that can't hold them) and otherwise delegates to `default__escape_special_chars`; `databricks__escape_special_chars` is removed so dbt-databricks uses the default `\\n` escaping. This also covers customers running the `dbt-spark` adapter against Databricks (target type `spark`), which a Databricks-only override would have missed.

Regression test: `test_compiled_code_preserves_newlines` runs a model whose compiled SQL starts with a `-- comment` line (`models/one.sql`) and asserts the persisted `compiled_code` keeps that comment on its own line with the query after it. Runs on all targets incl. spark/fabricspark; skipped on dbt-fusion, which currently writes no `compiled_code` rows in this path.

Verified locally on the Docker Spark Thrift setup: full `test_artifacts.py` (16 passed incl. `test_artifacts_collection_in_multiple_row_batches` and the new test), insert-rows/long-string tests, `dbt parse --target spark`, pre-commit. Also checked `spark_file_format: hive` one-off: multi-batch test still 37/37 (flattening protects the text table), newline test fails there as expected.

Note for Cloud: affected customers need the package upgrade + a dbt run; flattened models will recompute automatically since the compiled-code hash changes.

Link to Devin session: https://app.devin.ai/sessions/7c9a2683e2d446428a12decba40a004b
Open in Devin Desktop: https://app.devin.ai/desktop/session/7c9a2683e2d446428a12decba40a004b?variant=devin
Requested by: @haritamar

## Summary by CodeRabbit

* **New Features**
  * Added Spark and Fabric Spark defaults, including Parquet as the default Spark file format.
  * Spark-created non-temporary tables now explicitly use the configured file format.

* **Bug Fixes**
  * Fixed compiled SQL formatting so leading comments and newline characters are preserved.
  * Expanded compiled-code validation for Spark and Fabric Spark.
  * Standardized special-character escaping across Spark-based adapters.

### DIFF
--- a/macros/utils/table_operations/insert_rows.sql
+++ b/macros/utils/table_operations/insert_rows.sql
@@ -321,22 +321,6 @@
     {{- return(string_value | replace("'", "''")) -}}
 {%- endmacro -%}
 
-{# spark__escape_special_chars: Newlines and carriage returns are replaced with
-   spaces (not escape sequences) because Spark SQL does not support multi-line
-   string literals inside INSERT VALUES. Backslashes and single quotes use
-   C-style escaping (\\, \') which is the Spark SQL convention. #}
-{%- macro spark__escape_special_chars(string_value) -%}
-    {{-
-        return(
-            string_value
-            | replace("\\", "\\\\")
-            | replace("'", "\\'")
-            | replace("\n", " ")
-            | replace("\r", " ")
-        )
-    -}}
-{%- endmacro -%}
-
 {# `escaper` lets callers pass a pre-resolved escape_special_chars implementation
    so the hot insert path avoids an adapter.dispatch per rendered cell. When it's
    not provided, the adapter's render_value resolves it (see default__render_value). #}
--- a/macros/utils/table_operations/insert_rows.sql
+++ b/macros/utils/table_operations/insert_rows.sql
@@ -321,6 +321,28 @@
     {{- return(string_value | replace("'", "''")) -}}
 {%- endmacro -%}
 
+{# spark__escape_special_chars: Newlines and carriage returns are replaced with
+   spaces (not escape sequences) because Spark SQL does not support multi-line
+   string literals inside INSERT VALUES. Backslashes and single quotes use
+   C-style escaping (\\, \') which is the Spark SQL convention. #}
+{%- macro spark__escape_special_chars(string_value) -%}
+    {{-
+        return(
+            string_value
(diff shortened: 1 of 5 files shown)


macros/utils/table_operations/insert_rows.sql
```diff
@@ -321,22 +321,6 @@
     {{- return(string_value | replace("'", "''")) -}}
 {%- endmacro -%}
 
-{# spark__escape_special_chars: Newlines and carriage returns are replaced with
```
GuyEshdat, line 324: bring back this macro but run this logic only if elementary.get_config_var("spark_file_format") is hive, otherwise call default

GuyEshdat (approved):

Issue #1058, opened 2026-09-14
get_cache() returns None after build_flat_graph() rebinds graph — chained call sites crash every test (still present in 0.26.0; #1005 was closed by the stale bot)


AthenaSG:
Manifest.build_flat_graph() rebinds self.flat_graph to a new dict with no "elementary" key, so the on-run-start seed is discarded. get_cache(entry) falls to its none default → 'None' has no attribute 'setdefault'.

build_flat_graph() call sites in dbt-core 1.12.4 tabulated. Classic path rebuilds at parser/manifest.py:363 before on-run-start, so the seed survives. Fusion bypasses that loader, rebuilds at parser/fusion.py:129, and round-trips through from_writable_manifest (which fires __post_deserialize__ → another rebuild).

AthenaSG:
Secondary symptom: clean_current_invocation_test_tables() silently no-ops, leaking test%__tmp_% tables; Redshift has no redshift__cleanup_stale_test_tables, so no recovery.

Suggested fix: (1) setdefault_cache everywhere, (2) better — make get_cache re-seed the default dict when the key is absent.
Environment: elementary 0.26.0, dbt-core 1.12.4 with use_v2_parser: true, manifest dbt_version: 2.0.0-rc.2, dbt-redshift, Python 3.13, ERROR=14.

Daniel-Wiszowaty:
Hit this on a Snowflake project — every test in the project failed to compile (~180 of them), all with `'None' has no attribute 'update'` out of `handle_dbt_test`.

While reducing it I think I found a simpler cause than the `build_flat_graph()` rebind: **under `--use-v2-parser` the package's `on-run-start` hook never runs at all**, so the seed is never written in the first place. It isn't being written and then discarded.

Reproduction on a stock `dbt-duckdb` project — one model, two generic tests, elementary as the only other package. Same project, same command, only the parser differs:

```
$ dbt run
Elementary: Runtime data: {... "elementary_version": "0.26.0" ...}
1 of 1 START hook: elementary.on-run-start.0 ................... [RUN]
1 of 1 OK hook: elementary.on-run-start.0 ...................... [OK in 0.01s]
1 of 1 START hook: elementary.on-run-end.0 ..................... [RUN]
1 of 1 OK hook: elementary.on-run-end.0 ........................ [OK in 0.32s]

$ dbt run --use-v2-parser
(no hook lines at all)
```

[...]

PR #1060, opened 2026-09-15
fix: re-seed the elementary graph cache on access

Daniel-Wiszowaty:
Fixes the crash class in #1058.

## Problem

`get_cache` / `set_cache` / `setdefault_cache` all read `graph["elementary"]`, which is written exactly once, by `init_elementary_graph()` from the `on-run-start` hook. When that seed is absent the accessors fall through to their `none` default, and the five call sites that chain a method onto the result take the whole run down:

| call site | expression | error |
|---|---|---|
| `macros/edr/materializations/test/test.sql:382` | `get_cache("elementary_test_results").update(...)` | `'None' has no attribute 'update'` |
| `macros/edr/materializations/test/test.sql:54` | `get_cache("elementary_test_failed_row_counts").update(...)` | same |
| `macros/edr/tests/test_metric_stability.sql:196` | `get_cache("tables").get("metrics")...` | `'None' has no attribute 'get'` |
| `macros/edr/data_monitoring/anomaly_detection/store_metrics_in_cache.sql:3` | `get_cache("tables").get("metrics")...` | same |
| `macros/edr/data_monitoring/schema_changes/store_schema_snapshot_tables_in_cache.sql:2` | `get_cache("tables").get(...)` | same |

The first one runs for every test in the project, so a missing seed means every test fails to compile.

## Fix

Route the three accessors through one helper that re-seeds the full default structure when the key is absent, and have `init_elementary_graph()` share the same defaults macro so there is a single source of truth. A lost seed now degrades (empty cache) instead of raising.

This matches the two suggestions on #1058 — it is option (2), applied in one place rather than at each call site, so future chained call sites are safe too.

## How this is reachable today

Under dbt-core's v2 (Fusion) parser, the package's `on-run-start` hook **never executes**, so the seed is never written at all. Reproduction on a stock `dbt-duckdb` project with elementary 0.26.0 and dbt-core 1.12.4 — same project, same command, only the parser differs:

```
$ dbt run
Elementary: Runtime data: {... "elementary_version": "0.26.0" ...}
1 of 1 START hook: elementary.on-run-start.0 ................... [RUN]
1 of 1 OK hook: elementary.on-run-start.0 ...................... [OK in 0.01s]
1 of 1 START hook: elementary.on-run-end.0 ..................... [RUN]
1 of 1 OK hook: elementary.on-run-end.0 ........................ [OK in 0.32s]

$ dbt run --use-v2-parser
(no hook lines at all — neither hook runs)
```

`dbt build` reports `2 project hooks` on the classic path and omits them entirely under `--use-v2-parser`.

To be clear about scope: **the hooks not running is a dbt-side problem, not something this PR fixes.** It also means `on-run-end` artifact upload is skipped, which is a separate and larger issue. This PR only stops elementary from crashing when the seed is absent, whatever the reason.

## Verification

Stock `dbt-duckdb` project, elementary installed as a local package, dbt-core 1.12.4 + `dbt-core-experimental-parser` 2.0.1.

Reproducing the exact failing expression from `handle_dbt_test`:

```
# before
$ dbt run-operation probe_cache
[ERROR]: Encountered an error while running operation: Compilation Error
  'None' has no attribute 'update'

# after
$ dbt run-operation probe_cache
PROBE ok, cache now = {'x': 1}
```

where `probe_cache` is just `{% do elementary.get_cache("elementary_test_results").update({"x": 1}) %}`.

No regression on a full build:

| | before | after |
|---|---|---|
| `dbt build` (classic) | PASS=35, 2 project hooks | PASS=35, 2 project hooks |
| `dbt build --use-v2-parser` | PASS=33, no hooks | PASS=33, no hooks |

Files formatted with the repo's `sqlfmt` 0.29.0 pre-commit hook.

I have not run the warehouse integration suite — no access to those targets. Happy to adjust if you would rather see `setdefault_cache` at each call site instead.

## Summary by CodeRabbit

- **Refactor**
  - Standardized initialization of the elementary graph cache to use shared default settings.
  - Preserved existing graph values when already present.
  - Ensured default graph settings are restored when the parser rebuilds the graph.


Daniel-Wiszowaty:
Filed the hook half of this upstream: dbt-labs/dbt#16310 — `--use-v2-parser` never executes `on-run-start` / `on-run-end`, for root-project and package hooks alike, silently.

That is why the seed is missing here, so if it gets fixed this issue's crash stops being reachable via that path. The re-seeding in #1060 is still worth having as defence in depth, but the hook bug is the one that actually restores Elementary under the v2 parser — `on-run-end` artifact upload is skipped too, which no change inside this package can work around.

Worth noting the v2 parser already *reads* `on-run-start` (it parses `CREATE FUNCTION` there to register UDFs), so it sees the hooks and just doesn't run them.

PR #1061, opened 2026-09-20
fix(duckdb): use regular tables instead of temp tables under dbt Fusion

junoha:
DuckDB temp tables are session-scoped; dbt Fusion runs statements on pooled connections, so Elementary's intermediate tables are never visible to the statement that reads them. Same root cause as #920 for Redshift/Databricks — this adds the missing `duckdb__` variants.

- `duckdb__has_temp_table_support` → `false` under `is_dbt_fusion()`
- `duckdb__edr_make_temp_relation` → regular relation with explicit database/schema under Fusion
- `duckdb__edr_get_create_table_as_sql` → drop `TEMPORARY` under Fusion (DuckDB also rejects `TEMPORARY` on a qualified name)

`edr_create_table_as` is called with a hardcoded `temporary=true` in `handle_tests_results.sql` (L220, L280) and `default__create_temp_table`, so the DDL-side branch is needed in addition to `has_temp_table_support`.

dbt Core behavior is unchanged (temp tables still used). Cleanup needs no work: `duckdb__get_clean_elementary_test_tables_queries` already exists, and nothing leaks — verified: no `%__tmp_%` tables left behind.

No docs change — this is an internal adapter dispatch fix with no user-facing configuration.

## Second commit — making the integration tests runnable under Fusion

This one is separate on purpose: it touches the test harness rather than the package, and it reaches into a private attribute of dbt-duckdb. If you'd rather not carry it, drop that commit. Without it, though, `--target duckdb --runner-method fusion` cannot run at all, because the harness assumes dbt runs in-process for DuckDB:

- `DBT_FUSION_SUPPORTED_TARGETS` did not list `duckdb`, so the run aborted immediately
- DuckDB is embedded and takes an exclusive file lock, so `adapter_query_runner`'s idle connection blocked every dbt subprocess: `IO Error: Could not set lock on file ...`. `connections.cleanup_all()` is not enough — `DuckDBConnectionManager._ENV` caches the `duckdb` connection at class level and only closes it when its handle count reaches zero, so I drop the cached environment explicitly, from a `finally` block so a failed query does not leave the lock behind. It is a no-op unless the DuckDB database is a local file: non-DuckDB adapters, `path: ":memory:"` and MotherDuck are all skipped, mirroring dbt-duckdb's own `_keep_open` set (`environments/local.py:56-58`).

Results — `py.test --target duckdb --runner-method fusion`:

```
2 failed, 229 passed, 41 skipped in 1365.28s (0:22:45)
```

The 2 failures are pre-existing and unrelated: they fail identically on dbt Core with duckdb, both with a file path and with `path: ":memory:"`. Every test that exercises the code paths this patch touches passes.

- test_data_freshness_sla.py::test_deadline_not_passed_does_not_fail
- test_event_freshness_anomalies.py::test_stop_event_freshness

Two caveats if you want this in CI, which I left out of this PR: the duckdb profile needs a file path instead of `:memory:` (a Fusion subprocess cannot share an in-memory database with the harness), and it has to run serially. Happy to send that as a follow-up. Note the consequence of the guard above: as CI stands today the duckdb target uses `:memory:`, so this commit is inert there and cannot affect any existing job — it only starts doing work once the profile points at a file.

## Upstream context

This is tracked on the dbt side as https://github.com/dbt-labs/dbt/issues/15758, still open. This PR does not depend on that being fixed — the `is_dbt_fusion()` gate makes the package work on today's dbt 2.0.x either way, exactly like the Redshift/Databricks branches from #920.

This looks engine-level rather than DuckDB-specific. dbt 2.x keeps connections thread-local on a bounded worker pool (`crates/dbt-adapter/src/connection.rs`: *"A connection stays in the slot after a node finishes, so it is found again by the next node that runs on the same worker"*), so session-local state does not survive between statements. I measured this without temp tables at all — `SET search_path` set in one `run_query` is gone in the next four on Fusion, while it persists on dbt Core; a GLOBAL-scoped setting like `memory_limit` survives on both. Since Postgres temp tables are session-scoped too, `postgres__edr_get_create_table_as_sql` may have the same problem under Fusion — but there is no postgres × fusion job in CI, so I have not verified it.

## Summary by CodeRabbit

- **New Features**
  - Added support for running DuckDB integration workflows with dbt Fusion.
  - Improved temporary-relation handling for DuckDB under dbt Fusion by using regular tables when session-scoped temporary tables are unavailable.
  - Added support for detecting when temporary tables are unavailable in DuckDB Fusion workflows.
  - Improved DuckDB connection cleanup after queries and SQL execution, helping release embedded database file locks promptly.


junoha:
## Vertica

The 59 vertica failures come from [integration_tests/docker-compose-vertica.yml:12](https://github.com/elementary-data/dbt-data-reliability/blob/master/integration_tests/docker-compose-vertica.yml#L12), not from this PR. Dropping that line, or setting it to `UTC`, should fix them:

```yaml
TZ: "America/Los_Angeles"
```

It has been there since #963 (Mar 12), so this isn't a regression — it's pre-existing, and whether it bites depends on what time of day CI runs.

### Why that breaks the tests

`America/Los_Angeles` is UTC-7 in September. The bucket window is computed by a query *on the warehouse*, not in Jinja — `get_buckets_configuration.sql`:

```jinja
{%- set detection_end = elementary.get_detection_end(detection_delay) %}
{%- set detection_end_expr = elementary.edr_cast_as_timestamp(
    elementary.edr_datetime_to_sql(detection_end)) %}
```

`detection_end` derives from dbt's `run_started_at` (UTC), but it is embedded as a SQL literal and cast by Vertica in the session timezone. Truncated to a day boundary, it lands one calendar day early whenever the run starts between 00:00 and 07:00 UTC.

The tests seed the anomaly into "yesterday" (`utc_today - timedelta(1)`, computed in Python as UTC). That bucket then never exists, so nothing can be flagged and the test passes when it should fail.

### Evidence

`max_bucket_end` is one day short. From the job log, run started 2026-09-20 05:13 UTC:

```
Elementary: min_bucket_start: '2026-09-06T00:00:00' | max_bucket_end: '2026-09-19T00:00:00'
```

It should be `2026-09-20T00:00:00`, making 09-19 → 09-20 the last complete bucket. The tests that don't truncate to a day boundary show the offset outright ([link](https://github.com/elementary-data/dbt-data-reliability/actions/runs/35491106168/job/106026070971?pr=1061#step:25:30325)):

| logged `max_bucket_end` | actual UTC then |
| --- | --- |
| `2026-09-19T22:16:16` | `2026-09-20T05:16` |
| `2026-09-19T22:23:42` | `2026-09-20T05:23` |

Exactly 7 hours. And all 59 failures are that one missing bucket — 47 are `assert 'pass' == 'fail'`, and the `test_anomalies_backfill_logic.py` ones compare the buckets directly:

```
expected: 09-19, 09-18, 09-17, ... 09-06   (14 buckets)
actual:          09-18, 09-17, ... 09-06   (13 buckets)
```

No other compose file sets `TZ` — postgres, sqlserver and trino stay on UTC. That is why it is vertica-only.

## Spark / trino / dremio

All three fail for one shared reason that has nothing to do with either this PR or vertica — the MinIO images are no longer pullable from Docker Hub:

```
minio Error pull access denied for minio/minio, repository does not exist
      or may require 'docker login': denied: requested access to the resource is denied
```

integration_tests/tests/conftest.py
```diff
@@ -18,6 +18,7 @@
     "bigquery",
     "redshift",
     "databricks_catalog",
+    "duckdb",
```
joostboon, line 21: The duckdb profile is still `:memory:`, which can't work with Fusion. With this added, `--target duckdb --runner-method fusion` runs the whole suite and fails instead of aborting early. Gate on a file path, or switch the profile here?

integration_tests/tests/adapter_query_runner.py
```diff
@@ -246,8 +246,43 @@ def has_non_ref_jinja(query: str) -> bool:
...
+        if self._adapter.type() != "duckdb":
+            return
+
+        credentials = self._adapter.config.credentials
+        if credentials.path == ":memory:" or credentials.is_motherduck:
+            return
+
+        from dbt.adapters.duckdb.connections import DuckDBConnectionManager
+
+        self._adapter.connections.cleanup_all()
```
joostboon, line 275: `cleanup_all()` is outside the try, so if it raises it masks the original query error from the `finally`.

joostboon (commented):
Thanks for this! A few notes, mostly on the Fusion path:

- `handle_tests_results.sql`: under Fusion the staging table is now a real table, but `fully_drop_relation` runs after the last DuckDB `adapter.commit()`. If a rollback follows, the drop is undone. `delete_and_insert` commits again after its drops; worth doing the same here?
- `create_temp_table.sql`: with `tests_use_temp_tables: true`, each test now leaves a `<alias>__tmp_<ts>` table in the elementary schema that nothing drops.
- No CI job covers duckdb + Fusion, so the new branches aren't exercised in CI.

integration_tests/tests/adapter_query_runner.py
```diff
@@ -246,8 +246,43 @@ def has_non_ref_jinja(query: str) -> bool:
...
+        with DuckDBConnectionManager._LOCK:
+            if DuckDBConnectionManager._ENV is not None:
+                try:
+                    DuckDBConnectionManager._ENV.close()
```
joostboon, line 279: Nit: `DuckDBConnectionManager.close_all_connections()` is public and resets `_ENV` under the lock. Using it would cut down on private attribute access.

integration_tests/tests/adapter_query_runner.py
```diff
@@ -267,8 +302,11 @@ def run_query(self, prerendered_query: str) -> List[Dict[str, Any]]:
         if self.has_non_ref_jinja(prerendered_query):
             raise UnsupportedJinjaError(prerendered_query)
         sql = self.resolve_refs(prerendered_query)
-        with self._adapter.connection_named("run_query"):
-            _response, table = self._adapter.execute(sql, fetch=True)
+        try:
+            with self._adapter.connection_named("run_query"):
+                _response, table = self._adapter.execute(sql, fetch=True)
+        finally:
+            self._release_duckdb_file_lock()
```
joostboon, line 309: This tears down and rebuilds the DuckDB env after every query for any file-backed profile, even when no subprocess needs the lock. Gate on runner method (fusion/subprocess), or release once before spawning dbt?

macros/utils/table_operations/create_table_as.sql
```diff
@@ -114,8 +114,17 @@
 {% macro duckdb__edr_get_create_table_as_sql(
     temporary, relation, sql_query, expiration_hours=none
 ) %}
-  create or replace {% if temporary %} temporary {% endif %} table {{ relation }}
-  as {{ sql_query }}
+    {% if temporary and elementary.is_dbt_fusion() %}
```
joostboon, line 117: Nit: this branch is the same as the else branch with `temporary=false`. `{% set temporary = temporary and not elementary.is_dbt_fusion() %}` plus the original statement would do.

macros/utils/table_operations/make_temp_relation.sql
```diff
@@ -90,6 +90,24 @@
     {% endif %}
 {% endmacro %}
 
+{% macro duckdb__edr_make_temp_relation(base_relation, suffix) %}
```
joostboon, line 93: Identical to `redshift__edr_make_temp_relation`. Could delegate to it (like sqlserver/fabric) or share a helper. Same for `duckdb__has_temp_table_support`.

macros/utils/table_operations/has_temp_table_support.sql
```diff
@@ -33,3 +33,13 @@
     {% else %} {% do return(true) %}
     {% endif %}
 {% endmacro %}
+
+
+{% macro duckdb__has_temp_table_support() %}
```
joostboon, line 38: Since the root cause is Fusion-wide, consider one gate (e.g. `default__has_temp_table_support` returning `not is_dbt_fusion()`) instead of per-adapter branches. Postgres likely needs this too.

